PR #1, opened 2024-09-23
Copy the Github Actions workflows from the main repo

martin-g:


.github/workflows/test-lang-rust-ci-ARM.yml
```diff
@@ -0,0 +1,71 @@
...
+  RUSTFLAGS: -Dwarnings
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  arm64:
+    name: Rust on Linux ARM64
+    runs-on: ["self-hosted", "asf-arm"]
```
tisonkun, line 39: Seems no runner matched and all this workflow failed with timeout.

Does it mean that we need to go to the INFRA for a new config?

martin-g: https://issues.apache.org/jira/browse/INFRA-26149

PR #6, opened 2024-09-24
chore: Remove unexpected fuzz corpus

Xuanwo:
The `fuzz/corpus` is in the `.gitignore`, but it seems to have been mistakenly committed to the repository.


martin-g:
Thank you, @Xuanwo !

PR #7, opened 2024-09-24
chore: Use `use abc` to replace `extern crate abc`

Xuanwo:
This PR will use Rust 2021 edition's `use abc` to replace `extern crate abc`.


martin-g (approved):

martin-g:
Thank you, @Xuanwo !

PR #9, opened 2024-09-27
fix: AVRO-4055: Validate records properly when parsing a schema

woile:
See https://issues.apache.org/jira/browse/AVRO-4055

### DIFF
--- a/avro/src/error.rs
+++ b/avro/src/error.rs
@@ -350,6 +350,9 @@ pub enum Error {
     #[error("Invalid namespace {0}. It must match the regex '{1}'")]
     InvalidNamespace(String, &'static str),
 
+    #[error("Invalid schema: There is no type called '{0}', if you meant to create a record, it should be defined inside type. Please review the specification")]
+    InvalidSchemaRecord(String),
+
     #[error("Duplicate enum symbol {0}")]
     EnumSymbolDuplicate(String),
 
--- a/avro/src/error.rs
+++ b/avro/src/error.rs
@@ -350,7 +350,7 @@ pub enum Error {
     #[error("Invalid namespace {0}. It must match the regex '{1}'")]
     InvalidNamespace(String, &'static str),
 
-    #[error("Invalid schema: There is no type called '{0}', if you meant to create a record, it should be defined inside type. Please review the specification")]
+    #[error("Invalid schema: There is no type called '{0}', if you meant to define a non-primitive schema, it should be defined inside `type` attribute. Please review the specification")]
     InvalidSchemaRecord(String),
 
     #[error("Duplicate enum symbol {0}")]
--- a/avro/src/schema.rs
+++ b/avro/src/schema.rs
@@ -669,7 +669,8 @@ impl RecordField {
         validate_record_field_name(&name)?;
 
         // TODO: "type" = "<record name>"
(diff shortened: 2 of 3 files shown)


avro/src/error.rs
```diff
@@ -350,6 +350,9 @@ pub enum Error {
     #[error("Invalid namespace {0}. It must match the regex '{1}'")]
     InvalidNamespace(String, &'static str),
 
+    #[error("Invalid schema: The record type should be inside the type field")]
+    InvalidSchemaRecord(String),
```
martin-g, line 354: The argument is not printed in the error message, i.e. there is no `{0}`.

avro/src/schema.rs
```diff
@@ -1353,6 +1358,7 @@ impl Parser {
         &mut self,
         complex: &Map<String, Value>,
         enclosing_namespace: &Namespace,
+        from_field: bool,
```
martin-g: IMO it would be better to use an `enum` with well named variants instead of `true` / `false`

avro/src/schema.rs
```diff
@@ -6821,4 +6832,63 @@ mod tests {
...
+                "default": null
+                },
+                { "name": "order_date", "type": "string" }
+            ]
+            }
+        ]
+        }"#;
+
+        let schema = Schema::parse_str(invalid_schema_str);
+        println!("{:?}", schema);
```
martin-g: debug leftover

avro/src/types.rs
```diff
@@ -2771,35 +2771,36 @@ Field with name '"b"' is not a member of the map items"#,
         use crate::ser::Serializer;
         use serde::Serialize;
 
-        let schema_str = r#"{
+        let schema_str = r#"
+        {
             "type": "record",
             "name": "NamespacedMessage",
-            [NAMESPACE]
```
martin-g, line 2777: Why is this removed ?

woile: Oh my bad, it took me some time to fix this schema, and then I forgot to add the namespace again, I'll update it

avro/src/schema.rs
woile: Great idea! I'm updating it

woile:
Updated based on the feedback!

avro/src/schema.rs
```diff
@@ -1294,6 +1307,11 @@ impl Parser {
             return Ok(resolving_schema.clone());
         }
 
+        // For good error reporting we add this check
+        if name.name == *"record" {
```
martin-g: I am not sure this is always valid.
`#fetch_schema_ref()` is called also by `#parse_known_schema()` where the `name` argument could be any non-primitive schema kind, like `"MyType"` (a Schema::Ref).

martin-g: I misunderstood you idea. Let me think how to improve it!

woile: In a nutshell, if you have defined a type called "record" somewhere else, then you can actually do `"type": "record"`, but if there's no schema already defined, then you probably intended to create a new record, so it should be inside type, as `"type": {"type": "record"}`. Does it make sense?

martin-g (approved):

martin-g:
Thank you, @woile !

PR #14, opened 2024-10-01
AVRO-4063: Call `flush` on the inner `writer` during `Writer::flush`

### LINKED ISSUE #10: `apache_avro::Writer::flush` does not call `std::io::Write::flush` on the inner writer
(Moving this over from [AVRO-4063](https://issues.apache.org/jira/browse/AVRO-4063))

## Issue Overview

The Rust documentation for `apache_avro::Writer::flush` describes the function as follows:

> Flush the content appended to a `Writer`. Call this function to make sure all the content has been written before releasing the `Writer`.

However, this function does not actually guarantee that all the content will be written out after the `flush()` call, because it does not call `std::io::Write::flush` on the inner writer.

This can be a problem when the inner writer uses its own buffer.

## Example

```rust
fn main() {
    let buffered_writer = std::io::BufWriter::new(std::fs::File::create("test.avro").unwrap());

    let schema = apache_avro::Schema::parse_str(
        r#"
    {
        "type": "record",
        "name": "example_schema",
        "fields": [
            {"name": "example_field", "type": "string"}
        ]
    }
"#,
    )
    .unwrap();

    let mut writer = apache_avro::Writer::new(&schema, buffered_writer);

    let mut record = apache_avro::types::Record::new(writer.schema()).unwrap();
    record.put("example_field", "value");

    writer.append(record).unwrap();
    writer.flush().unwrap();

    let test_file_contents = std::fs::read("test.avro").unwrap();
    assert_ne!(test_file_contents.len(), 0); // this will fail
}
```

In this example, the internal `BufWriter` had not yet flushed its internal buffer after `writer.flush().unwrap()` was called. In fact, the buffer is only written out once `writer` is dropped.

## Solution

`std::io::Write::flush` should be called on the inner writer at the end of `apache_avro::Writer::flush`.

snowsignal:
## Overview

Fixes #10.

## Description

This PR calls `std::io::Write::flush` on the inner `writer` of `apache_avro::Writer` at the end of `apache_avro::Writer::flush`. This ensures that the flush is propagated and the previously written bytes are fully sent.

This PR also includes a unit test that confirms `Writer::flush` now works as expected.

### DIFF
--- a/avro/src/writer.rs
+++ b/avro/src/writer.rs
@@ -1450,7 +1450,7 @@ mod tests {
     }
 
     #[test]
-    fn avro_4063_flush_applies_to_inner_writer() {
+    fn avro_4063_flush_applies_to_inner_writer() -> TestResult {
         const SCHEMA: &str = r#"
         {
             "type": "record",
@@ -1470,7 +1470,7 @@ mod tests {
             }
         }
 
-        impl std::io::Write for TestBuffer {
+        impl Write for TestBuffer {
             fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
                 self.0.try_lock().unwrap().write(buf)
             }
@@ -1484,16 +1484,22 @@ mod tests {
 
         let buffered_writer = std::io::BufWriter::new(shared_buffer.clone());
 
-        let schema = Schema::parse_str(SCHEMA).unwrap();
+        let schema = Schema::parse_str(SCHEMA)?;
 
         let mut writer = Writer::new(&schema, buffered_writer);
 
         let mut record = Record::new(writer.schema()).unwrap();
         record.put("exampleField", "value");
 
-        writer.append(record).unwrap();
-        writer.flush().unwrap();
+        writer.append(record)?;
+        writer.flush()?;
 
-        assert_eq!(shared_buffer.len(), 167, "the test buffer was not fully written to after Writer::flush was called");
+        assert_eq!(
+            shared_buffer.len(),
+            167,
+            "the test buffer was not fully written to after Writer::flush was called"
+        );
+
+        Ok(())
     }
 }
--- a/avro/src/writer.rs
+++ b/avro/src/writer.rs
@@ -315,6 +315,8 @@ impl<'a, W: Write> Writer<'a, W> {
         self.buffer.clear();
         self.num_values = 0;
 
+        self.writer.flush().map_err(Error::FlushWriter)?;
+
         Ok(num_bytes)
     }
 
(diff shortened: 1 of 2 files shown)


avro/src/writer.rs
```diff
@@ -1444,4 +1448,52 @@ mod tests {
         }
         Ok(())
     }
+
+    #[test]
+    fn avro_4063_flush_applies_to_inner_writer() {
```
```suggestion
    fn avro_4063_flush_applies_to_inner_writer() -> TestResult {
```

avro/src/writer.rs
```diff
@@ -1444,4 +1448,52 @@ mod tests {
...
+        let buffered_writer = std::io::BufWriter::new(shared_buffer.clone());
+
+        let schema = Schema::parse_str(SCHEMA).unwrap();
+
+        let mut writer = Writer::new(&schema, buffered_writer);
+
+        let mut record = Record::new(writer.schema()).unwrap();
+        record.put("exampleField", "value");
+
+        writer.append(record).unwrap();
```
```suggestion
        writer.append(record)?;
```

avro/src/writer.rs
```diff
@@ -1444,4 +1448,52 @@ mod tests {
...
+        writer.flush().unwrap();
```
```suggestion
        writer.flush()?;
```

avro/src/writer.rs
```diff
@@ -1444,4 +1448,52 @@ mod tests {
...
+
+        assert_eq!(shared_buffer.len(), 167, "the test buffer was not fully written to after Writer::flush was called");
```
```suggestion
        assert_eq!(shared_buffer.len(), 167, "the test buffer was not fully written to after Writer::flush was called");

        Ok(())
```

avro/src/writer.rs
```diff
@@ -1444,4 +1448,52 @@ mod tests {
...
+            fn flush(&mut self) -> std::io::Result<()> {
+                Ok(())
+            }
+        }
+
+        let shared_buffer = TestBuffer::default();
+
+        let buffered_writer = std::io::BufWriter::new(shared_buffer.clone());
+
+        let schema = Schema::parse_str(SCHEMA).unwrap();
```
```suggestion
        let schema = Schema::parse_str(SCHEMA)?;
```

avro/src/writer.rs
```diff
@@ -1444,4 +1448,52 @@ mod tests {
...
+            "type": "record",
+            "name": "ExampleSchema",
+            "fields": [
+                {"name": "exampleField", "type": "string"}
+            ]
+        }
+        "#;
+
+        #[derive(Clone, Default)]
+        struct TestBuffer(Arc<Mutex<Vec<u8>>>);
```
martin-g: Why the Arc ?
It is all synchronous here.

martin-g: Replaced it with `Rc<RefCell<Vec>>`

martin-g (approved):

martin-g:
Thank you, @snowsignal !

PR #15, opened 2024-10-03
fix: deserializing ignored enum

xordi:
# Issue description

Ignored enum values cause a deserialization error when using the`from_value` function.

## How to reproduce

This code produces an error saying `incorrect value of type: Enum`:

```rust
struct SampleStruct {
    a: String
}

let record = Value::Record(vec![
    (
        "a".to_string(),
        Value::String("a valid message field".to_string()),
    ),
    (
        "a_non_existing_enum".to_string(),
        Value::Enum(0, "Variant1".to_owned()),
    )
]);

let deserialized: SampleStruct = from_value(&record)?;
```

## Proposal

I see the `deserialize_any` function is basically called only from `deserialize_ignored_any`. 

I might have also made a wrong assumption, but it seems unlikely that `from_value` will be used to deserialize into a type that its `Deserialize` implementation calls `deserialize_any`, as the `Value` passed as a parameter already represents an arbitrary Avro value. For that reason I opted to call the `visit_borrowed_str` on the `visitor` when finding a `Value::Enum`  inside `deserialize_any`, avoiding allocating further memory. I would appreciate any further insight on this though, as I'm not sure if it's the right approach.

### DIFF
--- a/avro/src/de.rs
+++ b/avro/src/de.rs
@@ -272,6 +272,7 @@ impl<'a, 'de> de::Deserializer<'de> for &'a Deserializer<'de> {
                 Value::Map(ref items) => visitor.visit_map(MapDeserializer::new(items)),
                 Value::Bytes(ref bytes) | Value::Fixed(_, ref bytes) => visitor.visit_bytes(bytes),
                 Value::Decimal(ref d) => visitor.visit_bytes(&d.to_vec()?),
+                Value::Enum(_, ref s) => visitor.visit_borrowed_str(s),
                 _ => Err(de::Error::custom(format!(
                     "unsupported union: {:?}",
                     self.input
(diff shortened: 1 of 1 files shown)


martin-g:
> I would appreciate any further insight on this though, as I'm not sure if it's the right approach.

Me too :-/

How is `a_non_existing_enum` supposed to be handled ? It seems it is just being ignored in your example. Why ?

martin-g (approved):

martin-g:
> It seems it is just being ignored in your example. Why ?

I still do not understand why it behaves like that but I see that the test does the same for all other fields, so you just follow the pattern.
If you understand why the extra record fields are being silently ignored please share!

xordi:
Ignoring those fields is on purpose, as I'd like to be able to deserialize the `Value` into a struct with just a subset of the fields that are present in the `Value::Record`. 

`serde` produces a `Deserialize` implementation on the type that accounts for ignored fields. Basically that implementation calls `deserialize_ignored_any` in the `Deserializer` when handling a field that's not present in the target element (a `struct` in this case). I guess the main reason of this is because in some data formats, you need to read the underlying stream of data and ignore it in order to decode the next elements. In this crate case, the `Deserializer` forwards `deserialize_ignored_any` (called by the `serde` implementation of `Deserialize` on the struct) to `deserialize_any`,  where the Enum values are not handled.

I followed the same pattern as for the other `Value` types in this fix, as this is actually implemented for mostly all of them except for `Enum`. I wonder if there was any reason why this behavior was not replicated for enums. I don't see any, but of course I might lack a lot of context.

Maybe another option not impacting this crate would be to manually implement `Deserialize` for a struct in which we just want to deserialize certain fields from the `Value::Record`, but given that the `Deserializer` implementation was already ignoring fields of other types, I opted to implement the same for enums.

PR #20, opened 2024-10-13
fix: all links to the documentation

baltendo:


martin-g:
Thank you for the PR!
But I think the problem is in /apache/avro > doc/. Those should be aliases and both links should work.

baltendo:
Ah ok, thanks!

PR #49, opened 2024-11-08
feat: Support (de)serializing BigDecimal values

martin-g:


avro/src/de.rs
```diff
@@ -367,6 +369,7 @@ impl<'a, 'de> de::Deserializer<'de> for &'a Deserializer<'de> {
             }
             Value::Uuid(ref u) => visitor.visit_bytes(u.as_bytes()),
             Value::Decimal(ref d) => visitor.visit_bytes(&d.to_vec()?),
+            Value::BigDecimal(ref d) => visitor.visit_string(d.to_string()),
```
ollis, line 372: This will never match because `deserialize_bytes` is not invoked for bigdecimal? It is either `deserialize_any` or `deserialize_str` (if feature `string-only` is enabled)?

avro/src/ser.rs
```diff
@@ -691,11 +694,13 @@ mod tests {
             a: 27,
             b: "foo".to_owned(),
             decimal: Decimal::from(vec![1, 24]),
+            big_decimal: BigDecimal::new(BigInt::from(12), 2),
         };
         let expected = Value::Record(vec![
             ("a".to_owned(), Value::Long(27)),
             ("b".to_owned(), Value::String("foo".to_owned())),
             ("decimal".to_owned(), Value::Bytes(vec![1, 24])),
+            ("big_decimal".to_owned(), Value::String("0.12".into())),
```
ollis, line 703: The deserialization tests can in some cases use a different Value type than what is expected i the serialization test, e.g. for decimal/big_decimal. Maybe one should add tests that verifies a serialization-deserialization roundtrip: struct = fromValue(toValue(struct)) (It works now because both Value::String and Value::BigDecimal ends up calling the `visit_str` on BigDecimal)

avro/src/de.rs
ollis: (So it has to be added to `deserialize_str` in order to get `string-only` to work as expected)

martin-g: Thanks for the review!
Feel free to open an alternative PR!

Issue #64, opened 2024-12-02
Schema representation - 'importing' types


chupaty:
The specification states the following:

> A schema or protocol may not contain multiple definitions of a fullname. Further, a name must be defined before it is used (“before”  in the depth-first, left-to-right traversal of the JSON parse tree, where the types attribute of a protocol is always deemed > to come “before” the messages attribute.)

This means that the following schema is valid (the A type is defined in b_field_one and the type is not defined with b_field_two):
```
{
  "name":"B",
  "type":"record",
  "fields":[
    {
      "name":"b_field_one",
      "type":{"name":"A","type":"record","fields":[]}
    },
    {
      "name":"b_field_two",
      "type":{"name":"A"}
    },    
  ]
}
```

This crate makes it easy to load a schemata in the form of a sequence (Schema::parse_list) of schema where:
- The ordering rule is relaxed (no need to load the schemata in a particular order)
- The defined-once rule is applied to the whole sequence, instead of per schema

However I can't find a way to output an individual schema in a schemata that is 'complete' (ie does not depend on other schemas, and otherwise complies with the rules).  For example:
```
   let schema_str_1 = r#"{
        "name": "A",
        "doc": "A's schema",
        "type": "record",
        "fields": [
        ]
    }"#;
    let schema_str_2 = r#"{
        "name": "B",
        "doc": "B's schema",
        "type": "record",
        "fields": [
            {"name": "b_field_one", "type": "A"},
            {"name": "b_field_two", "type": "A"}
        ]
    }"#;
    let schema_strs = [schema_str_1, schema_str_2];
    let schemata = Schema::parse_list(&schema_strs)?;

    for d in schemata {
        println!("{}",d.canonical_form());
    }
```

Gives the following canonical schemas (for A I think this is OK, B is problematic as there is no definition for A):
```
{"name":"A","type":"record","fields":[]}
{"name":"B","type":"record","fields":[{"name":"b_field_one","type":"A"},{"name":"b_field_two","type":"A"}]}
```

I believe the canonical form for B should actually be:
```
{"name":"B","type":"record","fields":[{"name":"b_field_one","type":{"name":"A","type":"record","fields":[]}},{"name":"b_field_two","type":"A"}]}
```

Do we have a way of producing this 'correct' form? This is necessary for fingerprint calculation and some schema registry interactions.

chupaty:
I've raised a PR to provide this here: https://github.com/apache/avro-rs/pull/66

[...]

PR #66, opened 2024-12-03
Added Schema::independent_canonical_form

chupaty:
Added independent_canonical_form, which populates types available in the schemata, so that the given schema can be encoded for PCF without the schemata.

See https://github.com/apache/avro-rs/issues/64

### DIFF
--- a/avro/src/schema.rs
+++ b/avro/src/schema.rs
@@ -1266,15 +1266,12 @@ impl Schema {
         match self {
             Ref { name } => {
                 let repl = schemata.iter().find(|s| {
-                    if let Some(n) = s.name() {
-                        if *n == *name {
-                            return true;
-                        }
-                    }
-                    false
+                    s.name().map(|n| *n == *name).unwrap_or(false)
                 });
                 if let Some(r) = repl {
-                    *self = r.clone();
+                    let mut denorm = r.clone();
+                    denorm.denormalize(schemata);
+                    *self = denorm;
                 }
             }
             Schema::Record(r) => {
--- a/avro/src/schema.rs
+++ b/avro/src/schema.rs
@@ -43,8 +43,6 @@ use std::{
     io::Read,
     str::FromStr,
 };
-#[allow(unused_imports)]
-use std::borrow::Cow;
 use strum_macros::{Display, EnumDiscriminants, EnumString};
 
 /// Represents an Avro schema fingerprint
@@ -1265,9 +1263,9 @@ impl Schema {
     fn denormalize(&mut self, schemata: &Vec<Schema>) {
         match self {
             Ref { name } => {
-                let repl = schemata.iter().find(|s| {
-                    s.name().map(|n| *n == *name).unwrap_or(false)
-                });
+                let repl = schemata
+                    .iter()
+                    .find(|s| s.name().map(|n| *n == *name).unwrap_or(false));
                 if let Some(r) = repl {
                     let mut denorm = r.clone();
                     denorm.denormalize(schemata);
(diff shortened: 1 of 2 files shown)


chupaty:
Closing for now - will reopen when tested on production-like schemas

martin-g:
The test failure is not caused by this PR. It is an intermittent error that needs some debugging.

martin-g:
Fixed with https://github.com/apache/avro-rs/pull/72

avro/src/schema.rs
```diff
@@ -1202,6 +1215,38 @@ impl Schema {
...
         })
     }
+
+    fn denormalize(&mut self, schemata: &Vec<Schema>) {
+        match self {
+            Ref { name } => {
+                let repl = schemata
+                    .iter()
+                    .find(|s| s.name().map(|n| *n == *name).unwrap_or(false));
+                if let Some(r) = repl {
```
martin-g: What about the case when the Ref is not found in the provided `schemata` ?

chupaty: Error::Inconclusive?  Or shall I create a new error?

chupaty: SchemaResolutionError - added a test for this as well

avro/src/schema.rs
```diff
@@ -3238,7 +3304,7 @@ mod tests {
         assert_eq!(schema, expected);
 
         let canonical_form = &schema.canonical_form();
-        let expected = r#"{"name":"record","type":"record","fields":[{"name":"enum","type":{"name":"enum","type":"enum","symbols":["one","two","three"]}},{"name":"next","type":{"name":"enum","type":"enum","symbols":["one","two","three"]}}]}"#;
+        let expected = r#"{"name":"record","type":"record","fields":[{"name":"enum","type":{"name":"enum","type":"enum","symbols":["one","two","three"]}},{"name":"next","type":"enum"}]}"#;
```
martin-g, line 3492: What happened with the symbols of `"name":"next"` ?

avro/src/schema.rs
```diff
@@ -3322,7 +3388,7 @@ mod tests {
         assert_eq!(schema, expected);
 
         let canonical_form = &schema.canonical_form();
-        let expected = r#"{"name":"record","type":"record","fields":[{"name":"fixed","type":{"name":"fixed","type":"fixed","size":456}},{"name":"next","type":{"name":"fixed","type":"fixed","size":456}}]}"#;
+        let expected = r#"{"name":"record","type":"record","fields":[{"name":"fixed","type":{"name":"fixed","type":"fixed","size":456}},{"name":"next","type":"fixed"}]}"#;
```
martin-g, line 3576: Same here

avro/src/schema.rs
chupaty: This test was originally added with the update to the spec clarifying the use of certain names (ie 'enum' and 'fixed' for this and the following test). 

https://github.com/apache/avro/pull/1573/files

This test refers to a field named 'enum' with a type of 'enum' - fine.  The change here is necessary because the old canonical_form() incorrectly defined the type _twice_, which is not allowed by the spec:

> A schema or protocol may not contain multiple definitions of a fullname. Further, a name must be defined before it is used ("before" in the depth-first, left-to-right traversal of the JSON parse tree, where the types attribute of a protocol is always deemed to come "before" the messages attribute.)

canonical_form() now tracks the types as the schema is traversed, so it trims out the already defined type to meet the define-once rule.

martin-g (approved):

martin-g:
Thank you, @chupaty !

Issue #103, opened 2025-01-17
Parsing Avro unions with schema references


rayokota:
The docs state that the `Schema::parse_list` can parse a set of schemas that depend on one another,  but all schemas in the set need to be named.

How would I parse a set of schemas where one schema is a union that depends on other named schemas such as the following?

```
        let union = r#"["myschema1", "myschema2"]"#;

        let myschema1 = r#"
        {
            "type": "record",
            "name": "myschema1",
            "fields": [ {"name": "name", "type": "string"} ]
        }
        "#;

        let myschema2 = r#"
        {
            "type": "record",
            "name": 'myschema2',
            "fields": [ {"name": "age", "type": "int"} ]
        }
        "#;
```

It seems the  `Schema::parse_with_names(value: &Value, names: Names)` method would work for this use case, but it is not currently public.  Would it be possible to make that method public?

martin-g:
Please send a Pull Request with this snippet as a test case and making the method public!

rayokota:
Thanks @martin-g , I added a new method `Schema:: parse_str_with_list` that can be used to parse a root schema with a list of named schemas.   https://github.com/apache/avro-rs/pulls

PR #104, opened 2025-01-18
Add method `Schema::parse_str_with_list` to parse root schemas with references

rayokota:
Add a new method `Schema::parse_str_with_list` that can parse a root schema along with a list of named schemas.  This method can be used to parse a root union that has named schema references, for example.

This new method combines the functionality of `Schema::parse_str` with `Schema::parse_list`.

Fixes https://github.com/apache/avro-rs/issues/103

### DIFF
--- a/avro/src/schema.rs
+++ b/avro/src/schema.rs
@@ -1104,15 +1104,18 @@ impl Schema {
     /// during parsing.
     ///
     /// If two of the named input schemas have the same fullname, an Error will be returned.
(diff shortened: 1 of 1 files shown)


martin-g (approved):

rayokota:
Thanks for the quick merge @martin-g !  I realized as I went to sleep last night that I should extend the return type to also returned the parsed named schemas:

https://github.com/apache/avro-rs/pull/105

PR #105, opened 2025-01-18
Minor enhancement to `parse_str_with_list`

rayokota:
A minor enhancement to `parse_str_with_list` to return the parsed named schemata.  This is to make it easier to use with `to_avro_datum_schemata` and `from_avro_datum_schemata`, both of which take a root schema and a Vector of named schemas.

This is a follow-on PR to https://github.com/apache/avro-rs/pull/104

### DIFF
--- a/avro/src/schema.rs
+++ b/avro/src/schema.rs
@@ -1108,7 +1108,7 @@ impl Schema {
     /// # Arguments
     /// * `schema` - the JSON string of the schema to parse
     /// * `schemata` - a slice of additional schemas that is used to resolve cross-references
-    pub fn parse_str_with_list(schema: &str, schemata: &[&str]) -> AvroResult<Schema> {
+    pub fn parse_str_with_list(schema: &str, schemata: &[&str]) -> AvroResult<(Schema, Vec<Schema>)> {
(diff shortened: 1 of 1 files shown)


martin-g (approved):

avro/src/schema.rs
```diff
@@ -2761,7 +2766,44 @@ mod tests {
 
         let schema_str_c = r#"["A", "B"]"#;
 
-        let schema_c = Schema::parse_str_with_list(schema_str_c, &[schema_str_a, schema_str_b])?;
+        let (schema_c, schemata) =
+            Schema::parse_str_with_list(schema_str_c, &[schema_str_a, schema_str_b])?;
+
+        let schema_a_expected = Schema::Record(RecordSchema {
```
martin-g, line 2772: I think it would be a good idea to start using a builder for the non-primitive schema types and RecordFiled. There are many fields which have good defaults which could be omitted here. 
I'll file an issue!

martin-g:
Thank you, @rayokota !

rayokota:
Thanks @martin-g for the quick response!

avro/src/schema.rs
martin-g: https://github.com/apache/avro-rs/pull/108

PR #106, opened 2025-01-20
Add a "from" datum reader that can take reader schemata

rayokota:
Add a "from" datum reader that can take reader schemata. 

Currently the `from_avro_datum_schemata` method can take writer schemata so that references in the writer schema can be resolved correctly, but it does not take reader schemata so that references in the reader schema can be resolved correctly.

I've added a new method `from_avro_datum_reader_schemata` that can take both writer and reader schemata.  (Feel free to change the name of this method)


martin-g:
Thank you, @rayokota !

rayokota:
Thanks for the quick response @martin-g !

Issue #70, opened 2024-12-04
Find a way to use the provided Schema when using "the serde way"


martin-g:
See https://github.com/apache/avro-rs/blob/d6a1e4f5b66d429a7e6eb73a7679083c0e9f4d45/avro/src/schema.rs#L6446-L6482

Here the test uses 
```

 SpecificSingleObjectWriter::<Comment>::with_capacity(10)?
            .write_ref(&Comment {id: "de2df598-9948-4988-b00a-a41c0e287398".parse()?}, &mut buffer)?;
```

The Avro Schema is available via `Comment::get_schema()` but it is not used at https://github.com/apache/avro-rs/blob/d6a1e4f5b66d429a7e6eb73a7679083c0e9f4d45/avro/src/writer.rs#L551-L555 because `Serde` does not work with schemas.

We need to find a way to make use of the schema because now the `id` struct field is serialized/encoded as a `types::Value::String` instead of `Uuid` or even `Fixed(16)`, as defined in the schema.

jdarais:
Hi, thanks for your great work on this library.  I started playing around with avro/rust, and ended up creating a PR (#139) that proposes some improvements, which would fix this issue as well.  I think the crux of the matter is that the schema isn't being used by the serializer.  You're right that `Serde` doesn't work with schemas, and no schema information embedded in the type being serialized would be exposed to the serializer, but you can implement a `Serializer` that contains a schema to reference when serializing any value.  That's the approach that my PR takes, though you could likely do the same with a `Serializer` that serializes to a a `Value`.

PR #139, opened 2025-03-02
Direct serializer for "Serde Way"

jdarais:
Hi, I started playing with avro-rs to build some Rust-based Avro tools, and noticed some things I could take a stab at improving in the avro-rs library itself.  This PR has a lot of changes in it, but it mostly centers around a new `DirectSerializer`, who's behavior is described below.  Feel free to consider this PR as a whole, or take bits and pieces from it.  There are some things in it that could be useful even independently of the addition of `DirectSerializer`:

This PR proposes a direct serialization of objects when serializing the "Serde Way", which provides roughly a 5x performance improvement, and fixes issues caused by lost schema information when first converting to a `Value` before serializing to the Write stream (#70).

Some notes on these changes:
- An new Serializer implementation, `DirectSerializer`, provides direct serialization of a type using the "Serde Way".  The `DirectSerializer` is initialized with the schema to be used for writing data, and uses that schema as a guide to know how to serialize the serde types that it encounters.
- The existing serde benchmark tests seemed to only test serialization of `Value` types, which is only part of the process in the "Serde Way", so some additional benchmark tests were added to measure the end-to-end performance of "Serde Way" serialization
- The deserializer appears to be missing some "Avro union / Rust enum" (de)serialization capability.  In the existing implementation it appears that Rust enums can be (de)serialized to/from Avro union types, but only if the `#[serde(untagged)]` attribute is added, which serializes an enum value as the variant type itself rather than a union type, and it's up to the (de)serializer to detect that the schema calls for a union, search the union schemas for a match, and use the first schema found that matches the data.  This PR adds a `UnionDeserializer` to `de.rs`, which, in combination with `DirectSerializer`, allows "Avro union / Rust enum" (de)serialization without requiring the `#[serde(untagged)]` attribute.
- I was able to replicate almost all behavior of the existing implementation using `Value` as an intermediate step, but one thing I was not able to replicate is serialization of enums that do have the `#[serde(untagged)]` attribute.  The closest I got was making the `DirectSerializer` able to search a union schema for the first schema _type_ that matches the input data, but in the case of "record" types, there's no easy way to ensure that the field names are also a match.  This feature of schema matching on write does seem a bit odd to me, since the writer should know the structure of the data being written, and schema matching should only really be a concern for the reader, where schema resolution is happening between a reader and writer schema.  But, worth noting that this difference in behavior is not backwards-compatible.

All unit tests are passing.  You can take a look at the changes in the unit tests to see where there are differences in behavior.


martin-g:
Thanks for the PR!
Please resolve the merge conflicts!

jdarais:
Oops, I hadn't realized my branch had fallen behind that much.  Merge conflicts should be resolved now.

avro/src/de.rs
```diff
@@ -55,6 +55,11 @@ pub struct EnumDeserializer<'de> {
     input: &'de [(String, Value)],
 }
 
+struct UnionDeserializer<'de> {
```
martin-g, line 69: Please add some Rustdoc.
I know that the other structs here may not have docs, but this is one of the reasons why the SerDe functionality is not well maintained in Avro.
- what is this struct / why is it needed
- what are its field for
- etc.

avro/src/de.rs
```diff
@@ -230,6 +241,60 @@ impl<'de> de::VariantAccess<'de> for EnumDeserializer<'de> {
     }
 }
 
+impl<'de> de::EnumAccess<'de> for UnionDeserializer<'de> {
+    type Error = Error;
+    type Variant = Self;
```
martin-g, line 246: I am not sure this type is helpful here.
It usage below (Self::Variant) makes it more confusing.
Can we use just `Self` ?!

avro/src/ser_direct.rs
```diff
@@ -0,0 +1,2032 @@
...
+    error::Error,
+    schema::{Name, NamesRef, Namespace, RecordSchema, Schema},
+};
+
+const RECORD_FIELD_INIT_BUFFER_SIZE: usize = 64;
+const COLLECTION_SERIALIZER_ITEM_LIMIT: usize = 1024;
+const COLLECTION_SERIALIZER_DEFAULT_INIT_ITEM_CAPACITY: usize = 32;
+const SINGLE_VALUE_INIT_BUFFER_SIZE: usize = 128;
+
+pub struct DirectSerializeSeq<'a, 's, W: Write> {
```
martin-g, line 42: Rustdoc please for all public items!
With as much as possible details because even the current ser.rs and de.rs are hard to maintain.

avro/src/de.rs
jdarais: That's a good call.  I've added some rustdoc in the areas you pointed out.  Let me know if you'd like me to add more detail.

Thanks for adding those additional tests and cleanup, btw.

martin-g:
@jdarais Thanks for the Rustdoc updates!

I've added some `TODO`s in DirectSerializer's tests. Could you please take a look and either improve or answer why the current code is OK. Thanks!

martin-g:
@jdarais I just pushed a failing test with a recursive RecordSchema. Could you please check what is the problem ? Thanks!

avro/src/ser_direct.rs
```diff
@@ -0,0 +1,2505 @@
...
+            "logicalType": "big-decimal"
+        }"#,
+        )?;
+
+        let mut buffer: Vec<u8> = Vec::new();
+        let names = HashMap::new();
+        let mut serializer = DirectSerializer::new(&mut buffer, &schema, &names, None);
+
+        let val = BigDecimal::new(BigInt::new(Sign::Plus, vec![50024]), 2);
+        ByteBuf::from(big_decimal_as_bytes(&val)?).serialize(&mut serializer)?;
```
martin-g: `big_decimal_as_bytes` is a `pub(crate)` method at the moment.
Is there a way to move the ` ByteBuf::from(big_decimal_as_bytes(&val)?)` in the non-test code so that `val.serialize(&mut serializer)` could work ?

Ideally a struct like:
```rust
#[serde(rename_all = "camelCase")]
struct TestRecord {
  string_field: String,
  int_field: i32,
  big_decimal_field: BigDecimal,
}
``` 
should "Just work"

martin-g:
Idea: 
https://github.com/lerouxrgd/rsgen-avro generates Rust structs/enums from Avro Schema.
It would be interesting to try to serialize those structs/enums with the new DirectSerializer.
CC @lerouxrgd

avro/src/ser_direct.rs
martin-g: Done with https://github.com/apache/avro-rs/pull/139/commits/35f5ccfa592d32a32d54b1a42e578f1f448bbb46

jdarais:
> @jdarais I just pushed a failing test with a recursive RecordSchema. Could you please check what is the problem ? Thanks!
> 
> Update: I think I fixed it with [a364ac9](https://github.com/apache/avro-rs/commit/a364ac90a4866625189af8e87fb1b9de829e5929)

Is that recursive schema a legal one, (or at least able to be non-infinite)?  For the recursive `innerRecord` field in the `TestRecord` schema, I think it would need to be `{"name": "innerRecord", "type": ["null", "TestRecord"]}` in order to accommodate the terminal record that has `None` for its `innerRecord` field.  With that modification to the schema, the object created in `test_serialize_recursive_record` should be able to be serialized without errors.

martin-g:
You might be right!
Please push a fix!

avro/src/ser_direct.rs
jdarais: It'd be great if `BigDecimal` "just worked", though I don't think the `Serializer` will be able to just call `to_bytes()` on the value passed into `serialize_str` for `BigDecimal` values.  From what I can see, the implementation of `Serialize` for `bigdecimal::BigDecimal` ends up passing a display string representation of the value (e.g. `"12.75"`) to the `Serializer`'s `serialize_str` function, not the byte representation of the `BigDecimal` value.

A few approaches I can think of to make this work more smoothly for `BigDecimal` values:
1. In `serialize_str`, if the schema type is `BigDecimal`, assume we've been given a display string representation of the decimal value.  Parse the string back into a `BigDecimal`, then pass that into `big_decimal_as_bytes` to get the byte sequence to write to the `Write` stream.  Pro: it serializes `bigdecimal::BigDecimal` out of the box.  Con: it incurs a performance penalty for having to re-parse the decimal string.
2. Define an Avro-aware wrapper struct that wraps around a `bigdecimal::BigDecimal` and implements `Serialize` in a way that passes an Avro-compatible bytes representation into the `Serializer`'s `serialize_bytes` function.  Pro: faster since it doesn't need to re-parse anything.  Con: structs intended to be written to Avro using a schema that contains `big-decimal` must use the wrapper struct for the value to be serialized successfully.
3. Same as (2), but instead of the wrapper struct passing `bytes` to the `Serializer`, pass a tuple/struct with the separate mantissa and scale components of the `BigDecimal` value, and leave it to the `Serializer` to produce the correct byte representation from those values.  More flexible than (2) since others could implement `Serialize` for structs that wrap other "big decimal" types (like `rust_decimal::Decimal`), and likely not much slower than (2)

In fact, we could have `DirectSerializer` support all of these options, and the option chosen is determined by the type being serialized, whether it calls `serialize_str`, `serialize_bytes`, `serialize_struct`, or `serialize_tuple`.  I'll see if I can get this implemented.

jdarais:
> You might be right! Please push a fix!

Will do 👍

avro/src/ser_direct.rs
jdarais: Ok, option (3) is looking a bit more involved, and option (2) is still supported if someone wants to create their own wrapper struct that converts the `BigDecimal` to bytes first and then passes those to the `Serializer` via `serialize_bytes`.  Option (1) is the easiest to do, so that's implemented now.

Now that I think about it, it may be possible to provide alternative `Serialize` implementations for `BigDecimal` and allow consumers to select which one to use with the `#[serde(with = ...)]` attribute as described [here](https://serde.rs/field-attrs.html#with).  I'll leave that as future work, though.  What we have here should be sufficient to start.

jdarais:
Ok, recursive schema and big decimal "just works" test cases should be updated and passing now.  Thanks for adding those test cases!

jdarais:
Oops, forgot to push.  _Now_ recursive schema and big decimal updates should be there.

martin-g:
@jdarais I don't have more ideas for improvements. 
Please take a final look and if you agree I will merge the PR! 
Thank you!

avro/src/ser_direct.rs
```diff
@@ -499,7 +498,7 @@ impl<'s, W: Write> DirectSerializer<'s, W> {
             root_schema: schema,
             names,
             enclosing_namespace,
-            schema_stack: Vec::new(),
+            schema_stack: Default::default(),
```
jdarais, line 501: You may want to run the benchmark tests to validate whether a linked list is faster in this case.  When running on my own machine, Vec is faster.  A vector may need to occasionally reallocate if its capacity is reached, but when it's used as a stack, inserts and deletes only happen at the end, which happens in constant time, (except for the scenario where capacity is exceeded and a larger space must be allocated, but those should be infrequent,) and you get the benefit of cache locality since all the elements are next to each other in memory.

jdarais:
Just had a comment on the container type to store the `schema_stack`, let me know what you think.  Otherwise, everything else looks good to me!

avro/src/ser_direct.rs
martin-g: For me it shows small improvements with LinkedList:
```
cargo bench serde
   Compiling apache-avro v0.18.0 (/home/martin/git/apache/avro-rs/avro)
   Compiling apache-avro-derive v0.18.0 (/home/martin/git/apache/avro-rs/avro_derive)
   Compiling hello-wasm v0.1.0 (/home/martin/git/apache/avro-rs/wasm-demo)
    Finished `bench` profile [optimized] target(s) in 12.52s
     Running benches/serde.rs (target/release/deps/serde-cc1227daf82840e0)
small schema, write 1 record (serde way)
                        time:   [2.3658 µs 2.4030 µs 2.4490 µs]
                        change: [-4.4973% -2.9172% -1.2806%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 10 outliers among 100 measurements (10.00%)
  6 (6.00%) high mild
  4 (4.00%) high severe

small schema, write 100 records (serde way)
                        time:   [10.437 µs 10.519 µs 10.620 µs]
                        change: [-5.0968% -2.9766% -0.9836%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 15 outliers among 100 measurements (15.00%)
  6 (6.00%) high mild
  9 (9.00%) high severe

big schema, write 1 record (serde way)
                        time:   [4.6004 µs 4.6281 µs 4.6591 µs]
                        change: [-10.071% -7.9251% -5.8801%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 10 outliers among 100 measurements (10.00%)
  5 (5.00%) high mild
  5 (5.00%) high severe

big schema, write 100 records (serde way)
                        time:   [41.988 µs 42.455 µs 43.000 µs]
                        change: [-1.8518% -1.0529% -0.1373%] (p = 0.02 < 0.05)
                        Change within noise threshold.
Found 13 outliers among 100 measurements (13.00%)
  4 (4.00%) high mild
  9 (9.00%) high severe

small schema, write 10k records (serde way)
                        time:   [822.03 µs 826.87 µs 832.20 µs]
                        change: [+0.4821% +1.4965% +2.4936%] (p = 0.01 < 0.05)
                        Change within noise threshold.

big scheam, write 10k records (serde way)
                        time:   [3.6829 ms 3.7066 ms 3.7332 ms]
                        change: [-4.5266% -3.5488% -2.6061%] (p = 0.00 < 0.05)
                        Performance has improved.
```

Going back to Vec:
```
cargo bench serde
   Compiling apache-avro v0.18.0 (/home/martin/git/apache/avro-rs/avro)
   Compiling apache-avro-derive v0.18.0 (/home/martin/git/apache/avro-rs/avro_derive)
   Compiling hello-wasm v0.1.0 (/home/martin/git/apache/avro-rs/wasm-demo)
    Finished `bench` profile [optimized] target(s) in 15.77s
     Running benches/serde.rs (target/release/deps/serde-cc1227daf82840e0)
small schema, write 1 record (serde way)
                        time:   [2.3788 µs 2.3980 µs 2.4242 µs]
                        change: [+5.6754% +7.4976% +9.3008%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 9 outliers among 100 measurements (9.00%)
  3 (3.00%) high mild
  6 (6.00%) high severe

small schema, write 100 records (serde way)
                        time:   [10.412 µs 10.479 µs 10.554 µs]
                        change: [-0.5517% +0.4947% +1.6596%] (p = 0.41 > 0.05)
                        No change in performance detected.
Found 12 outliers among 100 measurements (12.00%)
  5 (5.00%) high mild
  7 (7.00%) high severe

big schema, write 1 record (serde way)
                        time:   [4.6428 µs 4.6718 µs 4.7082 µs]
                        change: [+3.5714% +5.4697% +7.5133%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 13 outliers among 100 measurements (13.00%)
  4 (4.00%) high mild
  9 (9.00%) high severe

big schema, write 100 records (serde way)
                        time:   [41.769 µs 41.981 µs 42.247 µs]
                        change: [+0.9885% +2.1415% +3.5064%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 12 outliers among 100 measurements (12.00%)
  5 (5.00%) high mild
  7 (7.00%) high severe

small schema, write 10k records (serde way)
                        time:   [812.16 µs 818.92 µs 826.53 µs]
                        change: [-3.4624% -2.2517% -1.0241%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 20 measurements (10.00%)
  1 (5.00%) high mild
  1 (5.00%) high severe

big scheam, write 10k records (serde way)
                        time:   [3.6963 ms 3.7240 ms 3.7561 ms]
                        change: [-3.6033% -2.6406% -1.6162%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 20 measurements (5.00%)

```

and again LinkedList:
```
cargo bench serde
   Compiling apache-avro v0.18.0 (/home/martin/git/apache/avro-rs/avro)
   Compiling apache-avro-derive v0.18.0 (/home/martin/git/apache/avro-rs/avro_derive)
   Compiling hello-wasm v0.1.0 (/home/martin/git/apache/avro-rs/wasm-demo)
    Finished `bench` profile [optimized] target(s) in 11.19s
     Running benches/serde.rs (target/release/deps/serde-cc1227daf82840e0)
small schema, write 1 record (serde way)
                        time:   [2.2771 µs 2.2932 µs 2.3141 µs]
                        change: [-8.0503% -6.2473% -4.6485%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 9 outliers among 100 measurements (9.00%)
  4 (4.00%) high mild
  5 (5.00%) high severe

small schema, write 100 records (serde way)
                        time:   [10.690 µs 10.806 µs 10.940 µs]
                        change: [+0.1857% +1.6209% +2.9866%] (p = 0.02 < 0.05)
                        Change within noise threshold.
Found 9 outliers among 100 measurements (9.00%)
  5 (5.00%) high mild
  4 (4.00%) high severe

big schema, write 1 record (serde way)
                        time:   [4.6742 µs 4.7222 µs 4.7790 µs]
                        change: [-3.0179% -0.6372% +1.7778%] (p = 0.62 > 0.05)
                        No change in performance detected.
Found 10 outliers among 100 measurements (10.00%)
  3 (3.00%) high mild
  7 (7.00%) high severe

big schema, write 100 records (serde way)
                        time:   [41.741 µs 42.020 µs 42.347 µs]
                        change: [-2.2702% -0.7627% +0.7781%] (p = 0.33 > 0.05)
                        No change in performance detected.
Found 6 outliers among 100 measurements (6.00%)
  2 (2.00%) high mild
  4 (4.00%) high severe

small schema, write 10k records (serde way)
                        time:   [826.20 µs 830.19 µs 835.08 µs]
                        change: [+0.2122% +1.3489% +2.3329%] (p = 0.02 < 0.05)
                        Change within noise threshold.

big scheam, write 10k records (serde way)
                        time:   [3.7003 ms 3.7235 ms 3.7506 ms]
                        change: [-0.8469% +0.2209% +1.1796%] (p = 0.68 > 0.05)
                        No change in performance detected.

```

martin-g:
I will merge it!
We can make improvements in follow-ups!
Huge thank you for this contribution, @jdarais !

martin-g:
I wonder whether SchemaAwareSerializer is a better name ?!

The indirection thru types::Value was an implementation detail in the old serializer. IMO the real value of the new one is the schema validation! The speed improvements is very nice too!

jdarais:
> I wonder whether SchemaAwareSerializer is a better name ?!

Yeah it did cross my mind that the `DirectSerializer` using a schema to serialize the input is one of its distinguishing features, though I think it's also a distinguishing feature that it serializes directly to a `Write` stream, since the other `ser::Serializer` implementation "serializes" to a `Value`, not a `Write` stream.  That second distinguishing feature is what led me to the name `DirectSerializer`, though I realize that it's ambiguous what the "Direct" in the name is referring two.  Perhaps `WriteSerializer` would be a better name for that purpose.  Or `SchemaAwareWriteSerializer`?  (Another possibility for indicating usage of a schema would be `SchemaBased`.)  I'm not too opinionated on the name, though, just throwing out ideas.

avro/src/ser_direct.rs
jdarais: Ah, interesting.  It may be platform dependent, since I initially ran the bench tests on Windows, but after running the tests on my Linux machine, I get results consistent with yours.

I was a bit surprised to see that result, since the conventional wisdom is that linked list is almost always worse than vector, since it ends up being more spread out in memory, incurring a performance penalty simply from dealing with CPU cache misses.  Funny enough, looking again at the tests and the `DirectSerializer` implementation, I don't think anything is ever stored in the `schema_stack` field during the bench tests, since it is only used for unions and type references, which aren't used in the bench test data.  The linked list is probably being measured as faster because it costs less to construct an empty linked list vs an empty vector.

I may do another pass to see if I can improve the use of the `schema_stack` field.  If I can further improve the performance I'll submit another PR.

jdarais:
Thanks for taking the PR!

PR #158, opened 2025-03-12
Issue #139: Rename DirectSerializer to SchemaAwareWriteSerializer

martin-g:
~Deprecate the old Serializer impl (apache_avro::ser::Serializer)~


martin-g:
I wonder whether the deprecation is a good idea at the moment.
I guess some users need to serialize to `types::Value`. And the second reason is that `apache_avro::de::Deserializer` at the moment works only with `types::Value`.

PR #163, opened 2025-03-17
Remove schema_stack field from SchemaAwareWriteSerializer

jdarais:
This is a follow-up from my last PR: `SchemaAwareWriteSerializer` doesn't need the schema_stack field since it can make use of the call stack to track which schema it's currently using.  Making this switch makes the schema tracking a bit more consistent, whereas before, sometimes the schema_stack was used, and sometimes a new `SchemaAwareWriteSerializer` as created.  This switch also improves performance slightly, particularly when the serializer encounters union or reference values.

Changes in this PR:
- Remove schema_stack field in favor of `serialize_*_with_schema` member functions
- Benchmark tests updated to include some union fields to measure performance of `serialize_*_with_schema` logic
- Rename support classes to match SchemaAwareWriteSerializer

### DIFF
--- a/avro/benches/serde.rs
+++ b/avro/benches/serde.rs
@@ -192,10 +192,19 @@ fn make_big_record() -> anyhow::Result<(Schema, Value)> {
 
     let big_record = {
         let mut big_record = Record::new(&big_schema).unwrap();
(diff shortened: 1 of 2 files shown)


martin-g (approved):

martin-g:
Thank you, @jdarais !

PR #169, opened 2025-04-01
Use miniz_oxide - a better DEFLATE backend

Shnatsel:
Switch from [libflate](https://crates.io/crates/libflate) to [miniz_oxide](https://crates.io/crates/miniz_oxide) as the DEFLATE backend.

miniz_oxide is faster and more widely tested. It is the default backend for the de-facto standard [flate2](https://crates.io/crates/flate2) crate, and has 10x the amount of downloads compared to libflate.

I've taken care to preserve the same public API so that this change would be transparent to users.

### DIFF
--- a/avro/src/codec.rs
+++ b/avro/src/codec.rs
@@ -17,7 +17,7 @@
 
 //! Logic for all supported compression codecs in Avro.
 use crate::{types::Value, AvroResult, Error};
-use libflate::deflate::{Decoder, Encoder};
+use miniz_oxide;
 use std::io::{Read, Write};
 use strum_macros::{EnumIter, EnumString, IntoStaticStr};
 
@@ -61,13 +61,8 @@ impl Codec {
         match self {
             Codec::Null => (),
             Codec::Deflate => {
-                let mut encoder = Encoder::new(Vec::new());
-                encoder.write_all(stream).map_err(Error::DeflateCompress)?;
-                // Deflate errors seem to just be io::Error
-                *stream = encoder
-                    .finish()
-                    .into_result()
-                    .map_err(Error::DeflateCompressFinish)?;
+                let compressed = miniz_oxide::deflate::compress_to_vec(&stream, 6);
+                *stream = compressed;
             }
             #[cfg(feature = "snappy")]
             Codec::Snappy => {
@@ -120,14 +115,8 @@ impl Codec {
     pub fn decompress(self, stream: &mut Vec<u8>) -> AvroResult<()> {
         *stream = match self {
             Codec::Null => return Ok(()),
-            Codec::Deflate => {
-                let mut decoded = Vec::new();
-                let mut decoder = Decoder::new(&stream[..]);
-                decoder
-                    .read_to_end(&mut decoded)
-                    .map_err(Error::DeflateDecompress)?;
-                decoded
-            }
+            Codec::Deflate => miniz_oxide::inflate::decompress_to_vec(&stream)
+                .map_err(Error::DeflateDecompress)?,
             #[cfg(feature = "snappy")]
             Codec::Snappy => {
                 let decompressed_size = snap::raw::decompress_len(&stream[..stream.len() - 4])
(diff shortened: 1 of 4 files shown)


avro/src/error.rs
```diff
@@ -377,6 +377,8 @@ pub enum Error {
     #[error("Failed to compress with flate: {0}")]
     DeflateCompress(#[source] std::io::Error),
 
+    // no longer possible after migration from libflate to miniz_oxide
+    // TODO: remove in the next semver-breaking release
```
martin-g: Please add '#[deprecated(...)]'

Shnatsel: Done. I see the version in git is 0.18, while the latest version on crates.io is 0.17. That is a semver-incompatible version bump. Should I go ahead and remove the error entirely? Or are you planning on keeping compatibility with 0.17.x for a while?

avro/src/error.rs
```diff
@@ -377,6 +377,8 @@ pub enum Error {
     #[error("Failed to compress with flate: {0}")]
     DeflateCompress(#[source] std::io::Error),
 
+    // no longer possible after migration from libflate to miniz_oxide
+    #[deprecated(since = "0.18.0", note = "This error can no longer occur")]
```
```suggestion
    #[deprecated(since = "0.19.0", note = "This error can no longer occur")]
```
0.18.0 release is currently being voted at dev@avro.apache.org

avro/src/error.rs
martin-g: Despite still being 0.x we prefer to make it easier for the users. 
Keeping this error variant around for some months is not a problem!

avro/src/codec.rs
```diff
@@ -61,13 +61,8 @@ impl Codec {
...
             Codec::Null => (),
             Codec::Deflate => {
-                let mut encoder = Encoder::new(Vec::new());
-                encoder.write_all(stream).map_err(Error::DeflateCompress)?;
-                // Deflate errors seem to just be io::Error
-                *stream = encoder
-                    .finish()
-                    .into_result()
-                    .map_err(Error::DeflateCompressFinish)?;
+                let compressed = miniz_oxide::deflate::compress_to_vec(stream, 6);
```
martin-g, line 64: What is '6' here?
The compression level ?
Could we introduce settings like the other impls ?

avro/src/codec.rs
```diff
@@ -120,14 +115,20 @@ impl Codec {
...
-                    .read_to_end(&mut decoded)
-                    .map_err(Error::DeflateDecompress)?;
-                decoded
-            }
+            Codec::Deflate => miniz_oxide::inflate::decompress_to_vec(stream).map_err(|e| {
+                use miniz_oxide::inflate::TINFLStatus;
+                use std::io::ErrorKind;
+                let err_kind = match e.status {
+                    TINFLStatus::FailedCannotMakeProgress => ErrorKind::UnexpectedEof,
+                    TINFLStatus::BadParam => unreachable!(), // not possible for _to_vec()
```
martin-g: These 'unreachable's make me feel nervous. 
A library should not panic!
Better introduce a new error variant! Something like Error::Unexpected("An unexpected error occurred. Please report it https://github.com/apache/avro-ra/issues"). With some info about the cause would be perfect!

avro/src/codec.rs
Shnatsel: Yes, that is the compression level. Making it configurable is a good idea.

I kept the API the same in this PR to get the performance improvements to users without requiring them to change any code. Adding this configuration would be a good follow-up.

avro/src/codec.rs
Shnatsel: Done

martin-g (approved):

martin-g:
Thank you for the contribution, @Shnatsel !

avro/src/codec.rs
martin-g: Done with https://github.com/apache/avro-rs/pull/174

Issue #164, opened 2025-03-18
Support for different header types


masongup-mdsol:
While I was working with the `GenericSingleObjectReader` and `GenericSingleObjectWriter` objects to read and write the Kafka messages that the other tools in my company are using, I noticed that they are currently hard-coded to use a specific header fingerprint format: [reader](https://github.com/apache/avro-rs/blob/main/avro/src/reader.rs#L511-L527), [writer](https://github.com/apache/avro-rs/blob/main/avro/src/reader.rs#L511-L527). The messages that I am working with use an AWS Glue Schema Registry and [the corresponding header format](https://lenses.io/blog/2023/04/aws_glue_schema_registry_that_binds_apache_kafka/) on messages.

As a result of this, to handle these messages, I have to write a wrapper for the reader which reads out the Glue schema UUID manually, strips out those bytes, and adds on the header that the library expects, and for the writer that strips out the library-generated header and generates a new Glue header for the message. This all seems cumbersome to me. Are there any plans and/or interest to support either completely bare Avro data or proper interpretation of other types of message headers?

martin-g:
Would you like to send a PR ?

masongup-mdsol:
I'm open to building a solution. It seems complex enough that I'd like to discuss some possible designs to fix the issue though. I would expect we would want to provide a way to support other message header types without making API changes that would break current users.

martin-g:
I'm closing the issue since there is no activity for almost a month now.
Please refer to it if you decide to send a Pull Request with an improvement!

PR #190, opened 2025-04-25
Support different header types

masongup-mdsol:
Addresses #164 

I didn't get any response when I asked about preferred designs, so I went ahead and created one that seemed reasonable to me.

I created a new Header module with a HeaderBuilder trait, then reimplemented the existing Rabin fingerprint header as a struct impling this trait, and also another one with the Glue schema header type that I used. This design should be sufficiently generic to allow for the future implementation of more header types.

I also added a new method to the GenericSingleObjectReader and GenericSingleObjectWriter to allow the creation of each object type with a HeaderBuilder instance. I then modified the existing creator methods to keep the same signature but call these new methods with the RabinFingerprintHeader struct to maintain the current behavior, and I added a couple of new tests to exercise the functionality.

### DIFF
--- a/avro/src/headers.rs
+++ b/avro/src/headers.rs
@@ -1,0 +1,73 @@
+use uuid::Uuid;
+
+use crate::{rabin::Rabin, schema::SchemaFingerprint, Schema};
+
+pub trait HeaderBuilder {
+    fn build_header(&self) -> Vec<u8>;
+}
+
+pub struct RabinFingerprintHeader {
+    fingerprint: SchemaFingerprint,
+}
+
+impl RabinFingerprintHeader {
+    pub fn create_from_schema(schema: &Schema) -> Self {
+        let fingerprint = schema.fingerprint::<Rabin>();
+        RabinFingerprintHeader { fingerprint }
+    }
+}
+
+impl HeaderBuilder for RabinFingerprintHeader {
+    fn build_header(&self) -> Vec<u8> {
+        vec![
+            0xC3,
+            0x01,
+            self.fingerprint.bytes[0],
+            self.fingerprint.bytes[1],
+            self.fingerprint.bytes[2],
+            self.fingerprint.bytes[3],
+            self.fingerprint.bytes[4],
+            self.fingerprint.bytes[5],
+            self.fingerprint.bytes[6],
+            self.fingerprint.bytes[7],
+        ]
+    }
+}
+
+pub struct GlueSchemaUuidHeader {
+    schema_uuid: Uuid,
+}
+
+#[derive(thiserror::Error, Debug)]
+pub enum GlueHeaderError {
+    #[error("Message is too short to have a valid Glue header")]
+    MessageTooShort,
+    #[error("Failed to parse a valid UUID from the header: {0}")]
+    InvalidUuid(#[from] uuid::Error),
+}
+
+impl GlueSchemaUuidHeader {
+    pub fn create_from_uuid(schema_uuid: Uuid) -> Self {
+        GlueSchemaUuidHeader { schema_uuid }
+    }
+
+    pub fn parse_from_raw_avro(message_payload: &[u8]) -> Result<Self, GlueHeaderError> {
+        if message_payload.len() < 19 {
+            return Err(GlueHeaderError::MessageTooShort);
+        }
+        let schema_uuid = Uuid::from_slice(&message_payload[2..18])?;
+        Ok(GlueSchemaUuidHeader { schema_uuid })
+    }
+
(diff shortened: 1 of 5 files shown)


avro/src/headers.rs
```diff
@@ -0,0 +1,134 @@
...
+pub trait HeaderBuilder {
+    fn build_header(&self) -> Vec<u8>;
+}
+
+pub struct RabinFingerprintHeader {
+    fingerprint: SchemaFingerprint,
+}
+
+impl RabinFingerprintHeader {
+    pub fn create_from_schema(schema: &Schema) -> Self {
```
martin-g:
```suggestion
    pub fn from_schema(schema: &Schema) -> Self {
```

avro/src/headers.rs
```diff
@@ -0,0 +1,134 @@
+use uuid::Uuid;
+
+use crate::{rabin::Rabin, schema::SchemaFingerprint, AvroResult, Schema};
+
+pub trait HeaderBuilder {
```
martin-g, line 26: Please add rustdoc to all public items

avro/src/headers.rs
```diff
@@ -0,0 +1,134 @@
...
+        ]
+    }
+}
+
+pub struct GlueSchemaUuidHeader {
+    schema_uuid: Uuid,
+}
+
+impl GlueSchemaUuidHeader {
+    pub fn create_from_uuid(schema_uuid: Uuid) -> Self {
```
martin-g:
```suggestion
    pub fn from_uuid(schema_uuid: Uuid) -> Self {
```

avro/src/reader.rs
```diff
@@ -1034,6 +1031,24 @@ mod tests {
         Ok(())
     }
 
+    #[test]
+    fn test_avro_generic_reader_alternate_header() -> TestResult {
```
martin-g:
```suggestion
    fn avro_rs_164_generic_reader_alternate_header() -> TestResult {
```

avro/src/headers.rs
```diff
@@ -0,0 +1,163 @@
...
+    /// Create an instance of the struct based on parsing the UUID out of the header of a raw
+    /// message
+    ///
+    /// Code for reading messages will most likely want to use this. Once you receive the raw bytes
+    /// of a message, use this function to build the struct from it. That struct can then be used
+    /// with the below `schema_uuid` function to retrieve the UUID in order to retrieve the correct
+    /// schema for the message. You can then use the raw message, the schema, and the struct
+    /// instance to read the message.
+    pub fn parse_from_raw_avro(message_payload: &[u8]) -> AvroResult<Self> {
+        if message_payload.len() < 19 {
```
martin-g: Why `19` ?
The special prefix (`3, 0`) needs two bytes + 16 for the Uuid. What is the 19th ?

martin-g: OK. I guess the 19th is a byte from the actual payload.
Could you please check whether it works when the Avro schema is `Schema::Null`? In that case the value is zero sized.

masongup-mdsol: I had presumed the message would be invalid if there wasn't at least one byte of actual message data after the header. I'll try and write a test with the `Schema::Null`.

avro/src/headers.rs
```diff
@@ -0,0 +1,173 @@
...
+            return Err(crate::error::Error::HeaderMagic);
+        }
+        let schema_uuid =
+            Uuid::from_slice(&message_payload[2..18]).map_err(crate::Error::UuidFromSlice)?;
```
masongup-mdsol, line 92: If you really want it to have the `map_err`, that's okay, but note that [the UUID crate has no possible error for getting a UUID from a slice of bytes other than the length not being exactly 16](https://github.com/uuid-rs/uuid/blob/main/src/builder.rs#L328-L366).

avro/src/headers.rs
masongup-mdsol: Ah, I see you are correct, a message encoded with `Schema::Null` and `Value::Null` does indeed have zero bytes besides the header, so a message consisting of just a header can be perfectly valid and should be accepted. Updated.

avro/src/headers.rs
martin-g: I try to minimize the usage of `Result::unwrap()` as much as possible. A library should not panic! The user/caller should be able to react. Without this conversion I was not able to use `?` instead of `.unwrap()`.

I will add usage of 
```
#![deny(
    clippy::unwrap_used,
    clippy::expect_used,
    clippy::panic
)]
```
in another PR!

martin-g (approved):

martin-g:
Thank you for the contribution, @masongup-mdsol !

PR #193, opened 2025-05-01
feat: Expose the capability to do headerless avro encoding with serde

allenbenz:
My use case is I'm serializing large counts of individual objects to publish to google pubsub in an async setting where the "real" work being done elsewhere is so inexpensive to do that overhead dominates performance concerns.

Basically I want to use avro-rs to serialize with minimal allocation/cloning and I must not have headers or markers in the output.

Specifics you can skip:

> google's pubsub uses a stripped down version of the single record format (v1) - it is without any of the headers or markers, it only wants avro's raw binary encoding. Topics are configured with a the avro schema and the format so they are assumed when trying to publish to that topic.
> 
> The high volume makes it undesirable to construct a SpecificSingleObjectWriter per call as it clones the avro schema.
> The async setting prevents sharing the `SpecificSingleObjectWriter` safely as write_ref mutably borrows self.
> 
> My current workaround is using SyncUnsafeCell to share a mutable reference to a SpecificSingleObjectWriter that has already written a header once is not future proof in the slightest.

I'm not in love with `write_avro_datum_ref` as the name but it was close to the language of `SpecificSingleObjectWriter.write_ref` and `to_avro_datum`. I had `write_avro_datum_ref` take in a writer rather than make a Vec since that seems more sensible for serialization/allows it to be used in SpecificSingleObjectWriter.

### DIFF
--- a/avro/src/writer.rs
+++ b/avro/src/writer.rs
@@ -656,10 +656,10 @@ pub fn to_avro_datum<T: Into<Value>>(schema: &Schema, value: T) -> AvroResult<Ve
     Ok(buffer)
 }
 
-/// Write the referenced `Serialize` object to the provided `Write` object. Returns a result with
-/// the number of bytes written.
+/// Write the referenced [Serialize]able object to the provided [Write] object.
+/// Returns a result with the number of bytes written.
 ///
-/// **NOTE** This function has a quite small niche of usage and does NOT generate headers and sync
+/// **NOTE** This function has a quite small niche of usage and does **NOT** generate headers and sync
 /// markers; use [`append_ser`](struct.Writer.html#method.append_ser) to be fully Avro-compatible
 /// if you don't know what you are doing, instead.
 pub fn write_avro_datum_ref<T: Serialize, W: Write>(
@@ -750,12 +750,6 @@ mod tests {
     }
     "#;
 
-    #[derive(Serialize)]
-    struct TestStruct {
-        a: i64,
-        b: String,
-    }
-
     const UNION_SCHEMA: &str = r#"["null", "long"]"#;
 
     #[test]
@@ -776,7 +770,13 @@ mod tests {
     }
 
     #[test]
-    fn test_write_avro_datum_ref() -> TestResult {
+    fn avro_rs_193_write_avro_datum_ref() -> TestResult {
+        #[derive(Serialize)]
+        struct TestStruct {
+            a: i64,
+            b: String,
+        }
+
         let schema = Schema::parse_str(SCHEMA)?;
         let mut writer: Vec<u8> = Vec::new();
         let data = TestStruct {
--- a/avro/src/writer.rs
+++ b/avro/src/writer.rs
@@ -583,9 +583,7 @@ where
             self.header_written = true;
         }
 
-        let names: HashMap<Name, &Schema> = HashMap::new();
-        let mut serializer = SchemaAwareWriteSerializer::new(writer, &self.schema, &names, None);
-        bytes_written += data.serialize(&mut serializer)?;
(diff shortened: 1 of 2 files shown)


avro/src/writer.rs
```diff
@@ -658,6 +656,19 @@ pub fn to_avro_datum<T: Into<Value>>(schema: &Schema, value: T) -> AvroResult<Ve
     Ok(buffer)
 }
 
+/// Write the referenced `Serialize` object to the provided Write object. Returns a result with
```
```suggestion
/// Write the referenced `Serialize` object to the provided `Write` object. Returns a result with
```

avro/src/writer.rs
```diff
@@ -658,6 +656,19 @@ pub fn to_avro_datum<T: Into<Value>>(schema: &Schema, value: T) -> AvroResult<Ve
...
+/// the number of bytes written.
+///
+/// **NOTE** This function has a quite small niche of usage and does NOT generate headers and sync
+/// markers; use [`Writer`](struct.Writer.html) to be fully Avro-compatible if you don't know what
```
martin-g: > use [`Writer`](struct.Writer.html) to be fully Avro-compatible if you don't know what

I don't quite follow you here. What do you mean by `use Writer instead` ?
IMO it would be clearer to say something like `use write_ref(...) instead`.

martin-g (commented):
Please add some unit tests!

avro/src/writer.rs
allenbenz: So this was me copying the comment from `to_avro_datum`.

The issue with this and that function is they look like what someone would want, they're simple to call but they will do the wrong thing for most users.

I think the idea was to direct people to the Writer struct documentation which is what I imagine most people would want? I can change the link to (ignore the off formatting, can't write markdown in markdown)
['append_ser'] (struct.Writer.html#method.append_ser)

avro/src/writer.rs
allenbenz: Done

avro/src/writer.rs
```diff
@@ -658,6 +656,23 @@ pub fn to_avro_datum<T: Into<Value>>(schema: &Schema, value: T) -> AvroResult<Ve
     Ok(buffer)
 }
 
+/// Write the referenced [Serialize]able object to the provided [Write] object.
+/// Returns a result with the number of bytes written.
+///
+/// **NOTE** This function has a quite small niche of usage and does **NOT** generate headers and sync
+/// markers; use [`append_ser`](struct.Writer.html#method.append_ser) to be fully Avro-compatible
+/// if you don't know what you are doing, instead.
+pub fn write_avro_datum_ref<T: Serialize, W: Write>(
```
martin-g, line 665: How do you intent to use this method ?
In its current state it is not visible to the user application.
Should we add it to lib.rs ?
```diff
diff --git i/avro/src/lib.rs w/avro/src/lib.rs
index 930541a..abf775e 100644
--- i/avro/src/lib.rs
+++ w/avro/src/lib.rs
@@ -906,7 +906,7 @@ pub use ser::to_value;
 pub use util::{max_allocation_bytes, set_serde_human_readable};
 pub use uuid::Uuid;
 pub use writer::{
-    to_avro_datum, to_avro_datum_schemata, GenericSingleObjectWriter, SpecificSingleObjectWriter,
+    to_avro_datum, to_avro_datum_schemata, write_avro_datum_ref, GenericSingleObjectWriter, SpecificSingleObjectWriter,
     Writer,
 };
```

allenbenz: Yes, I'll add it to lib.rs. 🤦

martin-g:
Thank you for the contribution, @allenbenz !

allenbenz:
Thank you for the reviews @martin-g

PR #195, opened 2025-05-14
feat: Accept more inputs for `Schema::parse_(str_with)_list`

Kriskras99:
They now accept anything that can iterate over types that can be converted into strings. For example: `Vec<String>`, `&[&str]`, and `files.iter().map(std::fs::read_to_string).filter_map(Result::ok)`.

### DIFF
--- a/avro/src/schema.rs
+++ b/avro/src/schema.rs
@@ -1096,11 +1096,14 @@ impl Schema {
     /// during parsing.
     ///
     /// If two of the input schemas have the same fullname, an Error will be returned.
-    pub fn parse_list(input: &[&str]) -> AvroResult<Vec<Schema>> {
-        let mut input_schemas: HashMap<Name, Value> = HashMap::with_capacity(input.len());
-        let mut input_order: Vec<Name> = Vec::with_capacity(input.len());
-        for js in input {
-            let schema: Value = serde_json::from_str(js).map_err(Error::ParseSchemaJson)?;
+    pub fn parse_list(input: impl IntoIterator<Item = impl AsRef<str>>) -> AvroResult<Vec<Schema>> {
+        let input = input.into_iter();
+        let input_len = input.size_hint().0;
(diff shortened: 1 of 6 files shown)


avro/src/schema.rs
```diff
@@ -2782,7 +2788,7 @@ mod tests {
             ]
         }"#;
 
-        let schema_c = Schema::parse_list(&[schema_str_a, schema_str_b, schema_str_c])?
+        let schema_c = Schema::parse_list([schema_str_a, schema_str_b, schema_str_c])?
```
martin-g, line 2791: The PR title says `feat` and `Accept more inputs ...` but I just tried to revert this change and check whether the old input is still allowed but it fails with:
```
error: the borrowed expression implements the required traits
    --> avro/src/schema.rs:2791:43
     |
2791 |         let schema_c = Schema::parse_list(&[schema_str_a, schema_str_b, schema_str_c])?
     |                                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: change this to: `[schema_str_a, schema_str_b, schema_str_c]`
     |
```
So it is actually an API break.

This is not a big problem because Avro-rs is still 0.x but it would be nice if the old API is still supported.

martin-g: Actually this was `clippy` !
The `build` is just fine!

martin-g (approved):

martin-g:
Thank you for the contribution, @Kriskras99 !

Kriskras99:
@martin-g thanks for merging them so quickly! 

Would you accept a PR that fixes some newer clippy lints, or a PR that adds a rust-toolchain.toml file so that Clippy doesn't trigger on the new lints?

martin-g:
> Would you accept a PR that fixes some newer clippy lints, or a PR that adds a rust-toolchain.toml file so that Clippy doesn't trigger on the new lints?

Sure! Any help is welcome!
Please use separate PRs for the different improvements!

PR #196, opened 2025-05-15
chore: Fix a bunch of new Clippy lints

Kriskras99:
This fixes two lints:
 - `clippy::uninlined_format_args`: which will be moved from `pedantic` to `style` in the next Rust version.
 - `clippy::manual_contains`: which will be added in the next Rust version.
 
 This does not change the MSRV of 1.74.


martin-g (approved):

martin-g:
Thank you, @Kriskras99 !

PR #201, opened 2025-05-27
feat: Add `get_ref` and `get_mut` methods to `Writer`

Kriskras99:
This can be useful for checking the size of the output. For example, you might want to batch values every N megabytes.

I've also added two commits, one which simplifies the doc comments and one which makes the `WriterBuilder` type public.

The alternative to `peek_inner` is doing:
- `let aw = Writer::builder().schema(schema).marker(marker).writer(writer)`
- `aw.append_ser(value)`
- `writer = aw.into_inner()`
- `writer.len()`
- `aw = Writer::append_to(schema, writer, marker)`
And this recreates the writer every time, which does a bunch of allocations.

### DIFF
--- a/avro/src/writer.rs
+++ b/avro/src/writer.rs
@@ -351,6 +351,24 @@ impl<'a, W: Write> Writer<'a, W> {
         Ok(self.writer)
     }
 
+    /// Gets a reference to the underlying writer.
+    ///
+    /// **NOTE**: There is likely data still in the buffer. To have all the data
+    /// in the writer call [`flush`](Writer::flush) first.
+    pub fn get_ref(&self) -> &W {
+        &self.writer
+    }
+
+    /// Gets a mutable reference to the underlying writer.
+    ///
+    /// It is inadvisable to directly write to the underlying writer.
+    ///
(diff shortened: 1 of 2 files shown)


avro/src/writer.rs
```diff
@@ -343,14 +343,22 @@ impl<'a, W: Write> Writer<'a, W> {
...
         self.maybe_write_header()?;
         self.flush()?;
         Ok(self.writer)
     }
 
+    /// Peek at what the `Writer` is writing to.
+    ///
+    /// **NOTE**: There is likely data still in the buffer. To have all the data
+    /// in the writer call [`flush`](Writer::flush) first.
+    pub fn peek_inner(&self) -> &W {
```
erikbrinkman: I'm not sure if the name `peek_inner` has precedence anywhere, but if not, it may be better to adopt the convention of the standard library, which seems to just call this [`get_ref`](https://doc.rust-lang.org/std/io/struct.BufWriter.html#method.get_ref) for io wrappers like `BufWriter`

Kriskras99: You're right, renamed!

martin-g (approved):

PR #205, opened 2025-06-06
feat: Add support for rename of enum variants in AvroSchema macro

Vaalla42:
Improve the `AvroSchema` macro to support `#[avro(rename = "new_name")]` for enum variants.

Example:
```
enum A {
    #[avro(rename = "A3")]
    Item1,
}
```
will generate

```
Enum(EnumSchema { name: Name { name: "A", namespace: None }, aliases: None, doc: None, symbols: ["A3"], default: None, attributes: {} })
```

### DIFF
--- a/avro_derive/src/lib.rs
+++ b/avro_derive/src/lib.rs
@@ -224,8 +224,8 @@ fn get_data_enum_schema_def(
         let default = preserve_optional(default_value);
         let mut symbols = Vec::new();
         for variant in &e.variants {
-            let field_attrs = VariantOptions::from_attributes(&variant.attrs[..])
-                .map_err(darling_to_syn)?;
+            let field_attrs =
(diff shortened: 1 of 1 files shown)


avro_derive/src/lib.rs
```diff
@@ -215,11 +222,18 @@ fn get_data_enum_schema_def(
...
-        let symbols: Vec<String> = e
-            .variants
-            .iter()
-            .map(|variant| variant.ident.to_string())
-            .collect();
+        let mut symbols = Vec::new();
+        for variant in &e.variants {
+            let field_attrs = VariantOptions::from_attributes(&variant.attrs[..])
+                .map_err(darling_to_syn)
+                .unwrap();
```
martin-g: Can we use `?` instead ?

PR #207, opened 2025-06-09
feat: Add support for rename_all in AvroSchema macro

Vaalla42:
Supports the same settings as serde. 

Because we need to have the same behavior as serde, the case conversion code was copied from serde crate.

### DIFF
--- a/avro_derive/src/lib.rs
+++ b/avro_derive/src/lib.rs
@@ -710,7 +710,7 @@ mod tests {
     }
 
     #[test]
-    fn test_avro_3709_record_attributes() {
+    fn test_avro_rs_207_rename_all_attribute() {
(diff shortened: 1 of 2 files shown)


avro_derive/src/lib.rs
```diff
@@ -679,4 +708,47 @@ mod tests {
             ),
         };
     }
+
+    #[test]
+    fn test_avro_3709_record_attributes() {
```
martin-g: https://issues.apache.org/jira/browse/AVRO-3709 is about aliases

```suggestion
    fn test_avro_rs_207_rename_all_attribute() {
```

martin-g:
Thank you, @Vaalla42 !

PR #222, opened 2025-07-01
fix: Also (maybe) write header on `Writer::flush`.

### LINKED ISSUE #220: Should `Writer::flush` also write the header?
Currently `Writer::flush` only flushes the buffers but both `Writer::drop` and `Writer::into_inner` also write the header.

Does it make sense to also call `maybe_write_header` in `Writer::flush`?

Kriskras99:
Will only write the header if it wasn't written yet and it has not been disabled using `WriterBuilder`.

Closes #220

### DIFF
--- a/avro/src/writer.rs
+++ b/avro/src/writer.rs
@@ -53,6 +53,9 @@ pub struct Writer<'a, W: Write> {
     num_values: usize,
     #[builder(default = generate_sync_marker())]
     marker: [u8; 16],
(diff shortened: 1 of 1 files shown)


martin-g (commented):
Please add a new test case too

Kriskras99:
Test case added

martin-g (approved):

martin-g:
Thank you, @Kriskras99 !

PR #227, opened 2025-07-16
fix #226: Index-Out-Of-Bounds panic when using #[serde(skip_serializing_if=..)]

### LINKED ISSUE #226: Index-Out-Of-Bounds panic when using #[serde(skip_serializing_if = "Option::is_none")]
#### Code Example:
```rust
struct X {
  #[serde(skip_serializing_if = "Option::is_none")]
  y: Option<String>,
  z: Option<i8>
}

```

#### Issue Description:
while trying to serialize an instance such as `X { y: None, z: Some(0)}` using `Writer::append_ser(..)`, it panics with an index-out-of-bounds error pointing to: [#/src/ser_schema.rs:376:45](https://github.com/apache/avro-rs/blob/d8bc430148a0f160334a9b5ee661178f4e701892/avro/src/ser_schema.rs#L376)

#### RCA:
* Serde allows for field skipping using the `skip_serializing_if` directive.
* current implementation assumes that schema field count matchs the number of `serializer.serialize(..)` calls 100%.
* Use of unsafe random-access on a vec, panics when the assumption breaks.

elad-yosifon:
fix #226: Index-Out-Of-Bounds panic when using #[serde(skip_serializing_if=..)]

### DIFF
--- a/avro/src/ser_schema.rs
+++ b/avro/src/ser_schema.rs
@@ -18,6 +18,7 @@
 //! Logic for serde-compatible schema-aware serialization
 //! which writes directly to a `Write` stream
 
+use crate::schema::RecordField;
 use crate::{
     bigdecimal::big_decimal_as_bytes,
     encode::{encode_int, encode_long},
@@ -25,7 +26,7 @@ use crate::{
     schema::{Name, NamesRef, Namespace, RecordSchema, Schema},
 };
 use bigdecimal::BigDecimal;
-use serde::ser;
+use serde::{Serialize, ser};
 use std::{borrow::Cow, io::Write, str::FromStr};
 
 const RECORD_FIELD_INIT_BUFFER_SIZE: usize = 64;
@@ -328,12 +329,7 @@ impl<W: Write> ser::SerializeStruct for SchemaAwareWriteSerializeStruct<'_, '_, 
         }
 
         let next_field = &self.record_schema.fields[self.item_count];
-        let next_field_matches = match &next_field.aliases {
-            Some(aliases) => {
-                key == next_field.name.as_str() || aliases.iter().any(|a| key == a.as_str())
-            }
-            None => key == next_field.name.as_str(),
-        };
+        let next_field_matches = field_matches(next_field, key);
 
         if next_field_matches {
             self.serialize_next_field(&value).map_err(|e| {
@@ -342,18 +338,13 @@ impl<W: Write> ser::SerializeStruct for SchemaAwareWriteSerializeStruct<'_, '_, 
                     record_schema: Schema::Record(self.record_schema.clone()),
                     error: Box::new(e),
                 }
-            })?;
-            Ok(())
+                .into()
+            })
         } else {
             if self.item_count < self.record_schema.fields.len() {
                 for i in self.item_count..self.record_schema.fields.len() {
                     let field = &self.record_schema.fields[i];
-                    let field_matches = match &field.aliases {
-                        Some(aliases) => {
-                            key == field.name.as_str() || aliases.iter().any(|a| key == a.as_str())
-                        }
-                        None => key == field.name.as_str(),
-                    };
+                    let field_matches = field_matches(field, key);
 
                     if field_matches {
                         let mut buffer: Vec<u8> = Vec::with_capacity(RECORD_FIELD_INIT_BUFFER_SIZE);
@@ -382,13 +373,42 @@ impl<W: Write> ser::SerializeStruct for SchemaAwareWriteSerializeStruct<'_, '_, 
         }
     }
 
-    fn skip_field(&mut self, _key: &'static str) -> Result<(), Self::Error> {
-        self.item_count += 1;
+    fn skip_field(&mut self, key: &'static str) -> Result<(), Self::Error> {
+        match self.record_schema.fields.get(self.item_count) {
+            Some(skipped_field) => {
+                if field_matches(skipped_field, key) {
+                    self.item_count += 1;
+                    skipped_field
+                        .default
+                        .serialize(&mut SchemaAwareWriteSerializer::new(
+                            self.ser.writer,
+                            &skipped_field.schema,
+                            self.ser.names,
+                            self.ser.enclosing_namespace.clone(),
+                        ))?;
+                } else {
+                    return Err(Details::GetField(key.to_string()).into());
+                }
+            }
+            None => {
+                return Err(Details::GetField(key.to_string()).into());
+            }
+        }
         Ok(())
     }
 
     fn end(self) -> Result<Self::Ok, Self::Error> {
         self.end()
+    }
+}
+
+fn field_matches(record_field: &RecordField, expected_name: &str) -> bool {
+    let field_name = record_field.name.as_str();
+    match &record_field.aliases {
+        Some(aliases) => {
+            expected_name == field_name || aliases.iter().any(|a| expected_name == a.as_str())
+        }
+        None => expected_name == field_name,
     }
 }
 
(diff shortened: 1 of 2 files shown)


martin-g:
I applied the suggestions by Copilot (compare the serialized record with the deserialized one) and the assertion fails ... The `Some(1)` is lost ... I'll try to debug it soon!

martin-g:
This won't work with the current version of avro-rs.

With your PR the field is properly skipped, i.e. nothing is serialized for this field.
But Avro's Reader uses the T::schema() to lead the deserialization of the data and wrongly reads the `y`'s value into `x` field and `z`'s value into `y`.

For full roundtrip we will need a Serde-driven deserialization too.
ser.rs and de.rs are Schema-driven impls.
ser_schema.rs is Serde-driven serialization. We will need de_schema.rs (or some better names) for Serde-driven impl.

@jdarais Do you agree with me here ?

martin-g:
> @martin-g any ETA on this?

Dunno. 
I started working on it at https://github.com/apache/avro-rs/pull/237
Feel free to send PRs against my branch!

martin-g:
@elad-yosifon While working on the serde-driven deserialization I've realized that `skip_field()` needs to serialize something, otherwise the deserialization uses the wrong Avro Schema of the next field.
So I've added logic to use the RecordField's `default` value (an Option).
Now only the `multiple skipped fields` test fails due to the `skip` attribute. For some reason Serde does not call `skip_field()` for this fiels ...

martin-g:
I have some other things to do now but I will continue debugging it later.
Any help/hints why Serde does not call the method are welcome!

elad-yosifon:
@martin-g maybe it is a niche optimization, as all of the NONE optionals are at the tail, so one might think there is no need to indicate the skip.

martin-g:
@elad-yosifon I'd be OK to merge this PR in its current state if it is enough for your use case.
It supports `skip_serializing_if` but it does not support `skip_serializing`/`skip`! 
The latter will probably be supported once https://github.com/apache/avro-rs/pull/237 is finished (but it is a ton of work...).

elad-yosifon:
That’s good for now 👍

jdarais:
Sorry about the delayed response, I've been traveling.  On this question:

> For full roundtrip we will need a Serde-driven deserialization too.
ser.rs and de.rs are Schema-driven impls.
ser_schema.rs is Serde-driven serialization. We will need de_schema.rs (or some better names) for Serde-driven impl.

I was looking into creating a serde-driven deserializer, but in the end it seemed easier to first deserialize into an avro `Value`, which provides some flexibility that helps with schema resolution (if the reader schema is different from the writer schema,) and then use the existing `de.rs` implementation to convert from `Value` to the final type.

I feel like this approach should still work for a round-trip of serialization and deserialization.  If it doesn't then I think that would mean that either the serde-driven serialization or the schema-driven deserialization has a bug, or simply that the default value used for serialization, (provided by the schema,) is different from the default value used for deserialization, (which as far as I can tell uses `Default::default` and bypasses deserialization logic altogether.)

In general, the `skip_serializing_if` attribute seems to be at odds with serialization to Avro, since Avro structs give no accommodation for simply skipping serialization of a field: all fields of a struct must be present.  (Using the default value for the field provided by the schema is a reasonable workaround, since that simulates omiting the field, though per the [spec](https://avro.apache.org/docs/1.12.0/specification/#schema-record), the default value is "only used when reading instances that lack the field for schema evolution purposes. The presence of a default value does not make the field optional at encoding time."  Also, with the changes from this PR, if the default value is not present, then the struct will be serialized incorrectly.)

Using the `skip` attribute would be easier to use with Avro since you can just ensure that the skipped field isn't included in the Avro schema used to write.

Now that I think about it, I'm guessing that `skip_serializing_if` is mostly useful for removing attributes that are just `null` or the default value so they don't take up unnecessary space in the serialized payload.  It might work to just have the `skip_field` implementation do the exact same thing as `serialize_field`, effectively always including the field, since you're not allowed to omit the field anyway.

On a different topic: I see from the PR that there was some logic in there before that ensured that struct fields were written in the correct order, (they must be written in the order in which they appear in the schema,) even if the order of the fields in the rust struct is different from what's in the schema.  It'd be nice to get that back in.

martin-g:
Hi @jdarais !

> I was looking into creating a serde-driven deserializer, but in the end it seemed easier to first deserialize into an avro `Value`, which provides some flexibility that helps with schema resolution (if the reader schema is different from the writer schema,) and then use the existing `de.rs` implementation to convert from `Value` to the final type.

The problem here is that the users expect that the schema-driven deserialization in de.rs should take into account the Serde attributes, like the `skip` ones. But maybe as you said we just have to ignore them. Otherwise we will face issues later with attributes like `flatten`, `remote`, etc.
But in that case I have no good answer to "I want to use the same serde mechanism for JSON, CSV, Avro, ...) (https://github.com/apache/avro-rs/pull/227#issuecomment-3097968926).



> Using the `skip` attribute would be easier to use with Avro since you can just ensure that the skipped field isn't included in the Avro schema used to write.

Supporting `skip` is OK. 
But what to do with `skip_serializing` and `skip_deserialing` ?! We will need to have a field for such fields in the Avro Schema but Serde does not call `skip_field()` when `skip_serializing` is used ...


> It might work to just have the `skip_field` implementation do the exact same thing as `serialize_field`, effectively always including the field, since you're not allowed to omit the field anyway.

We kinda do this at the moment.
* `fn serialize_field<T>(&mut self, key: &'static str, value: &T)`
* `fn skip_field(&mut self, key: &'static str)`
`skip_field` does not provide the `value`, so we use the `default` from the Avro Schema. This might need some more work! At the moment it is treated as an Option but probably it should be unpacked.



> On a different topic: I see from the PR that there was some logic in there before that ensured that struct fields were written in the correct order, (they must be written in the order in which they appear in the schema,) even if the order of the fields in the rust struct is different from what's in the schema. It'd be nice to get that back in. (I realize there should have been a test for it.)

You are totally right here!
It should work fine when the schema is derived but it will break bad at deserialization time if the schema is created manually!
For now we can add a note to the documentation! 

I removed all logic that didn't break any test :-/
It could be returned with the respective unit tests!

martin-g:
> It should work fine when the schema is derived but it will break bad at deserialization time if the schema is created manually! For now we can add a note to the documentation!

https://github.com/apache/avro-rs/pull/242

jdarais:
Ah, those are some good points...

> Supporting skip is OK.
> But what to do with skip_serializing and skip_deserialing ?! We will need to have a field for such fields in the Avro Schema but Serde does not call skip_field() when skip_serializing is used ...
> 

I think `skip_serializing` and `skip_deserializing` are also doable.  Since the Serializer is the one with the schema, the schema is decoupled from the struct implementing `Serialize`, so you just have to make sure that the schema you use to serialize doesn't include the field with the `skip_serializing` attribute.  (Of course, reading the value back would require a different schema that includes the field :slightly_frowning_face: but it is doable with different read and write schemas.  If you use `skip_serializing` but not `skip_deserializing`, then maybe that's just what you're signing yourself up for.)

`skip_deserializing` should work more easily, since if we first read the value into an avro `Value` using the schema, then when we convert from `Value` to the struct using the `de.rs` implementation, it should be fine to ignore the value since doing so won't cause issues with parsing the rest of the record in the same way that ignoring a field does when serializing.

> skip_field does not provide the value, so we use the default from the Avro Schema. This might need some more work! At the moment it is treated as an Option but probably it should be unpacked.

Ah, that's a very good point.  That's a bummer that the value isn't provided in `skip_field`.  Yeah, if the default value is `None` (meaning no default value was specified) then we're kind of stuck, since there's no way to know what should go in that field.  Might be good to just return an `Err` in that case.

PR #228, opened 2025-07-17
fix: Improve the error message for unresolved union variant

### LINKED ISSUE #225: Improve error message for `Could not find matching type in union`
Hey @martin-g,

I've been working on Iceberg-Rust, which we depend on in PyIceberg. I get errors from PyIceberg `Could not find matching type in union`, but that makes it pretty hard to debug. Would it be possible to add a field-name in there? Happy to dig into this myself as well, but I'm not a rustacean (yet) :)

martin-g:
Fixes #225

Add the UnionSchema and the actual Value to the error message for easier debugging

### DIFF
--- a/avro/src/error.rs
+++ b/avro/src/error.rs
@@ -16,7 +16,7 @@
 // under the License.
 
 use crate::{
-    schema::{Name, Schema, SchemaKind},
+    schema::{Name, Schema, SchemaKind, UnionSchema},
     types::{Value, ValueKind},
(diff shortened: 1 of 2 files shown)


Fokko (approved):
Hey @martin-g This is great, and will save me a lot of time on debugging 🚀

martin-g:
Good point!
I will do it tomorrow unless someone beats me to it!

PR #230, opened 2025-07-19
feat: Wrap `Error` into a `Box` so it's size shrinks from 104 bytes to 8 bytes

Kriskras99:
Supersedes #229.

The size of a `Result` is as large as it's largest variant, thus `AvroResult` is currently always at least 104 bytes. Therefore, even if the function only returns a `usize` on success the function still needs to (stack) allocate 104 bytes.

I've renamed the current `Error` enum to `Details`, and actually made it a bit bigger by unboxing `Schema` arguments. The new `Error` type is just a wrapper around a `Box<Details>`.

To keep the code simple and make as few changes as possible, I've added a construction function for every variant of `Details`. I've also changed any `ok_or(Error...`) to an `ok_or_else(Error...` as constructing the type is now "expensive" because it does an allocation.

This is a breaking change if the user is matching on the error type. The fix is easy as the user can use `details` or `into_details` to get the inner `Details` to match on. When the user is constructing a struct-like variant of the type this is also a breaking change. This can be fixed by changing the curly braces to round brackets and removing the item names.


Kriskras99:
I also think `Details::GetLogicalTypeVariant(serde_json::Value)` is unused, if we're doing a breaking change on the error type might as well remove it.

In the same vein we could remove `Details::DeflateCompressFinish(std::io::Error)` which was deprecated in the last release.

avro/src/lib.rs
```diff
@@ -896,7 +896,7 @@ pub use codec::{Codec, DeflateSettings};
 pub use de::from_value;
 pub use decimal::Decimal;
 pub use duration::{Days, Duration, Millis, Months};
-pub use error::Error;
+pub use error::{Details, Error};
```
martin-g: It is not very clear that `apache_avro::Details` is about Error's details.
IMO we should either export it as `apache_avro::ErrorDetails` or not export it at all and the end users should use it as `apache_avro::error::Details`. I think I like the latter better.

avro/src/error.rs
```diff
@@ -21,8 +21,1154 @@ use crate::{
...
+    pub fn into_details(self) -> Details {
+        *self.details
+    }
+}
+
+/// Functions for constructing a specific error type.
+#[allow(non_snake_case, reason = "Want to mimic the `Details` variants")]
+impl Error {
+    /// Construct a new [`Error`] with a [`Details::SnappyCrc32`].
+    pub fn SnappyCrc32(expected: u32, actual: u32) -> Self {
```
martin-g: I wonder whether we really need to have these public factory methods.
Usually the library creates the "instance". The user applications check/match the error.
But even if the user applications need to create an instance they could replace `Error::Xyz{...}` with `Details::Xyz{...}.into()`, i.e. `s/Error/Details/` and append `.into()`. We could use the same in the library too.

martin-g (approved):
I like this better than #229 !
I understand that it is a bigger API break than #229 but I think it will be easier to maintain in the future since there is just one `Box<>` to have in mind.

martin-g:
> I also think `Details::GetLogicalTypeVariant(serde_json::Value)` is unused, if we're doing a breaking change on the error type might as well remove it.

Let's deprecate it for now. We could remove it few releases later.

We have few other items deprecated in older versions (e.g. 0.11 and 0.15) which we could remove for 0.20! But in a separate PR.

avro/src/error.rs
Kriskras99: I agree they don't need to be public.

We could remove them completely but then a lot more lines will have to be changed, if you want that let me know and I'll do it.

Kriskras99:
Shall I mark it as deprecated in this PR or do you also want that in a separate PR?

avro/src/lib.rs
Kriskras99: I've made `apache_avro::error` public and removed the `pub use Details`.
This does make the `CompatibalityError` type public, but I think it's supposed to be public anyway?

avro/src/error.rs
Kriskras99: I've marked the construction functions as `pub(crate)` for now, let me know if you want to completely remove them.

avro/src/lib.rs
martin-g: > but I think it's supposed to be public anyway

Yes, I think it should be!

avro/src/error.rs
martin-g: > We could remove them completely but then a lot more lines will have to be changed, if you want that let me know and I'll do it.

We change those lines anyway in this PR.
Please make the change for just a few of the variants to see how it looks like and then we can decide!

avro/src/bigdecimal.rs
```diff
@@ -50,27 +51,27 @@ pub(crate) fn deserialize_big_decimal(bytes: &Vec<u8>) -> AvroResult<BigDecimal>
     let mut bytes: &[u8] = bytes.as_slice();
     let mut big_decimal_buffer = match decode_len(&mut bytes) {
         Ok(size) => vec![0u8; size],
-        Err(err) => return Err(Error::BigDecimalLen(Box::new(err))),
+        Err(err) => return Err(Details::BigDecimalLen(Box::new(err)).into()),
```
martin-g, line 54: We can drop the usage of Box here, right ?

avro/src/codec.rs
```diff
@@ -91,7 +91,7 @@ impl Codec {
                 let mut encoded: Vec<u8> = vec![0; snap::raw::max_compress_len(stream.len())];
                 let compressed_size = snap::raw::Encoder::new()
                     .compress(&stream[..], &mut encoded[..])
-                    .map_err(Error::SnappyCompress)?;
+                    .map_err(Details::SnappyCompress)?;
```
martin-g, line 94: How does this work without the `.into()` call ?

avro/src/decimal.rs
```diff
@@ -93,7 +93,7 @@ impl Decimal {
         let mut decimal_bytes = vec![sign_byte; len];
         let raw_bytes = self.value.to_signed_bytes_be();
         let num_raw_bytes = raw_bytes.len();
-        let start_byte_index = len.checked_sub(num_raw_bytes).ok_or(Error::SignExtend {
+        let start_byte_index = len.checked_sub(num_raw_bytes).ok_or(Details::SignExtend {
```
martin-g, line 96: ok_or_else ?

avro/src/bigdecimal.rs
Kriskras99: No, that would make the Error struct infinitely large (Error -> Details -> Error -> Details ...)

avro/src/codec.rs
Kriskras99: The try operator (question mark: ?) automatically calls into

avro/src/decimal.rs
Kriskras99: Because it's not a function call but just an enum construction it's not necessary. Clippy even wants you to use `ok_or` if you're calling .len() on something

martin-g (approved):

PR #237, opened 2025-07-21
feat: Serde driven schema aware deserialization

martin-g:
Related to #226


Kriskras99:
I think this can be closed right?

Issue #136, opened 2025-02-25
Async versions for Reader and Writer


splix:
I'm wondering if there are any plans (or opposition to those) for async implementation for Reader and Writer? I.e., a reader from `tokio::io::AsyncRead`, etc

splix:
I've tried to add it, but quickly realized that it requires [mostly] duplicate versions for too many of internal functions.

Like for 
```
pub fn zag_i64<R: Read>(reader: &mut R) -> AvroResult<i64> {
    let z = decode_variable(reader)?;
    Ok(if z & 0x1 == 0 {
        (z >> 1) as i64
    } else {
        !(z >> 1) as i64
    })
}
```

it needs an accompanying:
```
#[cfg(feature = "tokio")]
pub async fn zag_i64_async<R: AsyncRead>(reader: &mut R) -> AvroResult<i64> {
    let z = decode_variable_async(reader).await?;
    Ok(if z & 0x1 == 0 {
        (z >> 1) as i64
    } else {
        !(z >> 1) as i64
    })
}
```

and so on.

It's definitely possible to implement, but the main question if that aligns well with `avro-rs` plans?

martin-g:
Duplicating the code is no go. 

IMO a more feasible way is to make everything async and use a same-thread thread pool for sync usage. But I haven't seen any other library doing it so probably I miss something.

splix:
For this particular example it's possible to move out the duplicate part into an _inline_ function, and keep only `decode_variable_async(reader).await?;` and `decode_variable_async(reader)?;` depending on the variant. Another option is wrapping those with some macros which depending on the compilation target adds `.await` or not. Anyway, I agree that both options are bit ugly.

Or actually I think that many functions like this one could be transformed to something more universal which could work in both ways. I'm worried mostly about the fact that too many things must be touched to support async.

splix:
Regarding the sync usage, there is https://docs.rs/tokio-util/0.7.4/tokio_util/io/struct.SyncIoBridge.html that allows it

martin-g:
Closing the issue since there is no activity for almost 2 months now.
We could re-open it if there are ideas how to do it!

Kriskras99:
Reqwest is async by default, but has a `blocking` feature flag. It wraps the async types and spawns a Tokio thread to execute the async functions (https://github.com/seanmonstar/reqwest/blob/master/src/blocking/client.rs#L1234).

[...]

PR #238, opened 2025-07-21
RFC for async

Kriskras99:
A start at #136

I've decided to do the RFC as a PR so it's easier to give inline commentary.

I strongly recommend reading the blog post mentioned in the references and watch the video about `rc-zip` (becomes relevant after ~10 minutes).

Feel free to edit/expand the RFC.


martin-g:
@Xuanwo @xxchan @Fokko Do you have experience with these matters in your libraries ?

rfc.md
```diff
@@ -0,0 +1,88 @@
...
+Care needs to be taken that the state machines are kept small and understandable.
+
+The second-best option is probably using `block_on` in a separate thread. But that seems unnecessarily heavy.
+
+# References
+
+- [Blog post by the maintainer of `RSpotify` who tried multiple of the above options](https://nullderef.com/blog/rust-async-sync/)
+- [A discussion about Sans I/O](https://sdr-podcast.com/episodes/sans-io/)
+- [A explanation of Sans I/O by the author of `rc-zip`](https://fasterthanli.me/articles/the-case-for-sans-io)
+  - The blog post is currently not freely available, but the video (which has the exact same content) is freely available
```
martin-g:
```suggestion
  - The blog post is currently not freely available, but the [video](https://www.youtube.com/watch?v=RYHYiXMJdZI) (which has the exact same content) is freely available
```

martin-g (commented):
I just watched the video about Sans IO and honestly I still have no idea how the implementation will look like ...
But looking at https://github.com/bearcove/rc-zip I see that we will need to add more crates for each runtime - tokio, async-std/smol, monoio, sync, etc. Tokio is pretty much the standard for async but the idea of having to add a crate for every runtime looks scary to me.

For now `maybe_async` looks most promising to me!

(This is not a final decision! We are discussing!)

Kriskras99:
That looks a lot better than `maybe_async`. On the state machine front, I also found [this article](https://jeffmcbride.net/blog/2025/05/16/rust-async-functions-as-state-machines/) about using async functions to generate state machines which are usable from sync code. It's code is a bit messy because they seem to want to use it for embedded and thus don't want to use the heap.

rfc.md
```diff
@@ -0,0 +1,88 @@
...
+## Implement in async, use `maybe_async` to generate sync implementation
+
+[`maybe_async`](https://crates.io/crates/maybe-async) is a proc macro that removes the `.await` from the async code and uses it to generate sync code.
+
+Pros:
+ - Only need to maintain/test/upgrade one implementation
+ - Optimal performance for both async and sync code
+
+Cons:
+ - Crate breaks if both the `sync` and `async` features are enabled
```
Kriskras99:
```suggestion
[`maybe_async`](https://crates.io/crates/maybe-async) is a proc macro that removes the `.await` from the async code and uses it to generate sync code. [`synca`](https://docs.rs/synca/latest/synca/) is another option where both sync and async code can coexist.

Pros:
 - Only need to maintain/test/upgrade one implementation
 - Optimal performance for both async and sync code

Cons:
 - Crate breaks if both the `sync` and `async` features are enabled (only for `maybe_async`)
 - `synca` hasn't seen an update in more than a year, but seems to be feature complete
```

rfc.md
```diff
@@ -0,0 +1,89 @@
...
+Implement the parser as a state machine that can be driven by both async and sync code. This is how [`rc-zip`](https://lib.rs/crates/rc-zip)
+is implemented.
+
+Pros:
+ - Only need to maintain/test/upgrade one implementation
+ - Optimal performance for both async and sync code
+
+Cons:
+ - Have to manually implement the state machines
+   - In the distant future [it's possible to use coroutines/generators](https://internals.rust-lang.org/t/using-coroutines-for-a-sans-io-parser/22968), but they're currently *very* unstable.
```
Kriskras99, line 53:
```suggestion
   - In the distant future [it's possible to use coroutines/generators](https://internals.rust-lang.org/t/using-coroutines-for-a-sans-io-parser/22968), but they're currently *very* unstable.
   - You can use async functions to generate the state machines for you, [according to this blogpost](https://jeffmcbride.net/blog/2025/05/16/rust-async-functions-as-state-machines/)
```

Kriskras99:
I've pushed a (non-functional) prototype for parsing with state machines.

The most interesting thing about this prototype is the use of a "command" tape and a "output" tape.
I created the "command" tape as a replacement for a Schema in the state machine, as it's really annoying to keep track of where you are in the Schema (lots of `if let Schema::`).
The "output" tape replaces the Value while the state machine is still parsing the input. This has the same reason as not using the Schema. As without the schema, the field names are not known (or they could be known but would take a lot of space in the "command" tape).

The implementation is far from perfect and there is a lot that could be improved. For one, the `ObjectStateMachine` has a lot of duplication going on in the `parse` function. I'm pretty sure that duplication can be removed quite easily, making it a lot more readable. This implementation also cuts a lot of corners, like EOF handling or compression.

I do think it gives a good idea of what it would be like to have the core be state machines, wrapped in async and sync wrappers.

I currently only have a sync wrapper, but I'll start on an async wrapper just using the `futures` crate which should be compatible with almost all async runtimes (either through adapters or directly).

Kriskras99:
I did not check clippy before pushing, as it doesn't function anyway (too many `todo!()`)

Kriskras99:
I also did an attempt at making a SyncA POC, but don't yet known what to do about the iterators

avro/src/state_machines/reading/mod.rs
```diff
@@ -0,0 +1,338 @@
...
+
+    /// Check if the section of the tape we're reading is finished.
+    pub fn is_finished(&self) -> bool {
+        self.read_range.is_empty()
+    }
+
+    /// Extract a part from the tape to give to a sub state machine.
+    ///
+    /// The tape will run from start to end (inclusive).
+    pub fn extract(&self, start: usize, end: usize) -> Self {
```
martin-g: assert that start < end ?!

Kriskras99: Yeah, that would be good. Although I'm thinking of changing it to `offset` and `length` as it's more likely that the length will fit in 4 bits, which means it can be inlined into the command byte.

Kriskras99:
@martin-g if you like the statemachine approach, I can start work on making it a proper PR (including stuff like tests and handling edgecases). If you'd like to see a POC with SyncA first, that's possible too.

Kriskras99:
That sounds like a good plan, lets do it in different branches than this rfc one.

Kriskras99:
I've completed the reader part!

There are a few issues/notes:
1. `snap` does not have a streaming decoder, it's easy to implement but getting the PR upstreamed will take some work [as the maintainer isn't active on this project anymore](https://github.com/BurntSushi/rust-snappy/issues/48)
2. I've disabled the `reader::test_from_avro_datum_with_union_to_struct` test. In my opinion the provided schema is not compatible with the bytes. The user should instead provide the right schema and use a reader schema for the schema evolution. If we keep this, we should inform users as [at least one does depend on it](https://github.com/apache/avro/pull/1379)
3. I don't support having `Schema::Null` as the root schema in combination with the `from_avro_datum*` functions. They see an empty byteslice as an EoF. It's possible to add support for this, but I don't think anybody should depend on it.
4. I've moved all the tests that were in `decode.rs` to `state_machines/reading/mod.rs`, so they're not gone!
5. The current code organization is temporary and can be improved. I can imagine making `decode` a module and putting the state machines in there, making `reader` also a module with `sync` and `async` submodules. The same goes for the errors.
6. I did not yet implement direct from tape deserialisation from Serde types. I plan on implementing this, as it's more efficient than going through `Value` first.
7. There are a bunch of `TODO` scattered all over the place, also in places I did not modify. They are potential improvement points I ran into while implementing all of this.
8. I've temporarily added the `rustfmt.toml` back, so I can sort imports more easily.

avro/Cargo.toml
```diff
@@ -73,6 +76,10 @@ thiserror = { default-features = false, version = "2.0.12" }
 uuid = { default-features = false, version = "1.17.0", features = ["serde", "std"] }
 xz2 = { default-features = false, version = "0.1.7", optional = true }
 zstd = { default-features = false, version = "0.13.3", optional = true }
+oval = { version = "2.0.0", features = ["bytes"] }
+futures = "0.3.31"
+async-stream = "0.3.6"
+colored-hexdump = "0.2.2"
```
martin-g: Is this used somewhere ? I guess you use it for debugging

avro/tests/io.rs
```diff
@@ -232,7 +234,11 @@ fn test_round_trip() -> TestResult {
     for (raw_schema, value) in schemas_to_validate().iter() {
         let schema = Schema::parse_str(raw_schema)?;
         let encoded = to_avro_datum(&schema, value.clone()).unwrap();
-        let decoded = from_avro_datum(&schema, &mut Cursor::new(encoded), None).unwrap();
+        let decoded = from_avro_datum(&schema, &mut Cursor::new(encoded), None);
+        if decoded.is_err() {
+            println!("{schema:#?}");
```
martin-g:
```suggestion
            // TODO: remove before merging
            println!("{schema:#?}");
```

avro/tests/io.rs
```diff
@@ -25,7 +25,8 @@ fn schemas_to_validate() -> &'static Vec<(&'static str, Value)> {
     static SCHEMAS_TO_VALIDATE_ONCE: OnceLock<Vec<(&'static str, Value)>> = OnceLock::new();
     SCHEMAS_TO_VALIDATE_ONCE.get_or_init(|| {
         vec![
-            (r#""null""#, Value::Null),
+            // TODO: Do we want to support this? It makes the implementation significantly more complex
+            // (r#""null""#, Value::Null),
```
martin-g: What is the problem with it?
I am not sure whether it is really needed by someone. It was here when I adopted the codebase.

Kriskras99: `from_avro_datum*` assume they can always read at least one byte, but a datum with only null doesn't have any bytes. If it can't read any bytes it will return an UnexpectedEof.
Thinking about it again, we can add a check to see if the schema is null and then just return immediately.

avro/Cargo.toml
Kriskras99: Yep, was really useful to see the avro files

Kriskras99:
I'm closing this as "won't do" for now.

PR #239, opened 2025-07-23
feat: avro_derive: Allow providing a name other than the ident

bryanburgers:
I'm trying to write Rust structs that match an external schema that is created outside of Rust.

The external schema uses names that aren't idiomatic Rust struct identifiers.

Allowing the renaming of structs and enums means I don't need to compromise on Rust identifiers just to match the schema exactly.

### DIFF
--- a/avro_derive/tests/derive.rs
+++ b/avro_derive/tests/derive.rs
@@ -657,7 +657,7 @@ mod test_derive {
     }
 
     #[test]
-    fn test_enum_named() {
+    fn avro_rs_239_test_enum_named() {
         #[derive(Debug, Serialize, Deserialize, AvroSchema, Clone, PartialEq, Eq)]
         #[avro(name = "Other", rename_all = "snake_case")]
         #[serde(rename_all = "snake_case")]


avro_derive/tests/derive.rs
```diff
@@ -209,6 +209,41 @@ mod test_derive {
...
+                    "type":"string"
+                }
+            ]
+        }
+        "#;
+        let schema = Schema::parse_str(schema).unwrap();
+        assert_eq!(schema, TestNamedRecord::get_schema());
+        if let Schema::Record(RecordSchema { name, .. }) = TestNamedRecord::get_schema() {
+            assert_eq!("Other".to_owned(), name.name);
+            assert_eq!("com.testing.namespace".to_owned(), name.namespace.unwrap())
```
Kriskras99:
```suggestion
            assert_eq!("Other", name.name.as_str());
            assert_eq!("com.testing.namespace", name.namespace.unwrap().as_str())
```
This is slightly nicer in my opinion

bryanburgers: Oh absolutely! Updated.

avro_derive/tests/derive.rs
```diff
@@ -209,6 +209,41 @@ mod test_derive {
         }
     }
 
+    #[test]
+    fn test_named_record() {
```
```suggestion
    fn avro_rs_239_test_named_record() {
```

avro_derive/tests/derive.rs
```diff
@@ -621,6 +656,55 @@ mod test_derive {
         serde_assert(enum_included);
     }
 
+    #[test]
+    fn test_enum_named() {
```
```suggestion
    fn avro_rs_239_test_enum_named() {
```

bryanburgers:
Thanks for looking at this so quickly!

Let me know if you’d like me to rebase into a single commit when you’re done reviewing.

PR #242, opened 2025-07-25
doc: Document that the fields' order is important for "the serde way"

martin-g:
https://github.com/apache/avro-rs/pull/227#issuecomment-3116660601

### DIFF
--- a/avro/README.md
+++ b/avro/README.md
@@ -268,12 +268,18 @@ writer.append_ser(test).unwrap();
 let encoded = writer.into_inner();
 ```
 
+#### Importance of the fields' order
+
+*Important*: The order of the fields in the struct must match the order of the fields in the Avro schema!
+
+#### Simple types
+
 The vast majority of the times, schemas tend to define a record as a top-level container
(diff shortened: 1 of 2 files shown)


Kriskras99 (approved):

Kriskras99:
If we're keeping the `README.md` and `src/lib.rs` manually in sync, it might make sense to do something like `#![doc = include_str!(../../README.md)]`

Kriskras99:
Check, I think having the badges, License and Contributing in `src/lib.rs` is totally fine. It also means that a contributor not aware of `cargo-rdme` doesn't forget to update the readme after fixing/adding something in the documentation

martin-g:
> t also means that a contributor not aware of `cargo-rdme` doesn't forget to update the readme after fixing/adding something in the documentation

The CI will let him/her know - https://github.com/apache/avro-rs/blob/bd1769143018f9be109ba7ed779f34a969f9567e/.github/workflows/test-lang-rust-ci.yml#L90-L96

Kriskras99:
Ah perfect. Than it does not really matter

PR #246, opened 2025-07-28
WIP: Use SyncA to generate [a]sync APIs

martin-g:
This PR should be reviewed with whitespaces ignored - https://github.com/apache/avro-rs/pull/246/files?w=1

# What 

The purpose of this PR is to experiment with https://github.com/sgr-team/rs_synca crate to generate both async and sync/blocking APIs of the apache_avro SDK. Synca uses Rust procedural macros to generate the blocking version of the code from the manually written asynchronous version.

# How it works

Synca works by "annotating" a module with async functions with `#[synca::synca]` attribute and some rules what to do when generating the sync/blocking version of the code, e.g. to remove all occurrences of `async` and `.await` and to replace some `use`s (imports) with others. In addition to removing `async/await` it also removes any usages of `Box::pin()` for the `await`ed calls which are needed to solve the issue with recursion in `async` methods.

Most of the data structures, like `schema::Schema`, `types::Value`, `error::Error` are out of Synca's scope and are reusable from both sync and async callers. Some of their methods need to be `async` and they are extracted to `XyzExt`, e.g. `SchemaExt` structs which are processed by Synca. For example `Schema::parse_str(&str) -> AvroResult<Self>` needs to be async now and that's why it is moved to `SchemaExt::parse_str(&str) -> AvroResult<Schema>`. Same is valid for `Value`'s `validate()` and `resolve()` methods.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -145,9 +145,9 @@ dependencies = [
 
 [[package]]
 name = "bigdecimal"
-version = "0.4.8"
+version = "0.4.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1a22f228ab7a1b23027ccc6c350b72868017af7ea8356fbdf19f8d991c690013"
+checksum = "560f42649de9fa436b73517378a147ec21f6c997a546581df4b4b31677828934"
 dependencies = [
  "autocfg",
  "libm",
@@ -174,9 +174,9 @@ dependencies = [
 
 [[package]]
 name = "bon"
-version = "3.7.2"
+version = "3.8.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c2529c31017402be841eb45892278a6c21a000c0a17643af326c73a73f83f0fb"
+checksum = "ebeb9aaf9329dff6ceb65c689ca3db33dbf15f324909c60e4e5eef5701ce31b1"
 dependencies = [
  "bon-macros",
  "rustversion",
@@ -184,9 +184,9 @@ dependencies = [
 
 [[package]]
 name = "bon-macros"
-version = "3.7.2"
+version = "3.8.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d82020dadcb845a345591863adb65d74fa8dc5c18a0b6d408470e13b7adc7005"
+checksum = "77e9d642a7e3a318e37c2c9427b5a6a48aa1ad55dcd986f3034ab2239045a645"
 dependencies = [
  "darling",
  "ident_case",
@@ -211,9 +211,9 @@ checksum = "d71b6127be86fdcfddb610f7182ac57211d4b18a3e9c82eb2d17662f2227ad6a"
 
 [[package]]
 name = "bzip2"
-version = "0.6.0"
+version = "0.6.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "bea8dcd42434048e4f7a304411d9273a411f647446c1234a65ce0554923f4cff"
+checksum = "f3a53fac24f34a81bc9954b5d6cfce0c21e18ec6959f44f56e8e90e4bb7c346c"
 dependencies = [
  "libbz2-rs-sys",
 ]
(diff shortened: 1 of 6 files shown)


avro/src/types.rs
```diff
@@ -332,793 +319,931 @@ impl TryFrom<Value> for JsonValue {
...
                 .collect::<Result<Vec<_>, _>>()
                 .map(|v| Self::Object(v.into_iter().collect())),
-            Value::Date(d) => Ok(Self::Number(d.into())),
+            Value::Date(d) => Ok(serde_json::Value::Number(d.into())),
             Value::Decimal(ref d) => <Vec<u8>>::try_from(d)
                 .map(|vec| Self::Array(vec.into_iter().map(|v| v.into()).collect())),
             Value::BigDecimal(ref bg) => {
-                let vec1: Vec<u8> = serialize_big_decimal(bg)?;
-                Ok(Self::Array(vec1.into_iter().map(|b| b.into()).collect()))
+                let vec1: Vec<u8> = crate::bigdecimal::sync::serialize_big_decimal(bg)?;
```
martin-g: @Kriskras99 I need some help here.
This is an impl of TryFrom (i.e. cannot be async) and it needs to use an async function here (`serialize_big_decimal` that uses several `encode_xyz()` async functions).
Ideas how to make it work ?

Kriskras99: You can't. You need to add the `try_from` function to a regular `impl` block instead of implementing the trait. Or as you did, always use the sync version because you're writing to a Vec anyway.

This is one of the reasons I separate the reading from creating the final output in my state machine implementation.

Kriskras99:
Nice, congrats! I do question if `Schema::parse_str` and some other functions really need to be async, as they don't do any IO?

`Schema::parse_str` is only async because `Parser::parse` is async, but that is async because it recursively depends on itself (I think?).

Anyway, really nice work!

martin-g:
Good point!
I also would prefer to minimize the API breaks as much as possible!
But `parse_reader()` will need to stay in `SchemaExt`.

That reminds me: is `SchemaExt` a good name for the IO related functionality ? Or maybe `SchemaIO` ?! Or something else ?

Kriskras99:
I think `SchemaExt` is good, `SchemaIo` and `SchemaIO` just look weird.

FYI, `SchemaExt` is currently not exported. It doesn't show up when I run `cargo doc` (even with `--document-private-items`) so not sure what's going on there.

Kriskras99:
Even better would be `SchemaSync` and `SchemaAsync` traits that you implement directly on `Schema`. That way you don't need the extra struct.

Kriskras99:
As there are some people waiting for a release, might be good to start one. There's nothing important coming that should be included in the next release right?

Kriskras99:
#259 and #260 both are not user visible changes, so don't need to be included in the release if they make rebasing too annoying.

martin-g:
> `Schema::parse_str` is only async because `Parser::parse` is async, but that is async because it recursively depends on itself (I think?).

It appeared to be more complex.

Parser::parse_record calls RecordFieldExt::parse() and it calls Self::resolve_default_value() and this one calls ValueExt::resolve_internal() and this calls bigdecimal::deserialize_big_decimal() and this calls decode_long() that needs to be `async`...

Kriskras99:
That's a shame, I missed that branch

martin-g:
Not sure whether it is bad or not but I needed to replace in several places usage of iterator combinators with `for/while` loops because I needed to call an async method in their closures.
IMO this way it is simpler/clearer than dealing with `futures::join_all!()`.
Now the code is the same for both async and sync versions and SyncA removes the `.await`s for the sync version.

martin-g:
https://github.com/apache/avro-rs/actions/runs/18131946414/job/51600631350?pr=246
```
error: method `len` is never used
  --> avro/src/decimal.rs:84:19
   |
83 | impl Decimal {
   | ------------ method in this implementation
84 |     pub(crate) fn len(&self) -> usize {
   |                   ^^^
   |
   = note: `-D dead-code` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(dead_code)]`
```

but it is used at https://github.com/apache/avro-rs/pull/246/files#diff-9ce454ff2db8b0477e609ee586622825012be35efe856246c9ec0de5936b3879R911

The macros somehow confuse the compiler ...

Kriskras99:
I'm still working on my PR, where I'm currently experimenting with an easier way to do this

PR #255, opened 2025-08-10
refactor: Simplify the UUID decoding logic

Kriskras99:
In the original implementation we first read the length and then read the remainder either as `Fixed` if the length is 16 and `String` if it is not. It uses a temporary `Vec` to be able to backtrack if `Fixed` is wrong.

The backtrack can only happen if it fails to read a `Fixed` or if it gets a different type than `Fixed` from `decode_internal`. The last case can never happen, as `decode_internal` always returns the type corresponding to the schema. The first case can only happen if there are not enough bytes, because any byte pattern is valid for `Fixed`. If `Fixed` cannot read enough bytes, than `String` won't be able to read enough bytes either.

I did not add a fallback from `Uuid::from_slice` to `Uuuid::parse_str` as an UUID string is at least 32 characters, so it would fail anyway.

### DIFF
--- a/avro/src/decode.rs
+++ b/avro/src/decode.rs
@@ -20,7 +20,6 @@ use crate::{
     bigdecimal::deserialize_big_decimal,
     decimal::Decimal,
     duration::Duration,
-    encode::encode_long,
     error::Details,
     schema::{
         DecimalSchema, EnumSchema, FixedSchema, Name, Namespace, RecordSchema, ResolvedSchema,
@@ -33,7 +32,6 @@ use std::{
     borrow::Borrow,
     collections::HashMap,
     io::{ErrorKind, Read},
-    str::FromStr,
 };
 use uuid::Uuid;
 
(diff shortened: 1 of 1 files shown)


avro/src/decode.rs
```diff
@@ -123,61 +121,18 @@ pub(crate) fn decode_internal<R: Read, S: Borrow<Schema>>(
...
-                reader,
-            )? {
-                Value::String(ref s) => {
-                    Uuid::from_str(s).map_err(|e| Details::ConvertStrToUuid(e).into())
-                }
-                value => Err(Error::new(Details::GetUuidFromStringValue(value))),
+            let Value::Bytes(bytes) =
+                decode_internal(&Schema::Bytes, names, enclosing_namespace, reader)?
+            else {
+                // Calling decode_internal with Schema::Bytes can only return a Value::Bytes or an error
```
martin-g, line 127: do we need the `else` at all ?

Kriskras99: Yes, otherwise the compiler will throw an error as it can't prove that this is the only option

Issue #252, opened 2025-08-08
Canonical form without default value makes it impossible to automatically generate a schema that is compatible with the previous one


leo-vital:
I have a struct like this:

```rust
#[derive(AvroSchema)]
struct Something {
   f: String
}
```

Recently I have had a problem with the schema evolution using the `BACKWARD` compatibility configuration in the Schema Registry. I first create an initial schema version just fine, but then when I add a field whose type is `Option<String>` to the struct above and try to encode it using schema-registry-converter's `encode_struct` function, I see this error in the Schema Registry:

```
io.confluent.kafka.schemaregistry.exceptions.IncompatibleSchemaException: [{errorType:'READER_FIELD_MISSING_DEFAULT_VALUE', description:'The field 'field_name' at path '/fields/6' in the new schema has no default value and is missing in the old schema', additionalInfo:'field_name'},...
```

Looking at it deeper, I noticed that the main problem seems to be the fact that the canonical form does not include a default value. Even when I mark the field with `#[avro(default="null")]` the same behavior can be observed. Even though I can see that `T::get_schema()` returns a schema in which the field contains a default value, when I get the schema's canonical form there is no default value.

My understanding is that this is by design due to the STRIP rule [here](https://github.com/apache/avro-rs/blob/d33d60014563cbe8d7b9050b9362e8fe89e1259a/avro/src/schema.rs#L2353). But if that is the case, then how exactly can I get a retro-compatible schema when adding an optional field? Is that simply not supported and I need to manually create a schema version in the Schema Registry?

martin-g:
You should ask this question at dev@avro.apache.org mailing list.

The strip rule is described in the Avro specification - https://avro.apache.org/docs/1.12.0/specification/#parsing-canonical-form-for-schemas
```
[STRIP] Keep only attributes that are relevant to parsing data, which are: type, name, fields, symbols, items, values, size. Strip all others (e.g., doc and aliases).
```

leo-vital:
> You should ask this question at [dev@avro.apache.org](mailto:dev@avro.apache.org) mailing list.
> 
> The strip rule is described in the Avro specification - https://avro.apache.org/docs/1.12.0/specification/#parsing-canonical-form-for-schemas
> 
> ```
> [STRIP] Keep only attributes that are relevant to parsing data, which are: type, name, fields, symbols, items, values, size. Strip all others (e.g., doc and aliases).
> ```

Thank you, I'll do that.

[...]

PR #260, opened 2025-08-12
Issue #252: Add an `Option<String>` field with a `#[avro(default = "null")]`

martin-g:
Extend the test case with an Option field with a default.

https://github.com/apache/avro-rs/issues/252#issuecomment-3177888121

### DIFF
--- a/avro/src/schema_equality.rs
+++ b/avro/src/schema_equality.rs
@@ -240,7 +240,7 @@ pub fn set_schemata_equality_comparator(
 pub(crate) fn compare_schemata(schema_one: &Schema, schema_two: &Schema) -> bool {
     SCHEMATA_COMPARATOR_ONCE
         .get_or_init(|| {
-            debug!("Going to use the default schemata equality comparator: SpecificationEq.",);
+            debug!("Going to use the default schemata equality comparator: StructFieldEq.",);
             Box::new(StructFieldEq {
                 include_attributes: false,
             })


avro_derive/tests/derive.rs
```diff
@@ -1389,6 +1389,9 @@ mod test_derive {
 
             #[avro(default = r#""Foo""#)]
             myenum: MyEnum,
+
+            #[avro(default = r#"null"#)]
```
Kriskras99: The `r##` is not necessary as there are no quotes in the string

Kriskras99 (approved):

avro_derive/tests/derive.rs
```suggestion
            #[avro(default = "null")]
```

Kriskras99:
Shouldn't only parse `r#""{something}""#` as a string? `"{something}"` shoud always be a type or integer literal right?

Kriskras99:
I think the bug is in the schema generated. It should be `"type": null` not `"type": "null"`. See the field default values: https://avro.apache.org/docs/1.11.1/specification/#schema-record

Kriskras99:
(Note that when a [default value](https://avro.apache.org/docs/1.11.1/specification/#schema-record) is specified for a record field whose type is a union, the type of the default value must match the first element of the union. Thus, for unions containing “null”, the “null” is usually listed first, since the default value of such unions is typically null.)

Kriskras99:
But that should generate a `"default": null` which should make this test fail: https://github.com/apache/avro-rs/pull/260/files#diff-2e1b9de8fab3f41491702f9d75b536a4a208002e0d84c489c8404a6d8bb7ec45R1452

martin-g:
Yes! There is a bug somewhere!

https://github.com/apache/avro-rs/blob/23a9af70b2db35d9bb30344efeb9012b487e5309/avro/src/schema.rs#L7203-L7226 produces:
```
fields: [
                                RecordField {
                                    name: "order_number",
                                    doc: None,
                                    aliases: None,
                                    default: Some(
                                        Null,
                                    ),
                                    schema: Union(
                                        UnionSchema {
                                            schemas: [
                                                Null,
                                                String,
                                            ],
                                            variant_index: {
                                                Null: 0,
                                                String: 1,
                                            },
                                        },
                                    ),
```
so it seems to be somewhere in the derive macro...

martin-g:
Right! The default equality (SpecificationEq) uses the canonical form of the schema and as we already know the `default` field is not part of it ...

martin-g:
Well, I am not sure whether it is a bug or not.
`StructFieldEq` behaves like `SpecificationEq` and compares only the schemas' names and their fields' schemas.

Kriskras99:
Looking at StructFieldEq it doesn't seem to check the default field

martin-g:
Right!
```diff
git diff
diff --git i/avro/src/schema_equality.rs w/avro/src/schema_equality.rs
index 1097594..83bf969 100644
--- i/avro/src/schema_equality.rs
+++ w/avro/src/schema_equality.rs
@@ -217,7 +217,9 @@ impl StructFieldEq {
             && fields_one
                 .iter()
                 .zip(fields_two.iter())
-                .all(|(f1, f2)| self.compare(&f1.schema, &f2.schema))
+                .all(|(f1, f2)| {
+                    self.compare(&f1.schema, &f2.schema) && f1.default == f2.default
+                })
     }
 }
 
diff --git i/avro_derive/tests/derive.rs w/avro_derive/tests/derive.rs
index a85caa0..aea426f 100644
--- i/avro_derive/tests/derive.rs
+++ w/avro_derive/tests/derive.rs
@@ -1449,7 +1449,7 @@ mod test_derive {
                 {
                     "name":"optional",
                     "type": ["null", "string"],
-                    "default": null
+                    "default": "null"
                 }
             ]
```
detects the issue!
But I am not sure whether it this should be done!

Hopefully someone will explain why the `default` should be `STRIP`ed for canonical form at https://lists.apache.org/thread/j0no5h8sjlyo1xdzw47dftgcn2k84w2h

avro_derive/tests/derive.rs
```diff
@@ -1442,6 +1445,11 @@ mod test_derive {
                         "symbols":["Foo", "Bar", "Baz"]
                     },
                     "default":"Foo"
+                },
+                {
+                    "name":"optional",
+                    "type": ["null", "string"],
+                    "default": null
```
Kriskras99, line 1452: This test doesn't work as intended. Removing this line does not make the test fail.

Kriskras99: Nevermind, this isn't used for the comparison. Comparison is done on line 1482

PR #266, opened 2025-08-13
Include license and notice files in published crates

ankane:
Hi, this PR symlinks `LICENSE` and `NOTICE.txt` so they're included in published crates.


martin-g:
Please do the same for the test helper crate too. Thank you!

Kriskras99:
Instead of doing symlinks, can't we just set it on the workspace Cargo.toml and do `license.workspace = true`?
I think that's better than a bunch of symlinks

martin-g:
> So is it even necessary to include the `LICENSE` file? The license is specified in the `Cargo.toml`.

Yes. Apache Software Foundation rules mandate both LICENSE and NOTICE files to be part of the distribution. 
This is valid for the **source** distributions and Rust crates are sources.

Kriskras99:
Check, then I think this PR is the right way to do it. (We could only symlink `NOTICE.txt` and use `license-file` for `LICENSE`, but because it's a normal license it's better to use the `license` field.

@ankane sorry for all the confusion. 
@martin-g looking at https://infra.apache.org/licensing-howto.html it seems it's preferred if we rename `NOTICE.txt` to `NOTICE`, should we also do that? And there's a small grammar mistake in the `NOTICE.txt` file: `donated to Apache Avro project at 2020.` should be `donated to the Apache Avro project in 2020.`

Kriskras99:
Thank you for the contribution, @ankane!

ankane:
Thank you both!

PR #299, opened 2025-09-22
fix: Set -o pipefail in CI

Kriskras99:
By default `curl | sh` will have `sh`'s exit code, with this option `curl`'s exit code is propagated.

### DIFF
--- a/.github/workflows/test-lang-rust-ci.yml
+++ b/.github/workflows/test-lang-rust-ci.yml
@@ -250,6 +250,7 @@ jobs:
           key: ${{ runner.os }}-target-cache1-stable-${{ hashFiles('**/Cargo.lock') }}
 
       - name: Install wasm-pack
+        shell: /usr/bin/bash -o pipefail -e {0}
         run: curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh
 
       - name: Build the Web Assembly demo app


.github/workflows/test-lang-rust-ci.yml
```diff
@@ -250,6 +250,7 @@ jobs:
           key: ${{ runner.os }}-target-cache1-stable-${{ hashFiles('**/Cargo.lock') }}
 
       - name: Install wasm-pack
+        shell: /usr/bin/bash -o pipefail -e {0}
```
kojiromike, line 253: An obscure fact about GitHub Actions is that if you don't set the shell, you get bash without pipefail, but if you set the shell to bash explicitly, you get pipefail implicitly.

martin-g: > you get pipefail implicitly

obscure indeed! 

Still it does not harm to be explicit!

.github/workflows/test-lang-rust-ci.yml
```diff
@@ -250,6 +250,7 @@ jobs:
...
         run: curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh
```
martin-g:
```suggestion
        run: curl --retry 5 https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh
```

martin-g (approved):

.github/workflows/test-lang-rust-ci.yml
kojiromike: agreed :)

PR #311, opened 2025-10-09
Support serde flatten annotation for schema derivation and de/serialization

### LINKED ISSUE #247: missing support for #[serde(flatten)] for derive AvroSchema
I tried to play around with the proc-macro, and was able to make this test pass:
```rust
#[test]
    fn avro_flatten_fields() {
        #[derive(Debug, Serialize, Deserialize, AvroSchema, Clone, PartialEq)]
        struct Nested {
            a: bool,
        }


        #[derive(Debug, Serialize, Deserialize, AvroSchema, Clone, PartialEq)]
        struct Foo {
            #[serde(flatten)]
            nested: Nested,
            b: i32,
        }

        let schema = r#"
        {
            "type":"record",
            "name":"Foo",
            "fields": [
                {
                    "name":"a1",
                    "type":"boolean"
                },
                {
                    "name":"b",
                    "type":"int"
                }
            ]
        }
        "#;

        let schema = Schema::parse_str(schema).unwrap();
        let derived_schema = Foo::get_schema();
        if let Schema::Record(RecordSchema { name, fields, .. }) = &derived_schema {
            assert_eq!("Foo", name.fullname(None));
            for field in fields {
                match field.name.as_str() {
                    "a" | "b" => {
                        // expected
                    },
                    name @ _ => {
                        panic!("Unexpected field name '{name}'")
                    }
                }
            }
        } else {
            panic!(
                "Foo schema must be a record schema: {derived_schema:?}"
            )
        }
        assert_eq!(schema, derived_schema);

       // deserializing not support
       // serde_assert(Foo {
       //     nested: Nested { a: true },
       //     b: 321,
       // });
    }
```

but while running the other tests, it broke the `serde(rename)` related tests since the schema code genrator changed in order to allow nested fields to be pushed, and the tests verify the code generation.

I am having a hard time implementing this correctly mainly due to my lack of familiarity with the project.

Sufflope:
Fixes #247

### DIFF
--- a/avro/src/schema.rs
+++ b/avro/src/schema.rs
@@ -816,6 +816,21 @@ pub struct RecordSchema {
     pub attributes: BTreeMap<String, Value>,
 }
 
+impl RecordSchema {
+    const SERDE_FLATTEN_SUPPORT: &str = "avro_rs_serde_flatten_support";
+
+    pub fn has_serde_flatten_support(&self) -> bool {
+        self.attributes
+            .get(Self::SERDE_FLATTEN_SUPPORT)
+            .is_some_and(|value| value.as_bool().unwrap_or(false))
+    }
+
+    pub fn set_serde_flatten_support(&mut self) {
+        self.attributes
+            .insert(Self::SERDE_FLATTEN_SUPPORT.into(), true.into());
+    }
+}
+
 /// A description of an Enum schema.
 #[derive(bon::Builder, Debug, Clone)]
 pub struct EnumSchema {
--- a/avro/src/writer.rs
+++ b/avro/src/writer.rs
@@ -23,7 +23,8 @@ use crate::{
     headers::{HeaderBuilder, RabinFingerprintHeader},
     schema::{AvroSchema, Name, ResolvedOwnedSchema, ResolvedSchema, Schema},
     ser_schema::SchemaAwareWriteSerializer,
-    types::Value,
+    to_value,
+    types::{Record, Value},
 };
 use serde::Serialize;
 use std::{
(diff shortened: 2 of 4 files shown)


avro/src/schema.rs
```diff
@@ -2456,6 +2456,9 @@ pub mod derive {
     use super::*;
     use std::borrow::Cow;
 
+    pub use crate::writer::SERDE_FLATTEN;
+    pub use serde_json;
```
Kriskras99: Why re-export `serde_json`?

avro/src/writer.rs
```diff
@@ -33,6 +34,8 @@ use std::{
 const DEFAULT_BLOCK_SIZE: usize = 16000;
 const AVRO_OBJECT_HEADER: &[u8] = b"Obj\x01";
 
+pub const SERDE_FLATTEN: &str = "serde-flatten";
```
Kriskras99: If we're going this direction, the name should probably changed to something like `avro_rs_serde_flatten`

avro/src/schema.rs
```diff
@@ -2456,6 +2456,9 @@ pub mod derive {
     use super::*;
     use std::borrow::Cow;
 
+    pub use crate::writer::SERDE_FLATTEN;
```
Kriskras99: This should have a `#[doc(hidden)]`

avro_derive/tests/derive.rs
```diff
@@ -1676,4 +1676,56 @@ mod test_derive {
             panic!("Unexpected schema type for Foo")
         }
     }
+
+    #[test]
+    fn test_serde_flatten() {
```
Kriskras99: Include the PR number in the test name. That makes it easier to search for context when this test starts failing.

avro/src/schema.rs
Sufflope: It is for this line that I had to change further down:

`let mut attributes: std::collections::BTreeMap<String, apache_avro::schema::derive::serde_json::Value> = Default::default();`

My experience writing proc-macros (in internal projects mostly, so not many public occurrences to show for credit) is that you should publicly re-export these external crates that you depend on in the generated code, because otherwise, if a project uses `avro-rs` but does not yet depend on `serde_json`, the macro will generate confusing compilation errors about a missing `serde_json` crate, with the user not finding a single `serde_json` in their source. They would then have to depend on that crate for "no reason" and this could mess e.g. linters which would see it as unused dependency. See [one PR](https://github.com/juhaku/utoipa/pull/1243) I made with this rationale :smile: 

I forgot to add a doc(hidden) on it though.

avro/src/schema.rs
Sufflope: In fact I need it public because, long story short I `impl AvroSchemaComponent` for one of my types and it is a flattened type, so I have to put this attribute manually in my hand-built schema. However I think a best approach would be to add a `pub fn set_serde_flatten_support(&mut self)` to `RecordSchema` which encapsulates these internal details of putting a custom attribute etc.

avro/src/writer.rs
Sufflope: I changed it to my encapsulated methods proposition, hopefully I understood what name change you wanted

PR #326, opened 2025-10-28
RFC - Split serialization logic from writing logic to make concurrent writes faster

brent-statsig:
Working on adding better ability to concurrently process avro - this moves the heavy work of serializing + compressing to not require a `mut` anymore. This allows users to wrap the serialization logic in whatever async/processing runtime they want easily, without introducing any opinions or bloat.

- `serialize_ser` does the heavy lifting, returning a private struct
- `extend_avro_serialized_buffer` accepts this struct, writing it directly to the writer and performing all avro bookkeeping.

Let me know what you think :)  its a lil messy factoring wise, but I think the abstraction is fairly reasonable, effective, and safe.

### DIFF
--- a/avro/src/lib.rs
+++ b/avro/src/lib.rs
@@ -868,6 +868,7 @@ mod decode;
 mod duration;
 mod encode;
 mod reader;
+mod ser;
 mod ser_schema;
 mod util;
 mod writer;
@@ -878,7 +879,6 @@ pub mod rabin;
 pub mod schema;
 pub mod schema_compatibility;
 pub mod schema_equality;
-pub mod ser;
 pub mod types;
 pub mod validator;
 
--- a/avro/src/lib.rs
+++ b/avro/src/lib.rs
@@ -868,7 +868,6 @@ mod decode;
 mod duration;
 mod encode;
 mod reader;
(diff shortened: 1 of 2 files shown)


Kriskras99:
I think this can be useful to have. However I don't entirely agree with the design.
I think the `AvroSerializedBuffer` should have it's own append function. It's also relatively easily now to create a buffer with one schema and use it for another. It would also need functions for appending `Value`s

Maybe more something like this?:
```rust
pub struct AvroSerializedBuffer<'a> {
    schema: &'a Schema,
    // We don't want this to be a reference to the writer, as that would block any writing,
    // But cloning this can be expensive (there's a Vec inside)
    resolved_schema: ResolvedSchema<'a>,
    codec: Codec,
    num_values: usize,
    // We want to do compression in the buffer not in the writer, but want to do it as late as possible
    // for maximum compression. So we need to track what we already have compressed.
    compressed_up_to: usize,
    buffer: Vec<u8>,
}
```

Then we can compare the schema, resolved schema and codec when adding it back to the writer. This setup would also allow one to reset the buffer so the allocation can be reused.
@martin-g what do you think?

PR #328, opened 2025-10-29
feat!: resolve schema(ta) in the builder of `Writer`

Kriskras99:
This moves the resolving error to the place where it should happen and simplifies `Writer` functions.

This is a breaking change, as the builder now returns a `AvroResult`.

The actual relevant code is in the first 400 lines of `avro/src/writer.rs`. All other changes are adding `?` or `.unwrap()` to test code.

### DIFF
--- a/avro/src/writer.rs
+++ b/avro/src/writer.rs
@@ -38,49 +38,72 @@ const AVRO_OBJECT_HEADER: &[u8] = b"Obj\x01";
 /// It is critical to call flush before `Writer<W>` is dropped. Though dropping will attempt to flush
 /// the contents of the buffer, any errors that happen in the process of dropping will be ignored.
 /// Calling flush ensures that the buffer is empty and thus dropping will not even attempt file operations.
-#[derive(bon::Builder)]
 pub struct Writer<'a, W: Write> {
     schema: &'a Schema,
     writer: W,
-    #[builder(skip)]
-    resolved_schema: Option<ResolvedSchema<'a>>,
-    #[builder(default = Codec::Null)]
+    resolved_schema: ResolvedSchema<'a>,
     codec: Codec,
-    #[builder(default = DEFAULT_BLOCK_SIZE)]
     block_size: usize,
-    #[builder(skip = Vec::with_capacity(block_size))]
     buffer: Vec<u8>,
-    #[builder(skip)]
     num_values: usize,
-    #[builder(default = generate_sync_marker())]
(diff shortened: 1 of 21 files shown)


avro/src/writer.rs
```diff
@@ -366,9 +366,10 @@ impl<'a, W: Write> Writer<'a, W> {
         let mut this = ManuallyDrop::new(self);
 
         // Extract every member that is not Copy and therefore should be dropped
-        let _resolved_schema = std::mem::take(&mut this.resolved_schema);
         let _buffer = std::mem::take(&mut this.buffer);
         let _user_metadata = std::mem::take(&mut this.user_metadata);
+        // SAFETY: resolved schema is not accessed after this and won't be dropped because of ManuallyDrop
+        unsafe { std::ptr::drop_in_place(&mut this.resolved_schema) };
```
martin-g, line 372: Why this does not use `std::mem::take()` as the others above ?

martin-g: Did I ask before why the `codec` is not dropped here ? I don't remember now

avro/tests/shared.rs
```diff
@@ -115,7 +115,7 @@ fn test_folder(folder: &str) -> Result<(), ErrorsDesc> {
         let reader =
             Reader::with_schema(&schema, BufReader::new(&file)).expect("Can't read data.avro");
 
-        let mut writer = Writer::with_codec(&schema, Vec::new(), Codec::Null);
+        let mut writer = Writer::with_codec(&schema, Vec::new(), Codec::Null).unwrap();
```
martin-g, line 118:
```suggestion
        let mut writer = Writer::with_codec(&schema, Vec::new(), Codec::Null)?;
```

Kriskras99: This doesn't work as an Avro Error cannot be turned into a ErrorsDesc

avro/src/writer.rs
Kriskras99: `Codec` is `Copy` so doesn't need to be dropped.
We can't use `std::mem::take` as `ResolvedSchema` doesn't implement `Default` (`take` replaces the value at the location with the default value). It used to work because it was an `Option` and that does implement `Default`.

martin-g (approved):

martin-g (changes requested):
Is Codec really a Copy ?
Some of the codecs have Settings? Don't they allocate heap memory ?

PR #335, opened 2025-11-11
feat: Make compression_level() pub

EmilyMatt:
All other compressions settings make the struct field itself pub, I'm guessing this isn't done here to streamline the flow into the miniz_oxide crate which accepts a u8 then converts it to i32.
I suggest just making this function pub at least, so we can get the current compression_level, if not modify it.


martin-g:
Thank you, @EmilyMatt !

PR #336, opened 2025-11-12
Small fixes

Kriskras99:
These are some small things I ran into when working on async support.

### DIFF
--- a/avro/src/schema_equality.rs
+++ b/avro/src/schema_equality.rs
@@ -140,17 +140,17 @@ impl SchemataEq for StructFieldEq {
             }
             (Schema::Union(_), _) => false,
             (
-                Schema::Decimal(DecimalSchema { precision: precision_one, scale: scale_one, ..}),
-                Schema::Decimal(DecimalSchema { precision: precision_two, scale: scale_two, .. })
+                Schema::Decimal(DecimalSchema { precision: precision_one, scale: scale_one, inner: inner_one }),
+                Schema::Decimal(DecimalSchema { precision: precision_two, scale: scale_two, inner: inner_two })
             ) => {
-                precision_one == precision_two && scale_one == scale_two
+                precision_one == precision_two && scale_one == scale_two && self.compare(inner_one, inner_two)
             }
             (Schema::Decimal(_), _) => false,
             (
                 Schema::Array(ArraySchema { items: items_one, ..}),
                 Schema::Array(ArraySchema { items: items_two, ..})
             ) => {
-                items_one == items_two
+                self.compare(items_one, items_two)
             }
             (Schema::Array(_), _) => false,
             (
@@ -177,7 +177,7 @@ impl StructFieldEq {
             && fields_one
                 .iter()
                 .zip(fields_two.iter())
-                .all(|(f1, f2)| self.compare(&f1.schema, &f2.schema))
+                .all(|(f1, f2)| f1.name == f2.name && self.compare(&f1.schema, &f2.schema))
     }
 }
 


avro/src/util.rs
```diff
@@ -100,7 +100,8 @@ pub(crate) fn zag_i64<R: Read>(reader: &mut R) -> AvroResult<i64> {
     })
 }
 
-fn encode_variable<W: Write>(mut z: u64, mut writer: W) -> AvroResult<usize> {
+fn encode_variable<W: Write>(n: i64, mut writer: W) -> AvroResult<usize> {
+    let mut z = (((n << 1) ^ (n >> 63)) as u64).to_le();
```
martin-g: This is to be complaint with `a double is written as 8 bytes. The double is converted into a 64-bit integer using a method equivalent to Java’s [doubleToRawLongBits](https://docs.oracle.com/javase/8/docs/api/java/lang/Double.html#doubleToRawLongBits-double-) and then encoded in little-endian format.` from the [specification](https://avro.apache.org/docs/1.12.0/specification/#binary-encoding), right ?

avro/src/util.rs
martin-g: Do we need to do something like `i = u64::from_le(i)` here too ?
On a big endian machine the number will be wrong now, no ?

avro/src/schema_equality.rs
```diff
@@ -54,160 +54,120 @@ pub struct StructFieldEq {
...
-                items: items_two, ..
-            }) = schema_two
-            {
-                return items_one == items_two;
+            (Schema::Decimal(_), _) => false,
+            (
+                Schema::Array(ArraySchema { items: items_one, ..}),
+                Schema::Array(ArraySchema { items: items_two, ..})
+            ) => {
+                items_one == items_two
```
```suggestion
                self.compare(items_one, items_two)
```

avro/src/schema_equality.rs
```suggestion
                .all(|(f1, f2)| f1.name == f2.name && self.compare(&f1.schema, &f2.schema))
```

avro/src/schema_equality.rs
```diff
@@ -54,160 +54,120 @@ pub struct StructFieldEq {
...
-                ..
-            }) = schema_two
-            {
-                return precision_one == precision_two && scale_one == scale_two;
+            (Schema::Union(_), _) => false,
+            (
+                Schema::Decimal(DecimalSchema { precision: precision_one, scale: scale_one, ..}),
+                Schema::Decimal(DecimalSchema { precision: precision_two, scale: scale_two, .. })
+            ) => {
+                precision_one == precision_two && scale_one == scale_two
```
martin-g: Do we need to care about the inner schema of the decimal (fixed vs. bytes)?

```suggestion
                precision_one == precision_two && scale_one == scale_two && self.compare(inner_one, inner_two)
```
I am not sure. Maybe yes.
If YES then please also add `inner_one` and `inner_two` at lines 143 and 144

avro/src/util.rs
Kriskras99: No, `encode_variable` is purely used for zigzag encoding `int` and `long` types.
The bit twiddling is to do the zigzag, while the remainder of the function is for doing the varint encoding.
I've moved the bit twiddling back to `zig_i64`, and documented why they are two different functions

avro/src/util.rs
Kriskras99: Yes, you're right.

avro/src/schema_equality.rs
Kriskras99: The first thing `self.compare` does is checking the name, so I think this optimization is unnecessary (it saves on a function call, but the function is in L1 cache so the impact should be minimal)

avro/src/schema_equality.rs
Kriskras99: I think so too, as a `Decimal(Bytes)` and `Decimal(Fixed)` are not compatible

avro/src/schema_equality.rs
martin-g: RecordField::name != Schema::name

Kriskras99: You're totally right, added!

avro/src/util.rs
```diff
@@ -100,25 +105,34 @@ pub(crate) fn zag_i64<R: Read>(reader: &mut R) -> AvroResult<i64> {
     })
 }
 
-fn encode_variable<W: Write>(mut z: u64, mut writer: W) -> AvroResult<usize> {
+/// Write the number as a varint to the writer.
+///
+/// Note: this function does not do zigzag encoding, for that see [`zig_i32`] and [`zig_i64`].
+fn encode_variable<W: Write>(mut zigzagged: u64, mut writer: W) -> AvroResult<usize> {
+    // Ensure the number is little endian for the varint encoding (no-op on LE systems)
+    zigzagged = zigzagged.to_le();
```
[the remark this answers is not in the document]
Kriskras99: I don't think this is correct. Encoding a varint on BE and then decoding that on LE should give the same result (but the specification doesn't actually specify which endianness to use).

PR #337, opened 2025-11-20
fix: Choose correct schema when serializing an Option<Enum>

PookieBuns:
There is a bug currently in serialization in which if you are serializing an `Option<Enum>` for a Union Schema (with null), it will automatically try to use the first not-null schema if you choose a not-null enum. This fix allows the serializer to correctly resolve the proper schema to serialize the Enum variant with.

## Notes
- This changes the error message for the Union Long test. Imo its now better because it properly shows the expected type vs the attempted type instead of showing the Union as the error. Because of this, I had to change the test assert message to allow it to pass
- Serializing unions containing records are a bit funky right now, since structs in rust should be in Capital case but if the name is not in Capital case, you have to rename the struct itself, instead of calling ex. `camelCase` on the Enum. However, this is pre-existing behavior so its not a regression from pre-existing behavior

Please let me know if there are any questions or things to address!

### DIFF
--- a/avro/src/ser_schema.rs
+++ b/avro/src/ser_schema.rs
@@ -2747,7 +2747,7 @@ mod tests {
     }
 
     #[test]
-    fn test_serialize_union_record_variant() -> TestResult {
+    fn avro_rs_337_serialize_union_record_variant() -> TestResult {
         let schema = Schema::parse_str(
             r#"{
             "type": "record",
@@ -2794,7 +2794,6 @@ mod tests {
             bar: String,
         }
 
-        assert!(!crate::util::is_human_readable());
         let mut buffer: Vec<u8> = Vec::new();
         let rs = ResolvedSchema::try_from(&schema)?;
         let mut serializer =
@@ -2824,7 +2823,7 @@ mod tests {
     }
 
     #[test]
-    fn test_serialize_option_union_record_variant() -> TestResult {
+    fn avro_rs_337_serialize_option_union_record_variant() -> TestResult {
         let schema = Schema::parse_str(
             r#"{
             "type": "record",
@@ -2872,7 +2871,6 @@ mod tests {
             bar: String,
         }
 
-        assert!(!crate::util::is_human_readable());
         let mut buffer: Vec<u8> = Vec::new();
         let rs = ResolvedSchema::try_from(&schema)?;
         let mut serializer =


Kriskras99:
Ah, that would map to a nested Union which is not allowed by spec, so we don't need to worry about that.

avro/src/ser_schema.rs
```diff
@@ -2770,4 +2745,161 @@ mod tests {
 
         Ok(())
     }
+
+    #[test]
+    fn test_serialize_union_record_variant() -> TestResult {
```
martin-g:
```suggestion
    fn avro_rs_337_serialize_union_record_variant() -> TestResult {
```

avro/src/ser_schema.rs
```diff
@@ -2770,4 +2745,161 @@ mod tests {
...
+            foo: String,
+        }
+
+        #[derive(Serialize)]
+        #[serde(rename = "innerRecordBar")]
+        struct InnerRecordBar {
+            bar: String,
+        }
+
+        assert!(!crate::util::is_human_readable());
```
martin-g:
```suggestion
        assert!(!crate::util::is_human_readable(), "Some nice message here please");
```

avro/src/ser_schema.rs
```diff
@@ -2770,4 +2745,161 @@ mod tests {
...
+        int_record.serialize(&mut serializer)?;
+        let string_record = TestRecord {
+            inner_union: InnerUnion::StringField(String::from("string")),
+        };
+        string_record.serialize(&mut serializer)?;
+        Ok(())
+    }
+
+    #[test]
+    fn test_serialize_option_union_record_variant() -> TestResult {
```
martin-g:
```suggestion
    fn avro_rs_337_serialize_option_union_record_variant() -> TestResult {
```

avro/src/ser_schema.rs
```diff
@@ -2770,4 +2745,161 @@ mod tests {
...
+            foo: String,
+        }
+
+        #[derive(Serialize)]
+        #[serde(rename = "innerRecordBar")]
+        struct InnerRecordBar {
+            bar: String,
+        }
+
+        assert!(!crate::util::is_human_readable());
```
martin-g:
```suggestion
        assert!(!crate::util::is_human_readable(), "Some nice message here please");
```

avro/src/ser_schema.rs
PookieBuns: Updated!

avro/src/ser_schema.rs
PookieBuns: I just removed it since it doesn't apply to this test because we aren't asserting bytes

avro/src/ser_schema.rs
PookieBuns: Updated!

avro/src/ser_schema.rs
PookieBuns: Similar to above, removed

PR #338, opened 2025-11-21
fix: Support deserialization of `BigDecimal` fields in Avro records when using the "the serde way"

eft-prima:
This pull request adds support for the deserialization of `BigDecimal` fields in Avro records when using the "the serde way".

Currently, the following fails when wrapping the call to `apache_avro::from_value(value)` with the error `Error { details: Failed to deserialize Avro value into value: incorrect value of type: BigDecimal }`. 

```rust
    #[test]
    fn test_big_decimal() {
        #[derive(Clone, Eq, PartialEq, Debug, Default, serde::Deserialize, serde::Serialize)]
        struct Test {
            big_decimal: apache_avro::BigDecimal,
        }

        let schema = {
            let value = r#"
                {
                    "name": "Test",
                    "type": "record",
                    "fields": [
                        {
                            "name": "big_decimal",
                            "type": "bytes",
                            "logicalType": "big-decimal"
                        }
                    ]
                }
            "#;
            Schema::parse_str(value).unwrap()
        };

        let test = Test::default();

        let mut writer = Writer::new(&schema, Vec::default());
        writer.append_ser(test.clone()).unwrap();

        let encoded = writer.into_inner().unwrap();

        let mut reader = Reader::with_schema(&schema, &encoded[..]).unwrap();

        if let Some(Ok(ref value)) = reader.next() {
            let decoded: Test = apache_avro::from_value(value).unwrap();
            assert_eq!(test, decoded);
        }
    }
```

I believe this is because there is no branch for `Value::BigDecimal` in `fn deserialize_any<V>(self, visitor: V) -> Result<V::Value, Self::Error>`.

I've added one using the same via string approach that's used for `Value::Uuid`.

### DIFF
--- a/avro/src/de.rs
+++ b/avro/src/de.rs
@@ -351,7 +351,7 @@ impl<'de> de::Deserializer<'de> for &Deserializer<'de> {
                 Value::Bytes(ref bytes) | Value::Fixed(_, ref bytes) => visitor.visit_bytes(bytes),
                 Value::Decimal(ref d) => visitor.visit_bytes(&d.to_vec()?),
                 Value::Enum(_, ref s) => visitor.visit_borrowed_str(s),
-                Value::BigDecimal(big_decimal) => {
+                Value::BigDecimal(ref big_decimal) => {
                     visitor.visit_str(big_decimal.to_plain_string().as_str())
                 }
                 _ => Err(de::Error::custom(format!(


avro/src/bigdecimal.rs
```diff
@@ -182,6 +182,46 @@ mod tests {
         Ok(())
     }
 
+    #[test]
+    fn test_avro_3779_record_with_bg_ser() -> TestResult {
```
Kriskras99: This issue number is not relevant for this test, please use `avro_rs_338` so it points to this PR

Kriskras99 (approved):
Looks good

avro/src/bigdecimal.rs
martin-g:
```suggestion
    fn avro_rs_338_deserialize_serde_way() -> TestResult {
```

martin-g (approved):

PR #339, opened 2025-11-22
feat: Support `fixed` as type for `uuid`

Kriskras99:
In Avro 1.12 official support for the `fixed` type was added for the logical type `uuid`. avro-rs had a different interpretation of that and would rewrite it to a `bytes` type.

To remain compatible with the legacy implementation, avro-rs now supports `string`, `fixed`, and `bytes`. `bytes` does not match spec, but does allow users to read already encoded data.

This is a severe breaking change, although only for people who only use avro-rs (the legacy encoding did not, and still does not work with other languages).

I've also reworked `Schema::Decimal` to use a similar setup as `Schema::Uuid` which saves a `Box` and does not make the type bigger.

### DIFF
--- a/avro/src/encode.rs
+++ b/avro/src/encode.rs
@@ -167,7 +167,12 @@ pub(crate) fn encode_internal<W: Write, S: Borrow<Schema>>(
             }
             _ => Err(Details::EncodeValueAsSchemaError {
                 value_kind: ValueKind::Uuid,
-                supported_schema: vec![SchemaKind::Uuid, SchemaKind::Fixed, SchemaKind::Bytes],
+                supported_schema: vec![
+                    SchemaKind::Uuid,
+                    SchemaKind::Fixed,
+                    SchemaKind::Bytes,
+                    SchemaKind::String,
+                ],
             }
             .into()),
         },
--- a/avro/src/schema.rs
+++ b/avro/src/schema.rs
@@ -110,7 +110,7 @@ pub enum Schema {
     /// Logical type which represents `Decimal` values without predefined scale.
     /// The underlying type is serialized and deserialized as `Schema::Bytes`
     BigDecimal,
-    /// A universally unique identifier, annotating a string.
+    /// A universally unique identifier, annotating a string, bytes or fixed.
     Uuid(UuidSchema),
     /// Logical type which represents the number of days since the unix epoch.
     /// Serialization format is `Schema::Int`.
@@ -487,7 +487,13 @@ impl<'s> ResolvedSchema<'s> {
                         self.resolve(vec![schema], enclosing_namespace, known_schemata)?
                     }
                 }
-                Schema::Enum(EnumSchema { name, .. }) | Schema::Fixed(FixedSchema { name, .. }) => {
+                Schema::Enum(EnumSchema { name, .. })
+                | Schema::Fixed(FixedSchema { name, .. })
+                | Schema::Uuid(UuidSchema::Fixed(FixedSchema { name, .. }))
+                | Schema::Decimal(DecimalSchema {
+                    inner: InnerDecimalSchema::Fixed(FixedSchema { name, .. }),
+                    ..
+                }) => {
                     let fully_qualified_name = name.fully_qualified_name(enclosing_namespace);
                     if self
                         .names_ref
@@ -574,7 +580,13 @@ pub(crate) fn resolve_names(
             }
             Ok(())
         }
-        Schema::Enum(EnumSchema { name, .. }) | Schema::Fixed(FixedSchema { name, .. }) => {
+        Schema::Enum(EnumSchema { name, .. })
+        | Schema::Fixed(FixedSchema { name, .. })
+        | Schema::Uuid(UuidSchema::Fixed(FixedSchema { name, .. }))
+        | Schema::Decimal(DecimalSchema {
+            inner: InnerDecimalSchema::Fixed(FixedSchema { name, .. }),
+            ..
+        }) => {
             let fully_qualified_name = name.fully_qualified_name(enclosing_namespace);
             if names
                 .insert(fully_qualified_name.clone(), schema.clone())
@@ -938,7 +950,7 @@ impl TryFrom<Schema> for InnerDecimalSchema {
 pub enum UuidSchema {
     /// [`Schema::Bytes`] with size of 16.
     ///
-    /// This is according to specification, but was what happened in `0.21.0` and earlier when
+    /// This is not according to specification, but was what happened in `0.21.0` and earlier when
     /// a schema with logical type `uuid` and inner type `fixed` was used.
     Bytes,
     /// [`Schema::String`].
@@ -1617,9 +1629,7 @@ impl Parser {
                                 Ok((precision, scale)) => Ok(Schema::Decimal(DecimalSchema {
                                     precision,
                                     scale,
-                                    inner: inner.try_into().unwrap_or_else(|_| {
-                                        unreachable!("inner is Fixed or Bytes")
-                                    }),
+                                    inner: inner.try_into()?,
                                 })),
                                 Err(err) => {
                                     warn!("Ignoring invalid decimal logical type: {err}");
@@ -7401,5 +7411,63 @@ mod tests {
         assert_eq!(canonical_form1, canonical_form2);
 
         Ok(())
+    }
+
+    #[test]
+    fn avro_rs_339_schema_ref_uuid() {
+        let schema = Schema::parse_str(
+            r#"{
+            "name": "foo",
+            "type": "record",
+            "fields": [
+                {
+                    "name": "a",
+                    "type": {
+                        "type": "fixed",
+                        "size": 16,
+                        "logicalType": "uuid",
+                        "name": "bar"
+                    }
+                },
+                {
+                    "name": "b",
+                    "type": "bar"
+                }
+            ]
+        }"#,
+        )
+        .unwrap();
+        let _resolved = ResolvedSchema::try_from(&schema).unwrap();
+        let _resolved_owned = ResolvedOwnedSchema::try_from(schema).unwrap();
+    }
+
+    #[test]
+    fn avro_rs_339_schema_ref_decimal() {
+        let schema = Schema::parse_str(
+            r#"{
+            "name": "foo",
+            "type": "record",
+            "fields": [
+                {
+                    "name": "a",
+                    "type": {
+                        "type": "fixed",
+                        "size": 16,
+                        "logicalType": "decimal",
+                        "precision": 4,
+                        "scale": 2,
+                        "name": "bar"
+                    }
+                },
+                {
+                    "name": "b",
+                    "type": "bar"
+                }
+            ]
+        }"#,
+        )
+        .unwrap();
+        let _resolved = ResolvedSchema::try_from(&schema).unwrap();
+        let _resolved_owned = ResolvedOwnedSchema::try_from(schema).unwrap();
     }
 }
--- a/avro/src/schema_compatibility.rs
+++ b/avro/src/schema_compatibility.rs
@@ -1181,7 +1181,6 @@ mod tests {
         ];
 
         for (reader, writer) in compatible_schemas {
-            println!("{reader:?}, {writer:?}");
             SchemaCompatibility::can_read(&writer, &reader).unwrap();
         }
     }
--- a/avro/src/serde/ser_schema.rs
+++ b/avro/src/serde/ser_schema.rs
@@ -1128,7 +1128,7 @@ impl<'s, W: Write> SchemaAwareWriteSerializer<'s, W> {
         match schema {
             Schema::String
             | Schema::Bytes
-            | Schema::Uuid(UuidSchema::Bytes)
+            | Schema::Uuid(UuidSchema::Bytes | UuidSchema::String)
             | Schema::BigDecimal => self.write_bytes(value),
             Schema::Fixed(fixed_schema) | Schema::Uuid(UuidSchema::Fixed(fixed_schema)) => {
                 if value.len() == fixed_schema.size {
@@ -1190,7 +1190,7 @@ impl<'s, W: Write> SchemaAwareWriteSerializer<'s, W> {
                     match variant_schema {
                         Schema::String
                         | Schema::Bytes
-                        | Schema::Uuid(UuidSchema::Bytes)
+                        | Schema::Uuid(_)
                         | Schema::BigDecimal
                         | Schema::Fixed(_)
                         | Schema::Duration
--- a/avro/src/types.rs
+++ b/avro/src/types.rs
@@ -480,8 +480,13 @@ impl Value {
                     None
                 }
             }
-            (&Value::Fixed(n, _), &Schema::Uuid(UuidSchema::Fixed(_))) => {
-                if n != 16 {
+            (&Value::Fixed(n, _), Schema::Uuid(UuidSchema::Fixed(size, ..))) => {
+                if size.size != 16 {
+                    Some(format!(
+                        "The schema's size ('{}') must be exactly 16 to be a Uuid",
+                        size.size
+                    ))
+                } else if n != 16 {
                     Some(format!(
                         "The value's size ('{n}') must be exactly 16 to be a Uuid"
                     ))
--- a/avro/src/writer.rs
+++ b/avro/src/writer.rs
@@ -1006,7 +1006,7 @@ mod tests {
             &Schema::Decimal(DecimalSchema {
                 precision: 20,
                 scale: 5,
-                inner: inner.clone(),
+                inner,
             }),
             Value::Decimal(Decimal::from(value.clone())),
             &Schema::Fixed(fixed),
--- a/avro/tests/serde_human_readable_true.rs
+++ b/avro/tests/serde_human_readable_true.rs
@@ -20,7 +20,7 @@ fn avro_rs_53_uuid_with_string_true() -> TestResult {
                           "type" : {
                             "type" : "string",
                             "logicalType" : "uuid",
-                            "name": "FixedUUID"
+                            "name": "StringUUID"
                           }
                         } ]
                      }"#,


avro/src/types.rs
```diff
@@ -478,6 +480,15 @@ impl Value {
                     None
                 }
             }
+            (&Value::Fixed(n, _), &Schema::Uuid(UuidSchema::Fixed(_))) => {
+                if n != 16 {
+                    Some(format!(
+                        "The value's size ('{n}') must be exactly 12 to be a Duration"
```
```suggestion
                        "The value's size ('{n}') must be exactly 16 to be a Uuid"
```

avro/src/schema_compatibility.rs
```diff
@@ -426,13 +427,48 @@ impl SchemaCompatibility {
...
+                    }
+                    Schema::Uuid(UuidSchema::String) => {
+                        return check_writer_type(
+                            writers_schema,
+                            readers_schema,
+                            vec![r_type, SchemaKind::String],
+                        );
+                    }
+                    Schema::Uuid(UuidSchema::Fixed(FixedSchema {
+                        name: w_name,
```
martin-g: This is the readers schema ^^
Below at line 454 you use again the reader schema

```suggestion
                        name: r_name,
```

avro/src/schema_compatibility.rs
```diff
@@ -426,13 +427,48 @@ impl SchemaCompatibility {
...
+                        size: w_size,
```
martin-g:
```suggestion
                        size: r_size,
```

avro/src/schema_compatibility.rs
```diff
@@ -426,13 +427,48 @@ impl SchemaCompatibility {
...
+                        ..
+                    })) => {
+                        if let Schema::Uuid(UuidSchema::Fixed(FixedSchema {
+                            name: r_name,
```
martin-g:
```suggestion
                            name: w_name,
```

avro/src/schema_compatibility.rs
```diff
@@ -426,13 +427,48 @@ impl SchemaCompatibility {
...
+                            size: r_size,
```
martin-g:
```suggestion
                            size: w_size,
```

avro/src/schema_compatibility.rs
```diff
@@ -426,13 +427,48 @@ impl SchemaCompatibility {
...
+                            ..
+                        })) = readers_schema
```
martin-g:
```suggestion
                        })) = writers_schema
```

avro/src/schema_compatibility.rs
```diff
@@ -426,13 +427,48 @@ impl SchemaCompatibility {
...
+                        {
+                            return (w_name.name == r_name.name && w_size == r_size)
+                                .then_some(())
+                                .ok_or(CompatibilityError::FixedMismatch);
+                        } else if let Schema::Fixed(FixedSchema {
+                            name: r_name,
```
martin-g:
```suggestion
                            name: w_name,
```

avro/src/schema_compatibility.rs
```diff
@@ -426,13 +427,48 @@ impl SchemaCompatibility {
...
+                            size: r_size,
```
martin-g:
```suggestion
                            size: w_size,
```

avro/src/schema_compatibility.rs
```diff
@@ -426,13 +427,48 @@ impl SchemaCompatibility {
...
+                            ..
+                        }) = readers_schema
```
martin-g:
```suggestion
                        }) = writers_schema
```

avro/src/schema_compatibility.rs
```diff
@@ -499,8 +535,15 @@ impl SchemaCompatibility {
...
-            SchemaKind::Uuid => check_reader_type(r_type, SchemaKind::String, w_type),
+            SchemaKind::Bytes => {
+                check_reader_type_multi(r_type, vec![SchemaKind::String, SchemaKind::Uuid], w_type)
+            }
+            SchemaKind::Uuid => check_reader_type_multi(
+                r_type,
+                vec![SchemaKind::String, SchemaKind::Bytes, SchemaKind::Fixed],
+                w_type,
+            ),
+            SchemaKind::Fixed => check_reader_type(r_type, SchemaKind::Uuid, w_type),
```
martin-g:
```suggestion
            SchemaKind::Fixed => check_reader_type_multi(r_type, vec![SchemaKind::Duration, SchemaKind::Uuid], w_type),
```
?

avro/src/decode.rs
```diff
@@ -101,39 +102,66 @@ pub(crate) fn decode_internal<R: Read, S: Borrow<Schema>>(
                 }
             }
         }
-        Schema::Decimal(DecimalSchema { ref inner, .. }) => match &**inner {
-            Schema::Fixed { .. } => {
-                match decode_internal(inner, names, enclosing_namespace, reader)? {
+        Schema::Decimal(DecimalSchema { inner, .. }) => match inner {
+            InnerDecimalSchema::Fixed(fixed) => {
+                match decode_internal(
+                    &Schema::Fixed(fixed.copy_only_size()),
```
martin-g, line 108: Why this copy is needed ?

avro/src/decode.rs
```diff
@@ -101,39 +102,66 @@ pub(crate) fn decode_internal<R: Read, S: Borrow<Schema>>(
...
+            let uuid = Uuid::parse_str(&string).map_err(Details::ConvertStrToUuid)?;
+            Ok(Value::Uuid(uuid))
+        }
+        Schema::Uuid(UuidSchema::Bytes) => {
             let Value::Bytes(bytes) =
                 decode_internal(&Schema::Bytes, names, enclosing_namespace, reader)?
             else {
                 // Calling decode_internal with Schema::Bytes can only return a Value::Bytes or an error
-                unreachable!();
+                unreachable!()
```
martin-g: Maybe move the comment to be a parameter of the `unreachable!("... here ...")` ?!
There are few more such occurrences below.

avro/src/decode.rs
```diff
@@ -101,39 +102,66 @@ pub(crate) fn decode_internal<R: Read, S: Borrow<Schema>>(
...
                 Value::Bytes(bytes) => deserialize_big_decimal(&bytes).map(Value::BigDecimal),
                 value => Err(Details::BytesValue(value).into()),
             }
         }
-        Schema::Uuid => {
+        Schema::Uuid(UuidSchema::String) => {
+            let Value::String(string) =
+                decode_internal(&Schema::String, names, enclosing_namespace, reader)?
+            else {
+                // Calling decode_internal with Schema::String can only return a Value::String or an error
```
martin-g: Actually it may return `Value::Null` too. See https://github.com/apache/avro-rs/blob/fdb0f5d6d6b4d137b371663885b7bf797991a16b/avro/src/decode.rs#L211

avro/src/schema_compatibility.rs
martin-g: This most probably also needs to check the `size`

avro/src/schema.rs
```diff
@@ -111,7 +111,7 @@ pub enum Schema {
     /// The underlying type is serialized and deserialized as `Schema::Bytes`
     BigDecimal,
     /// A universally unique identifier, annotating a string.
-    Uuid,
+    Uuid(UuidSchema),
```
martin-g, line 114: This is a breaking change, so we need to use `feat!: ...` for the PR title/commit message

avro/src/ser_schema.rs
```diff
@@ -982,31 +986,30 @@ impl<'s, W: Write> SchemaAwareWriteSerializer<'s, W> {
...
+                                0 => 0,
+                                _ => value[0],
+                            };
+                            let padding = vec![pad_val; pad];
+                            self.writer
+                                .write(padding.as_slice())
+                                .map_err(Details::WriteBytes)?;
+                            self.writer
+                                .write(value)
+                                .map_err(|e| Details::WriteBytes(e).into())
```
martin-g: This returns only the number of bytes written in the second `.write()` call. IMO it should add the number of the padding length too.

avro/src/encode.rs
martin-g:
```suggestion
                supported_schema: vec![SchemaKind::Bytes, SchemaKind::Fixed, Schema::Uuid(UuidSchema::Bytes)],
```

avro/src/schema_compatibility.rs
Kriskras99: I think that would be best, but that should probably be combined with other improvements to the compatibility checking. I'll open an issue for this.

avro/src/decode.rs
Kriskras99: We could clone the entire `FixedSchema` but for decoding only the size is relevant. (If you clone, you also clone `name`,`aliases`,`doc`,`default`, and `attributes`)

avro/src/decode.rs
Kriskras99: Hmm, you're right. Why are we doing that? We only do it for String and Union.

avro/src/encode.rs
Kriskras99: Unfortunately that doesn't work, I can only add SchemaKind::Uuid (without the subschema specification)

avro/src/decode.rs
martin-g: No idea!
If no tests fail we could remove it and re-add later if we/someone find(s) the reason.

Kriskras99: Removing it for `String` doesn't cause any tests to fail. Removing it for `Union` causes this test to fail:
https://github.com/apache/avro-rs/blob/4b8d3474da670beb731eda94995aff46b51428d0/avro/src/reader.rs#L665
This was added for [AVRO-3240](https://issues.apache.org/jira/browse/AVRO-3240?attachmentOrder=asc) where a user wanted to decode a datum with a newer schema that has additional (optional) fields compared to the writer schema. This is not allowed by the specification, and should not have been added I think as it can silently allow truncated (corrupted) files if a schema has `Union<Null, T>` at the end of the file.

avro/src/schema.rs
```diff
@@ -2539,7 +2618,20 @@ pub mod derive {
     impl_schema!(f32, Schema::Float);
     impl_schema!(f64, Schema::Double);
     impl_schema!(String, Schema::String);
-    impl_schema!(uuid::Uuid, Schema::Uuid);
+    impl_schema!(
+        uuid::Uuid,
+        Schema::Uuid(UuidSchema::Fixed(FixedSchema {
+            name: Name {
+                name: String::new(),
```
martin-g: Empty name is against the specification.
Maybe we should use `Schema::Uuid(UuidSchema::String)` or `Schema::Uuid(UuidSchema::Bytes)` instead ?!

avro/src/schema_compatibility.rs
```diff
@@ -426,13 +427,48 @@ impl SchemaCompatibility {
...
+                        } else if let Schema::Fixed(FixedSchema {
+                            name: w_name,
+                            size: w_size,
+                            ..
+                        }) = writers_schema
+                        {
+                            return (w_name.name == r_name.name && w_size == r_size)
+                                .then_some(())
+                                .ok_or(CompatibilityError::FixedMismatch);
+                        }
```
martin-g, line 463: We need to return an Err in an `else` clause here.

avro/tests/serde_human_readable_true.rs
```diff
@@ -18,8 +18,7 @@ fn avro_rs_53_uuid_with_fixed_true() -> TestResult {
                         "fields" : [ {
                           "name" : "id",
                           "type" : {
-                            "type" : "fixed",
-                            "size" : 16,
+                            "type" : "string",
```
martin-g, line 21: We need to rename the function. Currently it still says `uuid_with_fixed_true`

avro/src/schema.rs
```diff
@@ -894,7 +911,40 @@ pub struct DecimalSchema {
...
+        match value {
+            Schema::Bytes => Ok(InnerDecimalSchema::Bytes),
+            Schema::Fixed(fixed) => Ok(InnerDecimalSchema::Fixed(fixed)),
+            _ => Err(Details::ResolveDecimalSchema(value.into()).into()),
+        }
+    }
+}
+
+/// The inner schema of the Uuid type.
+#[derive(Debug, Clone)]
```
martin-g, line 949:
```suggestion
#[derive(Debug, Clone)]
#[non_exhaustive]
```
Would that help to avoid the `unreachable!()` at https://github.com/apache/avro-rs/pull/339/files#diff-7bb1045091377e3af9ff4a97817d8362470be45ba74a8327211241b30a0f8b89R470 ?
It would be better to have a compilation error when a new variant is added than crashing the app with `unreachable!()`

avro/src/types.rs
```diff
@@ -1870,7 +1886,34 @@ Field with name '"b"' is not a member of the map items"#,
...
+            value
+                .clone()
+                .resolve(&Schema::Uuid(UuidSchema::Bytes))
+                .is_ok()
+        );
+        assert!(
+            value
+                .clone()
+                .resolve(&Schema::Uuid(UuidSchema::Fixed(FixedSchema {
+                    name: Name {
```
martin-g, line 1910:
```suggestion
                    name: Name::new("some dummy name")?
```
instead of using an invalid Name

avro/src/schema_equality.rs
```diff
@@ -544,4 +556,60 @@ mod tests {
         assert_eq!(specification_eq_res, struct_field_eq_res);
         Ok(())
     }
+
+    #[test]
+    fn test_uuid_compare_uuid() -> TestResult {
+        let string = Schema::Uuid(UuidSchema::String);
+        let bytes = Schema::Uuid(UuidSchema::Bytes);
+        let mut fixed_schema = FixedSchema {
+            name: Name {
```
martin-g, line 565:
```suggestion
            name: Name::new("some dummy name")
```

avro/src/schema_equality.rs
```diff
@@ -544,4 +556,60 @@ mod tests {
...
+                name: String::new(),
+                namespace: None,
+            },
+            aliases: None,
+            doc: None,
+            size: 16,
+            default: None,
+            attributes: Default::default(),
+        };
+        let fixed = Schema::Fixed(fixed_schema.clone());
```
```suggestion
        let fixed = Schema::Uuid(UuidSchema::Fixed(fixed_schema.clone()));
```

avro/src/schema_equality.rs
```diff
@@ -544,4 +556,60 @@ mod tests {
...
+        fixed_schema
+            .attributes
+            .insert("Something".to_string(), Value::Null);
+        let fixed_different = Schema::Fixed(fixed_schema);
```
```suggestion
        let fixed_different = Schema::Uuid(UuidSchema::Fixed(fixed_schema.clone()));
```

avro/src/schema.rs
Kriskras99: No, the problem is with SchemaKind and Schema. A possible change is replacing the `_ => unreachable!(..)` with:
```
Schema::Null
| Schema::Boolean
| Schema::Int
| Schema::Long
| Schema::Float
| Schema::Double
| Schema::Bytes
| Schema::String
| Schema::Array(_)
| Schema::Map(_)
| Schema::Union(_)
| Schema::Record(_)
| Schema::Enum(_)
| Schema::Fixed(_)
| Schema::Decimal(_)
| Schema::BigDecimal
| Schema::Date
| Schema::TimeMillis
| Schema::TimeMicros
| Schema::TimestampMillis
| Schema::TimestampMicros
| Schema::TimestampNanos
| Schema::LocalTimestampMillis
| Schema::LocalTimestampMicros
| Schema::LocalTimestampNanos
| Schema::Duration
| Schema::Ref { .. } => {
    unreachable!("SchemaKind::Uuid can only be a Schema::Uuid")
}
```

avro/src/schema.rs
Kriskras99: I've changed it to String as Bytes is Rust only

avro/src/schema_equality.rs
martin-g: Was I wrong here or you forgot to push before/after resolving the comment ?

martin-g: OK! You just pushed!

Kriskras99: I always start resolving comments as I fix them, as otherwise I loose track of which ones I already fixed and keep revisiting them. So it might take some time for push to appear after a resolve 😅

avro/src/serde/ser_schema.rs
```diff
@@ -1186,7 +1190,7 @@ impl<'s, W: Write> SchemaAwareWriteSerializer<'s, W> {
                     match variant_schema {
                         Schema::String
                         | Schema::Bytes
-                        | Schema::Uuid
+                        | Schema::Uuid(UuidSchema::Bytes)
```
martin-g: What about `Schema::Uuid(UuidSchema::Fixed(..))` ?

avro/src/schema.rs
```diff
@@ -1547,7 +1617,9 @@ impl Parser {
                                 Ok((precision, scale)) => Ok(Schema::Decimal(DecimalSchema {
                                     precision,
                                     scale,
-                                    inner: Box::new(inner),
+                                    inner: inner.try_into().unwrap_or_else(|_| {
+                                        unreachable!("inner is Fixed or Bytes")
```
martin-g: Do we really need to unwrap here ?
`inner.try_into()?` should be enough, no ? It will return an Err - `Err(Details::ResolveDecimalSchema(value.into()).into()),`

avro/src/schema.rs
```diff
@@ -6713,14 +6792,38 @@ mod tests {
...
+          "type": "bytes",
+          "name": "BytesUUID",
+          "logicalType": "uuid"
+        });
+        let parse_result = Schema::parse(&schema)?;
+        assert_eq!(parse_result, Schema::Uuid(UuidSchema::Bytes));
+
+        Ok(())
+    }
+
```
martin-g, line 6836: Do we have a similar test already:
```rust
#[test]
fn test_uuid_fixed_invalid_size() {
    let schema = Schema::parse_str(r#"{
        "type": "fixed",
        "size": 8,
        "logicalType": "uuid",
        "name": "BadUUID"
    }"#)?;
    
}
```
It should fail due to `size!=16`.

avro/src/schema.rs
```diff
@@ -111,7 +111,7 @@ pub enum Schema {
     /// The underlying type is serialized and deserialized as `Schema::Bytes`
     BigDecimal,
     /// A universally unique identifier, annotating a string.
```
martin-g: `annotating a string` needs to be updated that it supports also Bytes and Fixed now

avro/src/schema.rs
martin-g, line 622: I think we should add an arm for `Schema::Uuid(UuidSchema::Fixed(FixedSchema { name, ..}))` here.
And add a test that resolves a Schema::Ref for it

martin-g, line 534: I think we should add an arm for `Schema::Uuid(UuidSchema::Fixed(FixedSchema { name, ..}))` here.
And add a test that resolves a Schema::Ref for it

avro/src/writer.rs
```diff
@@ -990,41 +991,41 @@ mod tests {
...
+        };
+        let inner = InnerDecimalSchema::Fixed(fixed.clone());
         let value = vec![0u8; size];
         logical_type_test(
             r#"{"type": {"type": "fixed", "size": 30, "name": "decimal"}, "logicalType": "decimal", "precision": 20, "scale": 5}"#,
             &Schema::Decimal(DecimalSchema {
                 precision: 20,
                 scale: 5,
-                inner: Box::new(inner.clone()),
+                inner: inner.clone(),
```
martin-g: nit: no need to clone, I think

avro/src/types.rs
```diff
@@ -478,6 +480,15 @@ impl Value {
                     None
                 }
             }
+            (&Value::Fixed(n, _), &Schema::Uuid(UuidSchema::Fixed(_))) => {
```
```suggestion
            (&Value::Fixed(n, _), &Schema::Uuid(UuidSchema::Fixed(size, ..))) => {
```
Should we check that `size` is also `16` here ?

avro/src/schema_equality.rs
```diff
@@ -143,9 +142,23 @@ impl SchemataEq for StructFieldEq {
                 Schema::Decimal(DecimalSchema { precision: precision_one, scale: scale_one, inner: inner_one }),
                 Schema::Decimal(DecimalSchema { precision: precision_two, scale: scale_two, inner: inner_two })
             ) => {
-                precision_one == precision_two && scale_one == scale_two && self.compare(inner_one, inner_two)
+                precision_one == precision_two && scale_one == scale_two && match (inner_one, inner_two) {
+                    (InnerDecimalSchema::Bytes, InnerDecimalSchema::Bytes) => true,
+                    (InnerDecimalSchema::Fixed(FixedSchema { size: size_one, .. }), InnerDecimalSchema::Fixed(FixedSchema { size: size_two, ..})) => {
```
martin-g, line 147: Do we need to compare the names here too ?!

avro/src/schema.rs
martin-g: Same for Decimal schema backed by FixedSchema

martin-g: Same for Decimal schema backed by FixedSchema

Kriskras99: I will do the same for `Schema::Decimal(Fixed)` as that is also missing

avro/src/schema.rs
Kriskras99: According to https://github.com/apache/avro-rs/blob/c0fef927bc97ee124d56aaa38bcb43421cad03c8/avro/src/schema.rs#L6725-L6751
it should not error, but result in a `Schema::Fixed` and log a warning

avro/src/types.rs
Kriskras99: `Schema::Uuid(UuidSchema::Fixed(size != 16))` shouldn't be able to exist, but an extra check doesn't hurt

avro/src/schema_equality.rs
Kriskras99: That's already done on line 60 through `Schema::name()`

avro/tests/serde_human_readable_true.rs
```diff
@@ -18,8 +18,7 @@ fn avro_rs_53_uuid_with_fixed_true() -> TestResult {
...
                             "logicalType" : "uuid",
                             "name": "FixedUUID"
```
```suggestion
                            "name": "StringUUID"
```

avro/src/encode.rs
```diff
@@ -136,23 +144,30 @@ pub(crate) fn encode_internal<W: Write, S: Borrow<Schema>>(
...
                 let bytes = uuid.as_bytes();
-                encode_bytes(bytes, writer)
+                writer
+                    .write(bytes.as_slice())
+                    .map_err(|e| Details::WriteBytes(e).into())
             }
             _ => Err(Details::EncodeValueAsSchemaError {
                 value_kind: ValueKind::Uuid,
-                supported_schema: vec![SchemaKind::Uuid, SchemaKind::Fixed],
+                supported_schema: vec![SchemaKind::Uuid, SchemaKind::Fixed, SchemaKind::Bytes],
```
martin-g:
```suggestion
                supported_schema: vec![SchemaKind::Uuid, SchemaKind::Fixed, SchemaKind::Bytes, SchemaKind::String],
```

avro/src/types.rs
```diff
@@ -478,6 +480,15 @@ impl Value {
                     None
                 }
             }
+            (&Value::Fixed(n, _), Schema::Uuid(UuidSchema::Fixed(size, ..))) => {
+                if n != 16 || size.size != 16 {
+                    Some(format!(
+                        "The value's size ('{n}') must be exactly 16 to be a Uuid"
```
martin-g, line 491: `size.size` checks the the schema's size. The error message always says `value's size`

avro/src/schema_compatibility.rs
```diff
@@ -1070,11 +1180,10 @@ mod tests {
             (nested_optional_record(), nested_record()),
         ];
 
-        assert!(
-            compatible_schemas
-                .iter()
-                .all(|(reader, writer)| SchemaCompatibility::can_read(writer, reader).is_ok())
-        );
+        for (reader, writer) in compatible_schemas {
+            println!("{reader:?}, {writer:?}");
```
martin-g: debug leftover

PR #342, opened 2025-11-24
feat!: Rework schema compatibility

Kriskras99:
I think this is better overall, but I can understand if you would want to revert the `Compatibility::Partial` and keep exactly to old behavior.

This new implementation is (in my opinion) simpler than the previous iteration and easier to modify when needed.

I've introduced the concept of partial and full compatibility. Partial compatibility means that there incompatibilities but that depends on whats written, for example an enum `[A, B, C]` is partially compatible with the enum `[A, B]` but depends on what the writer has written.

Users wishing to be absolutely sure, can treat `Compatibility::Partial` as an error, others can use it to slowly upgrade from an old schema.

Depends on #339, should be rebased after that's merged

### DIFF
--- a/avro/src/schema_compatibility.rs
+++ b/avro/src/schema_compatibility.rs
@@ -241,15 +241,9 @@ impl Checker {
             // float promotes to double
             (Schema::Float, Schema::Float | Schema::Double) => Ok(Compatibility::Full),
             (Schema::Double, Schema::Double) => Ok(Compatibility::Full),
-            // bytes and strings are interchangeable
-            (
-                Schema::Bytes | Schema::String | Schema::BigDecimal
-                | Schema::Uuid(UuidSchema::String | UuidSchema::Bytes)
-                | Schema::Decimal(DecimalSchema { inner: InnerDecimalSchema::Bytes, .. }),
-                Schema::Bytes | Schema::String | Schema::BigDecimal
-                | Schema::Uuid(UuidSchema::String | UuidSchema::Bytes)
-                | Schema::Decimal(DecimalSchema { inner: InnerDecimalSchema::Bytes, .. }),
-            ) => Ok(Compatibility::Full),
+            // This check needs to be above the other Decimal checks, so that if both schemas are
+            // Decimal then the precision and scale are checked. The other Decimal checks below will
+            // thus only hit if only one of the schemas is a Decimal
             (
                 Schema::Decimal(DecimalSchema { precision: w_precision, scale: w_scale, .. }),
                 Schema::Decimal(DecimalSchema { precision: r_precision, scale: r_scale, .. }),
@@ -266,6 +260,15 @@ impl Checker {
                     })
                 }
             }
+            // bytes and strings are interchangeable
+            (
+                Schema::Bytes | Schema::String | Schema::BigDecimal
+                | Schema::Uuid(UuidSchema::String | UuidSchema::Bytes)
+                | Schema::Decimal(DecimalSchema { inner: InnerDecimalSchema::Bytes, .. }),
+                Schema::Bytes | Schema::String | Schema::BigDecimal
+                | Schema::Uuid(UuidSchema::String | UuidSchema::Bytes)
+                | Schema::Decimal(DecimalSchema { inner: InnerDecimalSchema::Bytes, .. }),
+            ) => Ok(Compatibility::Full),
             (Schema::Uuid(_), Schema::Uuid(_)) => Ok(Compatibility::Full),
             (
                 Schema::Fixed(w_fixed) | Schema::Uuid(UuidSchema::Fixed(w_fixed))
--- a/avro/src/types.rs
+++ b/avro/src/types.rs
@@ -451,7 +451,7 @@ impl Value {
             (Value::Bytes(bytes), &Schema::Uuid(UuidSchema::Bytes)) => {
                 if bytes.len() != 16 {
                     Some(format!(
-                        "The value's size ({}) is not the right length for a fixed UUID (16)",
+                        "The value's size ({}) is not the right length for a bytes UUID (16)",
                         bytes.len()
                     ))
                 } else {
--- a/avro/tests/shared.rs
+++ b/avro/tests/shared.rs
@@ -17,6 +17,7 @@
 
 use apache_avro::{Codec, Reader, Schema, Writer, types::Value};
 use apache_avro_test_helper::TestResult;
+use std::path::PathBuf;
 use std::{
     fmt,
     fs::{DirEntry, File, ReadDir},
@@ -46,9 +47,9 @@ fn test_schema() -> TestResult {
         if let Ok(ft) = entry.file_type()
             && ft.is_dir()
         {
-            let sub_folder = ROOT_DIRECTORY.to_owned() + "/" + entry.file_name().to_str().unwrap();
+            let sub_folder = PathBuf::from(ROOT_DIRECTORY).join(entry.file_name());
 
-            let dir_result = test_folder(sub_folder.as_str());
+            let dir_result = test_folder(&sub_folder);
             if let Err(ed) = dir_result {
                 result = match result {
                     Ok(()) => Err(ed),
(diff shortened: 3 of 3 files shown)


avro/src/error.rs
```diff
@@ -602,15 +602,28 @@ pub enum CompatibilityError {
...
     MissingSymbols,
 
     #[error("Incompatible schemata! All elements in union must match for both schemas")]
     MissingUnionElements,
 
-    #[error("Incompatible schemata! Name and size don't match for fixed")]
+    #[error("Incompatible schemata! At least one element in the union must match the schema")]
+    SchemaMismatchAllUnionElements,
+
+    #[error("Incompatible schemata! Size don't match for fixed")]
```
```suggestion
    #[error("Incompatible schemata! Size doesn't match for fixed")]
```

Kriskras99:
I unfortunately couldn't get a fuzzer to work, the recursive branchy code makes it hard for the fuzzer to select new paths.

I took a look at `Compatibility` again, and I think it should stay. Here's a comparison of behaviour:

| Writer | Reader | Old | New |
| ------ | ------ | --- | --- |
| `Union<Null, Int>` | `Null` | Ok | Partial |
| `Union<Null, Int>` | `Int` | Ok | Partial |
| `Null` | `Union<Null, Int>` | Ok | Partial |
| `Int` | `Union<Null, Int>` | Ok | Partial |
| `Union<Null, String>` | `Union<Null, Int>` | Err | Partial |
| `Int` | `TimeMicros` | Err | Full |
| `Enum<A, B, C>` | `Enum<A, B, C, D>` | Ok | Full |
| `Enum<A, B, C>` | `Enum<A, B>` | Err | Partial |

So the new behaviour allows slightly more, but is also more honest about the compatibility.
A user who wants absolutely no schema evolution errors, can check for `Ok(Compatibility::Full)`, while a user who can allow that can just check `is_ok`.

What do you think based on this table?

avro/src/error.rs
```diff
@@ -605,15 +605,28 @@ pub enum CompatibilityError {
     #[error("Incompatible schemata! Field '{0}' in reader schema must have a default value")]
     MissingDefaultValue(String),
 
-    #[error("Incompatible schemata! Reader's symbols must contain all writer's symbols")]
+    #[error("Incompatible schemata! Reader's symbols contains none of the writer's symbols")]
```
```suggestion
    #[error("Incompatible schemata! Reader's symbols contain none of the writer's symbols")]
```

avro/src/schema_compatibility.rs
```diff
@@ -18,268 +18,363 @@
...
+            (Schema::Duration, Schema::Duration) => Ok(Compatibility::Full),
+            (Schema::Fixed(FixedSchema { size: 12, .. }), Schema::Duration) => {
+                Ok(Compatibility::Full)
+            }
+            (Schema::Duration, Schema::Fixed(FixedSchema { size: 12, .. })) => {
+                Ok(Compatibility::Full)
+            }
+            (
+                Schema::Decimal(DecimalSchema { precision: w_precision, scale: w_scale, .. }),
+                Schema::Decimal(DecimalSchema { precision: r_precision, scale: r_scale, .. }),
```
martin-g, line 249: Shouldn't this also check the `inner` field ?

avro/tests/serde_human_readable_true.rs
```diff
@@ -18,8 +18,7 @@ fn avro_rs_53_uuid_with_fixed_true() -> TestResult {
                         "fields" : [ {
                           "name" : "id",
                           "type" : {
-                            "type" : "fixed",
-                            "size" : 16,
+                            "type" : "string",
                             "logicalType" : "uuid",
                             "name": "FixedUUID"
```
```suggestion
                            "name": "StringUUID"
```

avro/src/schema_compatibility.rs
```diff
@@ -1000,12 +803,24 @@ mod tests {
...
             expected_error,
-            SchemaCompatibility::match_schemas(&writer_schema, &reader_schema).unwrap_err()
+            SchemaCompatibility::can_read(&writer_schema, &reader_schema).unwrap_err()
         )
     }
 
     #[test]
     fn test_compatible_reader_writer_pairs() {
+        let uuid_fixed = FixedSchema {
+            name: Name {
```
martin-g:
```suggestion
            name: Name::new("uuid-fixed")
```

avro/src/types.rs
```diff
@@ -441,14 +442,15 @@ impl Value {
...
             (&Value::BigDecimal(_), &Schema::BigDecimal) => None,
             (&Value::Duration(_), &Schema::Duration) => None,
-            (&Value::Uuid(_), &Schema::Uuid) => None,
+            (&Value::Uuid(_), &Schema::Uuid(_)) => None,
             (&Value::Float(_), &Schema::Float) => None,
             (&Value::Float(_), &Schema::Double) => None,
             (&Value::Double(_), &Schema::Double) => None,
             (&Value::Bytes(_), &Schema::Bytes) => None,
             (&Value::Bytes(_), &Schema::Decimal { .. }) => None,
+            (&Value::Bytes(_), &Schema::Uuid(UuidSchema::Bytes)) => None,
```
martin-g: Return None only if `bytes` (in Value::Bytes(bytes) length is 16, otherwise - Some

avro/src/schema_equality.rs
```diff
@@ -544,4 +556,60 @@ mod tests {
...
+            doc: None,
+            size: 16,
+            default: None,
+            attributes: Default::default(),
+        };
+        let fixed = Schema::Fixed(fixed_schema.clone());
+        fixed_schema
+            .attributes
+            .insert("Something".to_string(), Value::Null);
+        let fixed_different = Schema::Fixed(fixed_schema);
```
```suggestion
        let fixed_different = Schema::Uuid(UuidSchema::Fixed(fixed_schema));
```

avro/src/schema_equality.rs
```diff
@@ -544,4 +556,60 @@ mod tests {
...
+                name: String::new(),
+                namespace: None,
+            },
+            aliases: None,
+            doc: None,
+            size: 16,
+            default: None,
+            attributes: Default::default(),
+        };
+        let fixed = Schema::Fixed(fixed_schema.clone());
```
```suggestion
        let fixed = Schema::Uuid(UuidSchema::Fixed(fixed_schema.clone()));
```

avro/src/schema_compatibility.rs
Kriskras99: No, we can resolve two Decimals with the same scale and precision, doesn't matter how they're stored in the file

martin-g: https://github.com/apache/avro-rs/pull/342/files#diff-69e21fcef711fee94f82bf01f95510fb48057f5e01841c500dac5f6494637c72R145-R148 makes such check ...

avro/src/types.rs
Kriskras99: I've also done the same for Value::String with a minimum length of 32

avro/src/schema_compatibility.rs
Kriskras99: Looking at the implementation of [`Value::resolve` for `Decimal`](https://github.com/apache/avro-rs/blob/13625abd7d08caa2eff8535768d5547b21f4bdff/avro/src/types.rs#L746), the only thing that matters is that the underlying type can store enough bytes, so the check could actually be slightly relaxed to not require the exact same precision and scale (but lets not do that for now).

Kriskras99: I've added a test for this, and it does work

avro/src/types.rs
```diff
@@ -448,9 +448,28 @@ impl Value {
             (&Value::Double(_), &Schema::Double) => None,
             (&Value::Bytes(_), &Schema::Bytes) => None,
             (&Value::Bytes(_), &Schema::Decimal { .. }) => None,
-            (&Value::Bytes(_), &Schema::Uuid(UuidSchema::Bytes)) => None,
+            (Value::Bytes(bytes), &Schema::Uuid(UuidSchema::Bytes)) => {
+                if bytes.len() != 16 {
+                    Some(format!(
+                        "The value's size ({}) is not the right length for a fixed UUID (16)",
```
```suggestion
                        "The value's size ({}) is not the right length for a bytes UUID (16)",
```

avro/tests/shared.rs
```diff
@@ -43,17 +43,16 @@ fn test_schema() -> TestResult {
...
-                let dir_result = test_folder(sub_folder.as_str());
-                if let Err(ed) = dir_result {
-                    result = match result {
-                        Ok(()) => Err(ed),
-                        Err(e) => Err(e.merge(&ed)),
-                    }
+        if let Ok(ft) = entry.file_type()
+            && ft.is_dir()
+        {
+            let sub_folder = ROOT_DIRECTORY.to_owned() + "/" + entry.file_name().to_str().unwrap();
```
martin-g: Not caused by this PR but this path concatenation here is not very Windows-friendly.

avro/src/schema_compatibility.rs
```diff
@@ -16,271 +16,359 @@
...
+                | Schema::LocalTimestampMicros | Schema::LocalTimestampNanos,
+                Schema::Long | Schema::Float | Schema::Double | Schema::TimeMicros
+                | Schema::TimestampMillis | Schema::TimestampMicros | Schema::TimestampNanos
+                | Schema::LocalTimestampMillis | Schema::LocalTimestampMicros
+                | Schema::LocalTimestampNanos,
+            ) => Ok(Compatibility::Full),
+            // float promotes to double
+            (Schema::Float, Schema::Float | Schema::Double) => Ok(Compatibility::Full),
+            (Schema::Double, Schema::Double) => Ok(Compatibility::Full),
+            // bytes and strings are interchangeable
```
martin-g: Are the always interchangeable ?
UuidSchema::Bytes has to be exactly 16 bytes.
UuidString::String has to be at least 32 chars.
The checks for Decimal bytes schema also miss to check the precision and scale here.

Kriskras99: Because they are decoded to and encoded from the `uuid::Uuid` type, resolving them will always work.
I've moved the Decimal check with the precision and scale above this statement, so this will only match now if one is `Decimal::Bytes` and one isn't.

martin-g (approved):

PR #351, opened 2025-12-04
fix: Don't depend on Serde to provide fields in the right order

Kriskras99:
This is a (failing) test for when the field order differs between the struct (Serde) and the schema:
```
  assertion `left == right` failed
    left: Foo { a: "World", b: "Hello" }
   right: Foo { a: "Hello", b: "World" }
in avro/src/ser_schema.rs, line 2940
thread: ser_schema::tests::different_field_order_serde_vs_schema
```

There are two options:

- Require that the field order is the same
- Modify `ser_schema.rs` (and potentially `de.rs`) to care about the field order

### DIFF
--- a/avro/src/ser_schema.rs
+++ b/avro/src/ser_schema.rs
@@ -26,7 +26,7 @@ use crate::{
 };
 use bigdecimal::BigDecimal;
 use serde::ser;
-use std::{borrow::Cow, collections::HashMap, io::Write, str::FromStr};
+use std::{borrow::Cow, cmp::Ordering, collections::HashMap, io::Write, str::FromStr};
 
 const COLLECTION_SERIALIZER_ITEM_LIMIT: usize = 1024;
 const COLLECTION_SERIALIZER_DEFAULT_INIT_ITEM_CAPACITY: usize = 32;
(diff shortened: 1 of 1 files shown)


Kriskras99:
Fixed, CI will still fail as the test does fail and we need to decide what the right course of action is.

martin-g:
> we need to decide what the right course of action is

I won't have time to debug it in the next days but I was going to bet that we probably don't use the `lookup` map but we do - https://github.com/apache/avro-rs/blob/335e74edcfaf204e504d412bba0b52129603bc20/avro/src/ser_schema.rs#L298-L299
So, it must be something else.

Kriskras99:
I've added support for out-of-order fields in the serializer. In the deserializer no changed are necessary.
This changes test introduced in #227 as an `#[avro(default = ...)]` is now required when using any of the Serde `skip` attributes. I'm also working on a PR to check the Serde attributes in avro-derive (and add documentation for the attributes avro-derive supports as they don't seem to be documented anywhere).

avro/src/ser_schema.rs
```diff
@@ -249,6 +249,9 @@ impl<W: Write> ser::SerializeMap for SchemaAwareWriteSerializeMap<'_, '_, W> {
 pub struct SchemaAwareWriteSerializeStruct<'a, 's, W: Write> {
     ser: &'a mut SchemaAwareWriteSerializer<'s, W>,
     record_schema: &'s RecordSchema,
+    /// Fields we received in the wrong order
+    field_cache: Vec<(usize, Vec<u8>)>,
```
martin-g: Why a Vec ? Wouldn't it be simpler with a `HashMap<usize, Vec<u8>>` ?

avro/src/ser_schema.rs
```diff
@@ -249,6 +249,9 @@ impl<W: Write> ser::SerializeMap for SchemaAwareWriteSerializeMap<'_, '_, W> {
...
+    next_field: usize,
```
martin-g:
```suggestion
    field_position: usize,
```
I think this will be a better self-documenting name

avro/src/ser_schema.rs
```diff
@@ -268,19 +273,86 @@ impl<'a, 's, W: Write> SchemaAwareWriteSerializeStruct<'a, 's, W> {
...
+        }
+
+        // Check if all fields were written
+        if self.next_field != self.record_schema.fields.len() {
+            let name = self.record_schema.fields[self.next_field].name.clone();
+            return Err(Details::GetField(name).into());
+        }
+        assert!(
+            self.field_cache.is_empty(),
+            "There should be no more unwritten fields at this point"
```
martin-g: Let's print the contents of `self.field_cache` for easier debugging.

avro/src/ser_schema.rs
```diff
@@ -2900,4 +2978,50 @@ mod tests {
         string_record.serialize(&mut serializer)?;
         Ok(())
     }
+
+    #[test]
+    fn different_field_order_serde_vs_schema() -> TestResult {
```
martin-g:
```suggestion
    fn avro_rs_351_different_field_order_serde_vs_schema() -> TestResult {
```

avro/src/ser_schema.rs
```diff
@@ -2900,4 +2978,50 @@ mod tests {
...
+        #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+        struct Foo {
+            a: String,
+            b: String,
```
martin-g, line 2986: Let's make a struct with at least 5 fields. Just to make it a bit more real.

avro/src/ser_schema.rs
```diff
@@ -2900,4 +2978,50 @@ mod tests {
...
+        }
+        "#,
+        )?;
+
+        let mut writer = Writer::new(&schema, Vec::new())?;
+        if let Err(e) = writer.append_ser(Foo {
+            a: "Hello".into(),
+            b: "World".into(),
+        }) {
+            panic!("{e:?}");
```
martin-g: You can use 'append_ser(...)?'. The method returns `TestResult`

avro/tests/avro-rs-226.rs
```diff
@@ -128,3 +133,69 @@ fn avro_rs_226_index_out_of_bounds_with_serde_skip_multiple_fields() -> TestResu
...
+
+    ser_deser::<T>(
+        &T::get_schema(),
+        T {
+            x: Some(0),
+            y: None,
+            z: None,
+        },
+    )
+    .unwrap()
```
martin-g, line 200: Let's use `TestResult` return type and avoid using `.unwrap()`.

avro/tests/avro-rs-226.rs
```diff
@@ -128,3 +133,69 @@ fn avro_rs_226_index_out_of_bounds_with_serde_skip_multiple_fields() -> TestResu
...
+
+    ser_deser::<T>(
+        &T::get_schema(),
+        T {
+            x: None,
+            y: None,
+            z: Some(1),
+        },
+    )
+    .unwrap()
```
martin-g, line 178: Let's use `TestResult` return type and avoid using `.unwrap()`.

avro/tests/avro-rs-226.rs
```diff
@@ -128,3 +133,69 @@ fn avro_rs_226_index_out_of_bounds_with_serde_skip_multiple_fields() -> TestResu
...
```
martin-g, line 156: Let's use `TestResult` return type and avoid using `.unwrap()`.

Kriskras99: That doesn't work. `ser_deser` is supposed to error and because creating a TestResult is the same as panicking I need to check the panic message.
If I add a `-> TestResult` and return `Ok(())` clippy complains that my should_panic test should not have a return type.

avro/src/ser_schema.rs
```diff
@@ -268,19 +273,79 @@ impl<'a, 's, W: Write> SchemaAwareWriteSerializeStruct<'a, 's, W> {
...
+                return Err(Details::MissingDefaultForSkippedField {
+                    field_name: field_info.name.clone(),
+                    schema: Schema::Record(self.record_schema.clone()),
+                }
+                .into());
+            }
+        }
+
+        // Check if all fields were written
+        if self.field_position != self.record_schema.fields.len() {
```
martin-g: Is this check really needed ?
The `while` above makes sure of it or returns an Err.

avro/src/ser_schema.rs
```diff
@@ -268,19 +273,79 @@ impl<'a, 's, W: Write> SchemaAwareWriteSerializeStruct<'a, 's, W> {
...
+            let name = self.record_schema.fields[self.field_position].name.clone();
+            return Err(Details::GetField(name).into());
+        }
+        assert!(
```
martin-g: I am not sure about that.
A library should try to avoid panicking if possible.
I think `debug_assert!()` or returning an Err is better.

avro/tests/avro-rs-226.rs
martin-g: I see.

PR #359, opened 2025-12-07
feat: Implement support for `#[serde(flatten)]`

### LINKED ISSUE #247: missing support for #[serde(flatten)] for derive AvroSchema

Kriskras99:
This is done by adding a `#[avro(flatten]` attribute so that the schema (for that field) is also flattened, and by adding support in `SchemaAwareWriteSerializer` for serializing a struct via Map instead of Struct.

`flatten` does not work with `to_value`, as `to_value` does not have access to the schema.

I moved the Serde related modules to a separate `serde` module as I needed to add another module for the `StringSerializer` (to decode Map keys) and that would make the root too messy. Putting it in the root `util` module would also work, but it's a lot of (filler) code and would make that module unreadable.

Closes #247
Closes #311

### DIFF
--- a/avro/src/serde/ser_schema.rs
+++ b/avro/src/serde/ser_schema.rs
@@ -356,9 +356,10 @@ impl<'a, 's, W: Write> SchemaAwareWriteSerializeStruct<'a, 's, W> {
             "There should be no more unwritten fields at this point: {:?}",
             self.field_cache
         );
-        assert!(
+        debug_assert!(
             self.map_field_name.is_none(),
-            "There should be no field name at this point"
+            "There should be no field name at this point: field {:?}",
+            self.map_field_name
         );
         Ok(self.bytes_written)
     }
@@ -425,6 +426,7 @@ impl<W: Write> ser::SerializeStruct for SchemaAwareWriteSerializeStruct<'_, '_, 
     }
 }
 
+/// This implementation is used to support `#[serde(flatten)]` as that uses SerializeMap instead of SerializeStruct.
 impl<W: Write> ser::SerializeMap for SchemaAwareWriteSerializeStruct<'_, '_, W> {
     type Ok = usize;
     type Error = Error;
@@ -434,9 +436,11 @@ impl<W: Write> ser::SerializeMap for SchemaAwareWriteSerializeStruct<'_, '_, W> 
         T: ?Sized + Serialize,
     {
         let name = key.serialize(StringSerializer)?;
-        assert!(
-            self.map_field_name.replace(name).is_none(),
-            "Got two keys in a row"
+        let old = self.map_field_name.replace(name);
+        debug_assert!(
+            old.is_none(),
+            "Expected a value instead of a key: old key: {old:?}, new key: {:?}",
+            self.map_field_name
         );
         Ok(())
     }
@@ -445,7 +449,7 @@ impl<W: Write> ser::SerializeMap for SchemaAwareWriteSerializeStruct<'_, '_, W> 
     where
         T: ?Sized + Serialize,
     {
-        let key = self.map_field_name.take().expect("Got value without key");
+        let key = self.map_field_name.take().ok_or(Details::MapNoKey)?;
         let record_field = self
             .record_schema
             .lookup
@@ -1605,10 +1609,14 @@ impl<'s, W: Write> SchemaAwareWriteSerializer<'s, W> {
             Schema::Map(map_schema) => Ok(SchemaAwareWriteSerializeMapOrStruct::Map(
                 SchemaAwareWriteSerializeMap::new(self, map_schema.types.as_ref(), len),
             )),
+            Schema::Ref { name: ref_name } => {
+                let ref_schema = self.get_ref_schema(ref_name)?;
+                self.serialize_map_with_schema(len, ref_schema)
+            }
             Schema::Union(union_schema) => {
                 for (i, variant_schema) in union_schema.schemas.iter().enumerate() {
                     match variant_schema {
-                        Schema::Map(_) => {
+                        Schema::Map(_) | Schema::Record(_) | Schema::Ref { .. } => {
                             encode_int(i as i32, &mut *self.writer)?;
                             return self.serialize_map_with_schema(len, variant_schema);
                         }
--- a/avro/src/serde/util.rs
+++ b/avro/src/serde/util.rs
@@ -266,6 +266,7 @@ impl SerializeMap for StringSerializer {
         Err(Details::MapFieldExpectedString.into())
     }
 }
+
 impl SerializeStruct for StringSerializer {
     type Ok = String;
     type Error = Error;
@@ -281,6 +282,7 @@ impl SerializeStruct for StringSerializer {
         Err(Details::MapFieldExpectedString.into())
     }
 }
+
 impl SerializeStructVariant for StringSerializer {
     type Ok = String;
     type Error = Error;
--- a/avro_derive/src/lib.rs
+++ b/avro_derive/src/lib.rs
@@ -179,7 +179,7 @@ fn get_data_struct_schema_def(
                                 schema_fields.push(field)
                             }
                         } else {
-                            panic!("Can only flatten RecordSchema")
+                            panic!("Can only flatten RecordSchema, got {:?}", #flatten_ty::get_schema())
                         }
                     });
 
@@ -238,6 +238,8 @@ fn get_data_struct_schema_def(
     Ok(quote! {
         let mut schema_fields = Vec::with_capacity(#minimum_fields);
         #(#record_field_exprs)*
+        let schema_field_set: ::std::collections::HashSet<_> = schema_fields.iter().map(|rf| &rf.name).collect();
+        assert_eq!(schema_fields.len(), schema_field_set.len(), "Duplicate field names found: {schema_fields:?}");
         let name = apache_avro::schema::Name::new(#full_schema_name).expect(&format!("Unable to parse struct name for schema {}", #full_schema_name)[..]);
         let lookup: std::collections::BTreeMap<String, usize> = schema_fields
             .iter()
--- a/avro_derive/tests/derive.rs
+++ b/avro_derive/tests/derive.rs
@@ -1688,7 +1688,7 @@ mod test_derive {
     }
 
     #[test]
-    fn avro_247_serde_flatten_support() {
+    fn avro_rs_247_serde_flatten_support() {
         #[derive(Debug, Serialize, Deserialize, AvroSchema, Clone, PartialEq)]
         struct Nested {
             a: bool,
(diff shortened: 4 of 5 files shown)


avro/src/serde/ser_schema.rs
```diff
@@ -352,6 +356,10 @@ impl<'a, 's, W: Write> SchemaAwareWriteSerializeStruct<'a, 's, W> {
             "There should be no more unwritten fields at this point: {:?}",
             self.field_cache
         );
+        assert!(
```
```suggestion
        debug_assert!(
```

avro/src/serde/ser_schema.rs
```diff
@@ -352,6 +356,10 @@ impl<'a, 's, W: Write> SchemaAwareWriteSerializeStruct<'a, 's, W> {
...
+            self.map_field_name.is_none(),
+            "There should be no field name at this point"
```
```suggestion
            "There should be no field name at this point: {}", self.map_field_name
```

avro/src/serde/ser_schema.rs
```diff
@@ -420,6 +425,50 @@ impl<W: Write> ser::SerializeStruct for SchemaAwareWriteSerializeStruct<'_, '_,
...
+impl<W: Write> ser::SerializeMap for SchemaAwareWriteSerializeStruct<'_, '_, W> {
+    type Ok = usize;
+    type Error = Error;
+
+    fn serialize_key<T>(&mut self, key: &T) -> Result<(), Self::Error>
+    where
+        T: ?Sized + Serialize,
+    {
+        let name = key.serialize(StringSerializer)?;
+        assert!(
```
```suggestion
        debug_assert!(
```

avro/src/serde/ser_schema.rs
```diff
@@ -420,6 +425,50 @@ impl<W: Write> ser::SerializeStruct for SchemaAwareWriteSerializeStruct<'_, '_,
...
+            self.map_field_name.replace(name).is_none(),
+            "Got two keys in a row"
```
martin-g: The error message is not very clear to me.
Let's extract the result of `Option::replace()` to a local variable and print it in case of an error.

avro/src/serde/ser_schema.rs
```diff
@@ -420,6 +425,50 @@ impl<W: Write> ser::SerializeStruct for SchemaAwareWriteSerializeStruct<'_, '_,
...
+        );
+        Ok(())
+    }
+
+    fn serialize_value<T>(&mut self, value: &T) -> Result<(), Self::Error>
+    where
+        T: ?Sized + Serialize,
+    {
+        let key = self.map_field_name.take().expect("Got value without key");
```
martin-g: Let's return an Avro error instead of crashing.

avro/src/serde/ser_schema.rs
```diff
@@ -420,6 +425,50 @@ impl<W: Write> ser::SerializeStruct for SchemaAwareWriteSerializeStruct<'_, '_,
     }
 }
 
+impl<W: Write> ser::SerializeMap for SchemaAwareWriteSerializeStruct<'_, '_, W> {
```
martin-g, line 430: Please add a comment/docstring explaining when this impl is being used, i.e. for the flattening support.

avro_derive/src/lib.rs
```diff
@@ -142,26 +144,46 @@ fn get_data_struct_schema_def(
...
+                    // the schema here.
+                    let flatten_ty = &field.ty;
+                    record_field_exprs.push(quote! {
+                        if let ::apache_avro::schema::Schema::Record(::apache_avro::schema::RecordSchema { fields, .. }) = #flatten_ty::get_schema() {
+                            for mut field in fields {
+                                field.position = schema_fields.len();
+                                schema_fields.push(field)
+                            }
+                        } else {
+                            panic!("Can only flatten RecordSchema")
```
martin-g: Let's print the non-RecordSchema for easier debugging.

avro_derive/tests/derive.rs
```diff
@@ -1686,4 +1686,103 @@ mod test_derive {
             panic!("Unexpected schema type for Foo")
         }
     }
+
+    #[test]
+    fn avro_247_serde_flatten_support() {
```
```suggestion
    fn avro_rs_247_serde_flatten_support() {
```
The tests without `_rs` are pointing to the old JIRA tickets.

avro_derive/tests/derive.rs
```diff
@@ -1686,4 +1686,103 @@ mod test_derive {
...
+        assert_eq!(schema, derived_schema);
+
+        serde_assert(Foo {
+            nested: Nested { a: true },
+            b: 321,
+        });
+    }
+
+    #[test]
+    fn avro_247_serde_nested_flatten_support() {
```
```suggestion
    fn avro_rs_247_serde_nested_flatten_support() {
```

avro_derive/tests/derive.rs
```diff
@@ -1686,4 +1686,103 @@ mod test_derive {
...
+        let bar = Bar {
+            foo: Foo {
+                nested_foo: NestedFoo { one: 42 },
+            },
+            two: 2,
+            // test_enum: TestEnum::B,
+        };
+
+        let schema = Bar::get_schema();
+        println!("Generated schema: {:#?}", schema);
```
martin-g: debug leftover

avro_derive/tests/derive.rs
martin-g: Any reason against returning `TestResult` and get rid of the `.unwrap()`s below ?

avro_derive/tests/derive.rs
martin-g: Any reason against returning `TestResult` and get rid of the `.unwrap()`s below ?

avro_derive/tests/derive.rs
```diff
@@ -1686,4 +1686,103 @@ mod test_derive {
...
+
+        let bar = Bar {
+            foo: Foo {
+                nested_foo: NestedFoo { one: 42 },
+            },
+            two: 2,
+            // test_enum: TestEnum::B,
+        };
+
+        let schema = Bar::get_schema();
```
martin-g: IMO it would be nice to assert the generated schema here too.

avro_derive/src/lib.rs
```diff
@@ -39,6 +39,8 @@ struct FieldOptions {
     rename: Option<String>,
     #[darling(default)]
     skip: Option<bool>,
+    #[darling(default)]
+    flatten: Option<bool>,
```
martin-g, line 43: IMO we could define a 
```rust
#[derive(darling::FromAttributes)]
#[darling(attributes(serde))]
struct SerdeOptions {
...
}
```
and move all serde-related options here.
This way there won't be a need the user application to duplicate all of them in the `avro(...)` namespace.
This will be a breaking change but I think the users will appreciate it in the long term. At the moment it is possible to have a field with only serde's or only avro's flatten attribute and this will lead to problems.
WDYT ?

This could be done in a follow-up PR!

avro_derive/src/lib.rs
martin-g: Return a syn::Error instead of panicking ?!

avro_derive/tests/derive.rs
```diff
@@ -1686,4 +1686,103 @@ mod test_derive {
...
+        let mut writer = Writer::new(&schema, Vec::new()).unwrap();
+        writer.append_ser(&bar).unwrap();
+
+        // Check that it was correctly serialized and is deserializable
+        let encoded = writer.into_inner().unwrap();
+        let mut reader = Reader::new(&encoded[..]).unwrap();
+        let value = reader.next().unwrap().unwrap();
+        let result: Bar = from_value(&value).unwrap();
+        assert_eq!(result, bar);
+    }
```
martin-g, line 1852: Ideas for more tests:
* conflicting names in the outer struct and in the Nested/flatten one
* `skip` attribute on a flattened field
* empty flattened struct

avro/src/serde/util.rs
```diff
@@ -0,0 +1,298 @@
...
+    where
+        T: ?Sized + Serialize,
+    {
+        Err(Details::MapFieldExpectedString.into())
+    }
+
+    fn end(self) -> Result<Self::Ok, Self::Error> {
+        Err(Details::MapFieldExpectedString.into())
+    }
+}
```
martin-g, line 284:
```suggestion
}

```

avro/src/serde/ser_schema.rs
```diff
@@ -1532,6 +1619,9 @@ impl<'s, W: Write> SchemaAwareWriteSerializer<'s, W> {
                     "Expected a Map schema in {union_schema:?}"
                 )))
             }
+            Schema::Record(record_schema) => Ok(SchemaAwareWriteSerializeMapOrStruct::Struct(
+                SchemaAwareWriteSerializeStruct::new(self, record_schema),
+            )),
```
martin-g, line 1632: Do we need a branch for `Schema::Ref` too ?

avro/src/serde/ser_schema.rs
martin-g, line 1622: Add support for Schema::Record and maybe for Schema::Ref too

avro_derive/src/lib.rs
Kriskras99: This panic is happening at runtime (when the user/library calls `get_schema`). The function signature is `AvroSchema::get_schema() -> Schema` so we can't return an error either.
We could add a test that just calls `get_schema()`, then the user will see it when they run `cargo test`. I've added it to #361

avro_derive/src/lib.rs
Kriskras99: I think that is a good idea, but a bit more complicated for other attributes (especially `skip_serializing` and `default`). We can add this later and also keep supporting (or at least ignoring) `avro(flatten)`
I've added it to #361.

avro_derive/tests/derive.rs
Kriskras99: All tests in `avro_derive` use `unwrap`, `avro_test_helper` is not a dependency of `avro_derive` so `TestResult` does not exist.

avro_derive/tests/derive.rs
Kriskras99: I've added tests for the first two. The last one is not possible, as it gives compile errors, probems with type inference in the AvroSchema derive

Kriskras99: I also made the tests a lot better (like actually using serde_assert for the second test)

PR #362, opened 2025-12-09
test: uncomment attribute test

Kriskras99:
While looking for leftover TODO's in my own code I noticed this test.
I had too slightly modify it, as we don't store custom attributes for basic types.

I also deleted a commented test which was not relevant anymore.

### DIFF
--- a/avro/tests/schema.rs
+++ b/avro/tests/schema.rs
@@ -30,7 +30,7 @@ use apache_avro::{
 };
 use apache_avro_test_helper::{
     TestResult,
-    data::{DOC_EXAMPLES, examples, valid_examples},
+    data::{DOC_EXAMPLES, OTHER_ATTRIBUTES_EXAMPLES, examples, valid_examples},
     init,
 };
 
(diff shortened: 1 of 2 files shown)


avro/tests/schema.rs
```diff
@@ -708,39 +708,41 @@ fn test_doc_attributes() -> TestResult {
...
-               // all fields in the record have at least some user-defined attributes
-               assert!(f.schema.other_attributes.is_some());
-               for prop in f.schema.other_attributes?.iter() {
-                   assert_attribute_type(prop);
-               }
-           }
+        if let Schema::Record(RecordSchema { fields, .. }) = schema {
+            for f in fields {
+                // all fields in the record have at least some user-defined attributes
+                assert!(f.schema.custom_attributes().is_some());
```
martin-g: Here you use the schema's attributes, not the field's attributes.
Those are already verified at https://github.com/apache/avro-rs/pull/362/files#diff-4f38d574bfee56e26d21afb16dd9f0fc3d2554436da429f1e4facea752e15ffdR730-R732

Kriskras99: Fixed!

martin-g (approved):

PR #369, opened 2025-12-18
feat: Configure Clippy in the workspace instead of the CI

Kriskras99:
This way developers can run Clippy locally and get the same errors as on the CI. They've been configured as warnings, to ease development, and will only be errors in the CI.

### DIFF
--- a/.github/workflows/test-lang-rust-clippy.yml
+++ b/.github/workflows/test-lang-rust-clippy.yml
@@ -47,4 +47,4 @@ jobs:
         with:
           toolchain: ${{ matrix.rust }}
           components: clippy
-      - run: cargo clippy --all-features --all-targets -- -Dwarnings
+      - run: cargo clippy --all-features --all-targets


.github/workflows/test-lang-rust-clippy.yml
```diff
@@ -47,4 +47,4 @@ jobs:
         with:
           toolchain: ${{ matrix.rust }}
           components: clippy
-      - run: cargo clippy --all-features --all-targets -- -Dclippy::all -Dunused_imports
+      - run: cargo clippy --all-features --all-targets -- -Dwarnings
```
martin-g: Isn't `-Dwarnings` already enabled by https://github.com/apache/avro-rs/pull/369/changes#diff-0cd900b9eb208b7d935ea9244ef166db0ab25677400847e279dd58ad7c1aec83R30 ?

Cargo.toml
```diff
@@ -49,3 +49,7 @@ pretty_assertions = { default-features = false, version = "1.4.1", features = ["
 [profile.release.package.hello-wasm]
 # Tell `rustc` to optimize for small code size.
 opt-level = "s"
+
+[workspace.lints]
+# By setting the priority to -1, this lint group can be overridden for specific lints
+clippy.all = { level = "warn", priority = -1 }
```
martin-g, line 55: What happened with `-Dunused_imports` ?!
I see no entry for it at https://rust-lang.github.io/rust-clippy/master/index.html but Google returns results for such lint ...

.github/workflows/test-lang-rust-clippy.yml
Kriskras99: You're right

Cargo.toml
Kriskras99: That's a rustc lint. And it's warn by default, so -Dwarnings will also make it an error

martin-g (approved):

Issue #378, opened 2025-12-28
Name and Namespace fields for Schema::Duration


jdarais:
According to the avro spec, Duration [is a fixed type](https://avro.apache.org/docs/1.12.0/specification/#duration), and fixed types are [required to have a name](https://avro.apache.org/docs/1.12.0/specification/#fixed).

Should `Schema::Duration` contain a `FixedSchema` like other logical types that can use  fixed as the underlying type like `Schema::Decimal` and `Schema::Uuid`?  It can almost do without, since the fixed type's size is required to be 12 for the duration logical type, but it seems we still need a place to store name and namespace, otherwise that information will be lost at parsing time.

I'm happy to submit a PR for this, just wanted to check in case there's a reason Duration doesn't contain a FixedSchema already.

martin-g:
> I'm happy to submit a PR for this, just wanted to check in case there's a reason Duration doesn't contain a FixedSchema already.

No reason!
We just merged https://github.com/apache/avro-rs/pull/339 with [improvements](https://github.com/apache/avro-rs/pull/339/changes#diff-e9691d61212ac54252a9f4b8ffd76c9e527949019e82d66e5d176cc801f7d1faR490-R589) to DecimalSchema and UuidSchema related to this. We should do the same for Schema::Duration too!

We should also add support for all of them at https://github.com/apache/avro-rs/blob/8d6dd77454d6bdee39ccc54ae699d607d72ce320/avro/src/schema.rs#L179 too.

martin-g:
Sure!

martin-g:
This has been fixed with #382

PR #382, opened 2025-12-29
feat: Duration inner fixed schema and serialization

jdarais:
This PR adds some missing functionality for the Duration type:

- Schema::Duration is updated to contain a FixedSchema struct, which contains name, namespace, size, doc, etc. fields that can exist on Duration types. (#378)
- apache_avro's Duration type is updated to implement Serialize and Deserialize to support inclusion in structs de/serialized from/to avro "the serde way"
- Small fix to the Deserializer implementation in serde/de.rs to allow Value::Duration to be deserialized.  (Duration is deserialized as bytes.)

This PR also deprecates `SchemaKind::is_named` in favor of a new `Schema::is_named` function, which can correctly determine whether a schema is for a named type, even when it is for a logical type that could have either a named or non-named underlying type, (such as "decimal" or "uuid".)

One note on this change: I noticed that the derive macro for AvroSchema was using the Duration schema for core::time::Duration.  It is no longer trivial to create a schema for a Duration type, since the schema would need to be given a name.  It looks like other named types are excluded from these derive macros, so I just deleted the `impl_schema!(core::time::Duration, Schema::Duration)` line, but I'm not sure if there's something better that can be done here.

### DIFF
--- a/avro/src/schema_equality.rs
+++ b/avro/src/schema_equality.rs
@@ -163,9 +163,9 @@ impl SchemataEq for StructFieldEq {
             ) => {
                 self.compare(items_one, items_two)
             }
+            (Schema::Array(_), _) => false,
             (Schema::Duration(FixedSchema { size: size_one, ..}), Schema::Duration(FixedSchema { size: size_two, ..})) => size_one == size_two,
             (Schema::Duration(_), _) => false,
-            (Schema::Array(_), _) => false,
             (
                 Schema::Map(MapSchema { types: types_one, ..}),
                 Schema::Map(MapSchema { types: types_two, ..})
--- a/avro/src/serde/de.rs
+++ b/avro/src/serde/de.rs
@@ -475,8 +475,14 @@ impl<'de> de::Deserializer<'de> for &Deserializer<'de> {
             Value::Bytes(ref bytes) | Value::Fixed(_, ref bytes) => {
                 visitor.visit_byte_buf(bytes.to_owned())
             }
+            Value::Uuid(ref u) => visitor.visit_byte_buf(Vec::from(u.as_bytes())),
+            Value::Decimal(ref d) => visitor.visit_byte_buf(d.to_vec()?),
+            Value::Duration(ref d) => {
+                let d_bytes: [u8; 12] = d.into();
+                visitor.visit_byte_buf(Vec::from(d_bytes))
+            }
             _ => Err(de::Error::custom(format!(
-                "Expected a String|Bytes|Fixed, but got {:?}",
+                "Expected a String|Bytes|Fixed|Uuid|Decimal|Duration, but got {:?}",
                 self.input
             ))),
         }


avro/src/decode.rs
```diff
@@ -179,7 +179,7 @@ pub(crate) fn decode_internal<R: Read, S: Borrow<Schema>>(
         Schema::LocalTimestampMillis => zag_i64(reader).map(Value::LocalTimestampMillis),
         Schema::LocalTimestampMicros => zag_i64(reader).map(Value::LocalTimestampMicros),
         Schema::LocalTimestampNanos => zag_i64(reader).map(Value::LocalTimestampNanos),
-        Schema::Duration => {
+        Schema::Duration(_) => {
```
Kriskras99: It would be good to check if the FixedSchema is of size 12 (and error otherwise)

avro/src/duration.rs
```diff
@@ -134,12 +136,102 @@ impl From<Duration> for [u8; 12] {
...
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        let value_bytes: [u8; 12] = self.into();
+        serializer.serialize_bytes(&value_bytes)
+    }
+}
+
+struct DurationVisitor;
```
Kriskras99: Inline this struct and the implementation into the `deserialize` function, there is no value in it existing outside of `Deserialize`

avro/src/duration.rs
```diff
@@ -134,12 +136,102 @@ impl From<Duration> for [u8; 12] {
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::types::Value;
+    use anyhow::anyhow;
+    use apache_avro_test_helper::TestResult;
+
+    #[test]
+    fn test_duration_from_value() -> TestResult {
```
Kriskras99: Change all new tests to start with `avro_rs_382`, so we can easily get back to the PR

avro/src/schema_equality.rs
```diff
@@ -268,6 +267,29 @@ mod tests {
     test_primitives!(LocalTimestampMillis);
     test_primitives!(LocalTimestampNanos);
 
+    #[test]
+    fn test_avro_3939_compare_schemata_duration() {
```
Kriskras99: This should be `avro_rs_382`

Kriskras99 (commented):
First glance looks good, I'll have a more in depth look tomorrow

avro/src/serde/de.rs
```diff
@@ -354,9 +354,13 @@ impl<'de> de::Deserializer<'de> for &Deserializer<'de> {
                 Value::BigDecimal(ref big_decimal) => {
                     visitor.visit_str(big_decimal.to_plain_string().as_str())
                 }
-                _ => Err(de::Error::custom(format!(
```
martin-g, line 357: This is no more needed because the check is exhaustive, right ?

avro/src/schema.rs
```diff
@@ -134,7 +134,7 @@ pub enum Schema {
     /// An instant in local time represented as the number of nanoseconds after the UNIX epoch.
     LocalTimestampNanos,
     /// An amount of time defined by a number of months, days and milliseconds.
-    Duration,
+    Duration(FixedSchema),
```
martin-g, line 137: I am not sure about this.
To prevent API breaks in the future it would be better to use `Duration(DurationSchema)` where:
```rust
enum DurationSchema {
  Fixed(FixedSchema)
}
```
This way if some day Duration is represented by another (e.g. more compact) way it will be easier to add a second variant to DurationSchema.
Similar to UuidSchema.

avro/src/schema.rs
martin-g, line 1274: Please add an arm for Schema::Duration too.
Maybe also a new test for schema equality. Currently two duration schemas with different attributes would match or not depending on `include_attributes`

martin-g, line 1329: Please add support for Schema::Duration here too

martin-g, line 1345: And here

avro/src/types.rs
martin-g: Old issue: `GetDecimalFixedBytes` is not the correct error type here.
```suggestion
                    return Err(Details::ResolveDuration(Value::Fixed(size, bytes.clone())).into());
```

avro/src/schema_equality.rs
```diff
@@ -268,6 +267,29 @@ mod tests {
...
+            size: 12,
+            aliases: None,
+            doc: None,
+            default: None,
+            attributes: BTreeMap::new(),
+        });
+        let specification_eq_res = SPECIFICATION_EQ.compare(&schema_one, &schema_two);
+        let struct_field_eq_res = STRUCT_FIELD_EQ.compare(&schema_one, &schema_two);
+        assert_eq!(specification_eq_res, struct_field_eq_res)
+    }
```
martin-g, line 294: Let's add two more (negative) tests:
1) schemas with different names
2) schemas with different sizes

avro/src/duration.rs
```diff
@@ -1,3 +1,5 @@
+use serde::{Deserialize, Serialize, de};
```
martin-g: Please move this import below the ASF licence header

avro/src/schema.rs
martin-g, line 498: Please add Schema::Duration support here

avro/src/schema.rs
jdarais: I suspect that switching from `Duration(FixedSchema)` to `Duration(DurationSchema)` down the road may not be a whole lot more disruptive than starting with `Duration(DurationSchema)` and adding a new enum value down the road.  In the latter case, library consumers would still need to update match statements to accommodate the new enum variant, and nothing would have prevented consumers from writing code that assumes that duration is a fixed type.

Admittedly, having to migrate consumer code from using a `FixedSchema` to using an enum would be slightly more disruptive, but I think it would be worth it to not have to deal with the extra enum layer for now.  Let me know what you think.

Kriskras99: I agree with @jdarais, I think it's unlikely that Duration will get a new schema and code using it will break anyway.
Let's keep it at `Duration(FixedSchema)`

martin-g: OK!
Please continue with `Duration(FixedSchema)`.
I will make the change in a follow-up PR and decide. I like the codebase being consistent.

avro/src/schema.rs
```diff
@@ -2629,7 +2663,6 @@ pub mod derive {
     impl_schema!(f64, Schema::Double);
     impl_schema!(String, Schema::String);
     impl_schema!(uuid::Uuid, Schema::Uuid(UuidSchema::String));
-    impl_schema!(core::time::Duration, Schema::Duration);
```
Kriskras99, line 2632: The old code generated the following schema:
```json
{"type":{"name":"duration","type":"fixed","size":12}}
```
I think it would be good to keep it:
```rust
impl AvroSchemaComponent for core::time::Duration {
    fn get_schema_in_ctxt(
        named_schemas: &mut HashMap<String, Schema>,
        enclosing_namespace: &Namespace,
    ) -> Schema {
        let name = Name {
            name: "duration".to_string(),
            namespace: enclosing_namespace.clone(),
        };
        named_schemas
            .entry(name.to_string())
            .or_insert(Schema::Duration(FixedSchema {
                name,
                aliases: None,
                doc: None,
                size: 12,
                default: None,
                attributes: Default::default(),
            }))
            .clone()
    }
}
```
If `enclosing_namespace` is `Some` the old schema would stay the same (namespace would not be added), while in the above code it would be added. So that is a change, but I do think it's for the better.

avro/src/writer.rs
```diff
@@ -1047,7 +1047,14 @@ mod tests {
...
         logical_type_test(
             r#"{"type": {"type": "fixed", "name": "duration", "size": 12}, "logicalType": "duration"}"#,
-            &Schema::Duration,
+            &Schema::Duration(FixedSchema {
+                name: Name::from("duration"),
+                aliases: None,
+                doc: None,
+                size: 12,
+                default: None,
+                attributes: BTreeMap::new(),
```
Kriskras99: Using `Default::default()` means the new import on line 775 is not necessary.

avro/src/schema.rs
Kriskras99: @martin-g what do you think?

jdarais: I went ahead and added this to the PR, since it's probably better than removing the `AvroSchemaComponent` for `core::time::Duration` entirely.  I'm happy to make any other changes based on feedback.

avro/src/serde/de.rs
martin-g, line 484: I think we missed to add arms for Uuid, Decimal and Duration here to support `from_value::<Vec<u8>>(...)`

avro/src/schema.rs
```diff
@@ -2722,6 +2755,29 @@ pub mod derive {
             T::get_schema_in_ctxt(named_schemas, enclosing_namespace)
         }
     }
+
+    impl AvroSchemaComponent for core::time::Duration {
+        fn get_schema_in_ctxt(
+            named_schemas: &mut Names,
+            enclosing_namespace: &Namespace,
+        ) -> Schema {
+            let name = Name {
```
martin-g, line 2764: `let name = Name::new("duration")?.fully_qualified_name(enclosing_namespace);`

avro/src/schema_equality.rs
```diff
@@ -165,6 +163,8 @@ impl SchemataEq for StructFieldEq {
             ) => {
                 self.compare(items_one, items_two)
             }
+            (Schema::Duration(FixedSchema { size: size_one, ..}), Schema::Duration(FixedSchema { size: size_two, ..})) => size_one == size_two,
+            (Schema::Duration(_), _) => false,
             (Schema::Array(_), _) => false,
```
```suggestion
```

avro/src/schema_equality.rs
```diff
@@ -165,6 +163,8 @@ impl SchemataEq for StructFieldEq {
             ) => {
                 self.compare(items_one, items_two)
             }
```
martin-g, line 165:
```suggestion
            }
            (Schema::Array(_), _) => false,
```

avro/src/schema.rs
Kriskras99: `fully_qualified_name` takes the `Name` by reference, so it clones the original `Name`. It is probably optimized away, but we should maybe think about adding a `Name::new_with_namespace` function (not for this PR)

avro/src/serde/de.rs
jdarais: Ah you're right, I've added those cases

avro/src/schema.rs
jdarais: It looks like the `AvroSchemaComponent::get_schema_in_ctxt` function doesn't return a Result, so we'd have to do something like `Name::new("duration").unwrap().fully_qualified_name(enclosing_namespace);`.  Is that preferred over constructing the Name struct directly, or is there another function to construct a Name that can't fail that we could use here?

avro/src/schema_equality.rs
jdarais: I didn't realize I could commit these suggestions directly from the PR page.  I included these changes in my latest revision of the PR.

avro/src/schema.rs
martin-g: OK! Let's leave this as is for now!

martin-g (approved):

PR #393, opened 2026-01-08
fix: Print the Debug impl of TestError for more info

### LINKED ISSUE #370: `TestResult` does not display the whole error
Failed test with `TestError`:
```
    Backtrace omitted. Run with RUST_BACKTRACE=1 to display it.
    Run with RUST_BACKTRACE=full to include source snippets.

      apache_avro::error::Error: Failed to parse schema from JSON
    in avro/src/schema.rs, line 7418
    thread: schema::tests::avro_rs_339_schema_ref_uuid
```
Failed test with `unwrap()`:
```
    Backtrace omitted. Run with RUST_BACKTRACE=1 to display it.
    Run with RUST_BACKTRACE=full to include source snippets.

      called `Result::unwrap()` on an `Err` value: Error { details: Failed to parse schema from JSON: expected `,` or `}` at line 11 column 25 }
    in avro/src/schema.rs, line 7436
    thread: schema::tests::avro_rs_339_schema_ref_uuid
```

I think this is because `TestError` uses the `Display` trait instead of the `Debug` trait when printing errors.

martin-g:
Fixes #370

In addition make use of TestResult in two unit tests in schema.rs

### DIFF
--- a/avro_test_helper/src/lib.rs
+++ b/avro_test_helper/src/lib.rs
@@ -56,6 +56,8 @@ pub struct TestError;
 /// A converter of any error into [TestError].
 /// It is used to print better error messages in the tests.
 /// Borrowed from <https://bluxte.net/musings/2023/01/08/improving_failure_messages_rust_tests/>
+// The Display bound is needed so that the `From` implementation doesn't
+// apply to `TestError` itself.
 impl<Err: std::fmt::Display + std::fmt::Debug> From<Err> for TestError {
     #[track_caller]
     fn from(err: Err) -> Self {


avro_test_helper/src/lib.rs
```diff
@@ -56,10 +56,10 @@ pub struct TestError {}
 /// A converter of any error into [TestError].
 /// It is used to print better error messages in the tests.
 /// Borrowed from <https://bluxte.net/musings/2023/01/08/improving_failure_messages_rust_tests/>
-impl<Err: std::fmt::Display> From<Err> for TestError {
+impl<Err: std::fmt::Display + std::fmt::Debug> From<Err> for TestError {
```
Kriskras99: The Display bound can be removed

martin-g: I tried but it failed with:
```
error[E0119]: conflicting implementations of trait `From<TestError>` for type `TestError`
  --> avro_test_helper/src/lib.rs:59:1
   |
59 | impl<Err: std::fmt::Debug> From<Err> for TestError {
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: conflicting implementation in crate `core`:
           - impl<T> From<T> for T;

For more information about this error, try `rustc --explain E0119`.
error: could not compile `apache-avro-test-helper` (lib) due to 1 previous error
warning: build failed, waiting for other jobs to finish...
```
Let me try harder!

Kriskras99: Maybe remove the `#[derive(Debug)]` on `TestError`?

martin-g: This helped! Thanks!

martin-g: Actually, it broke things ...

Kriskras99: Crap, apparently `#[test]` requires the return type to implement `Debug`.
Then just keep the `Display` bound and add a comment that it is needed so that the `From` implementation doesn't apply to `TestError`

Issue #365, opened 2025-12-13
Enums don't serialize correctly


antonsmetanin:
Enum serialization doesn't seem to work correctly. I've prepared a test that demonstrates the issue, but basically I couldn't find a way to serialize an enum to match the following schema, other than manually building a `Value`, and even then schema resolution fails to find the correct union variant:

```json
{
    "name": "Root",
    "type": "record",
    "fields": [
        {"name": "field_union", "type": [
            {
                "name": "A",
                "type": "record",
                "fields": []
            },
            {
                "name": "B",
                "type": "record",
                "fields": []
            },
            {
                "name": "C",
                "type": "record",
                "fields": [
                    {"name": "field_a", "type": "long"},
                    {"name": "field_b", "type": ["null", "string"]}
                ]
            },
            {
                "name": "D",
                "type": "record",
                "fields": [
                    {"name": "field_a", "type": "float"},
                    {"name": "field_b", "type": "int"}
                ]
            }
        ]},
        {"name": "field_f", "type": "string"}
    ]
}
```

Here's the enum definition that's supposed to work with this schema, but it doesn't currently:
```rust
#[derive(Serialize)]
struct Root {
    field_union: Enum,
    field_f: String,
}

#[derive(Serialize)]
enum Enum {
    A {},
    B {},
    C {
        field_a: i64,
        field_b: Option<String>,
    },
    D {
        field_a: f32,
        field_b: i32
    },
}
```

So I've looked into the implementation, and found two issues:

1. For some reason enum variants get serialized as records with two fields:
- "type" with an Avro enum for the discriminator,
- "value" with a union variant containing the fields.
instead of just serializing union variant directly. Because of this bytecode ends up being different from what's expected (there's an additional enum byte for "type"), but also schema resolution fails because it expects to find actual fields at the top level, while they are under "value".
2. Separately from the first issue, even if you try to use `resolve()` on a `Value` built manually, schema resolution ignores the variant number coming from `Value::Union` and tries to find a schema by simply matching the fields, which causes serialization to lose field data in some cases, depending on the order in which these variants are listed in the schema. In the example above, `A` is tested first, and since it doesn't have any fields, it matches any value.

Kriskras99:
Over the past month I've done a lot of thinking about how to improve the enum support in Avro and I think I finally have something that can work. It's something that's workable in both the derive code and the general code, and remains compatible with current schemas.

# Current state of Rust enum support in Avro

## Plain enums
[...]

PR #394, opened 2026-01-08
feat!:Make AvroSchemaComponent available without `derive` feature

### LINKED ISSUE #363: Shouldn't `AvroSchemaComponent` always be available
Currently `AvroSchemaComponent` is gated behind the `derive` feature flag, even though in the documentation for `AvroSchema` (which is not gated) we strongly recommend to implement it instead of `AvroSchema`.

As it does not have any dependencies on `apache-avro-derive` I don't see why it's gated.

martin-g:
Fixes #363

### DIFF
--- a/avro/src/schema.rs
+++ b/avro/src/schema.rs
@@ -2673,21 +2673,10 @@ where
     T: AvroSchemaComponent,
 {
     fn get_schema_in_ctxt(named_schemas: &mut Names, enclosing_namespace: &Namespace) -> Schema {
-        let inner_schema = T::get_schema_in_ctxt(named_schemas, enclosing_namespace);
-
-        let mut variants = vec![Schema::Null];
-        match inner_schema {
-            Schema::Union(union_schema) => {
-                // Flatten and avoid repeating `null`.
-                for schema in union_schema.schemas {
-                    if !matches!(schema, Schema::Null) {
-                        variants.push(schema);
-                    }
-                }
-            }
-            Schema::Null => { /* keep just null */ }
-            other => variants.push(other),
-        }
+        let variants = vec![
+            Schema::Null,
+            T::get_schema_in_ctxt(named_schemas, enclosing_namespace),
+        ];
 
         Schema::Union(
             UnionSchema::new(variants).expect("Option<T> must produce a valid (non-nested) union"),
--- a/avro/tests/avro_schema_component.rs
+++ b/avro/tests/avro_schema_component.rs
@@ -15,28 +15,19 @@
 // specific language governing permissions and limitations
 // under the License.
 
+use apache_avro::{AvroSchemaComponent, Schema};
+use std::collections::HashMap;
+
 #[test]
 fn avro_rs_394_avro_schema_component_without_derive_feature() {
-    use apache_avro::{AvroSchemaComponent, Schema};
-    use std::collections::HashMap;
-
-    // Verify basic types work without derive feature
     let schema = i32::get_schema_in_ctxt(&mut HashMap::default(), &None);
     assert!(matches!(schema, Schema::Int));
 }
 
 #[test]
+#[should_panic(expected = "Option<T> must produce a valid (non-nested) union")]
 fn avro_rs_394_avro_schema_component_nested_options() {
-    use apache_avro::{AvroSchemaComponent, Schema};
-    use std::collections::HashMap;
+    type VeryOptional = Option<Option<i32>>;
 
-    type VeryOptional = Option<Option<Option<Option<Option<i32>>>>>;
-
-    let schema = VeryOptional::get_schema_in_ctxt(&mut HashMap::default(), &None);
-    match schema {
-        Schema::Union(union_schema) => {
-            assert_eq!(union_schema.variants(), &[Schema::Null, Schema::Int]);
-        }
-        _ => panic!("Unexpected schema type: {schema}"),
-    };
+    let _schema = VeryOptional::get_schema_in_ctxt(&mut HashMap::default(), &None);
 }


Kriskras99:
I think everything in the `derive` module should always be available. Otherwise you have the trait available, but can't use it because you can't implement foreign traits on foreign types.

avro/src/schema.rs
```diff
@@ -2568,215 +2568,185 @@ fn field_ordering_position(field: &str) -> Option<usize> {
...
-                    .enumerate()
-                    .map(|(idx, s)| (SchemaKind::from(s), idx))
-                    .collect(),
-            })
-        }
+impl<T> AvroSchemaComponent for Option<T>
+where
+    T: AvroSchemaComponent,
+{
+    fn get_schema_in_ctxt(named_schemas: &mut Names, enclosing_namespace: &Namespace) -> Schema {
```
martin-g, line 2675: This will lead to a nested Schema::Union if someone uses `Option<Option<T>>`

avro/src/schema.rs
```diff
@@ -2568,215 +2568,195 @@ fn field_ordering_position(field: &str) -> Option<usize> {
 
 /// Trait for types that serve as an Avro data model. Derive implementation available
 /// through `derive` feature. Do not implement directly!
-/// Implement `apache_avro::schema::derive::AvroSchemaComponent` to get this trait
+/// Implement [AvroSchemaComponent] to get this trait
```
Kriskras99:
```suggestion
/// Implement [`AvroSchemaComponent`] to get this trait
```

avro/src/schema.rs
```diff
@@ -2568,215 +2568,195 @@ fn field_ordering_position(field: &str) -> Option<usize> {
...
-    where
-        T: AvroSchemaComponent,
-    {
-        fn get_schema_in_ctxt(
-            named_schemas: &mut Names,
-            enclosing_namespace: &Namespace,
-        ) -> Schema {
-            Schema::array(T::get_schema_in_ctxt(named_schemas, enclosing_namespace))
+            Schema::Null => { /* keep just null */ }
+            other => variants.push(other),
```
Kriskras99: Original
```rust
impl<T> AvroSchemaComponent for Option<T>
where
    T: AvroSchemaComponent,
{
    fn get_schema_in_ctxt(named_schemas: &mut Names, enclosing_namespace: &Namespace) -> Schema {
        let inner_schema = T::get_schema_in_ctxt(named_schemas, enclosing_namespace);

        let mut variants = vec![Schema::Null];
        match inner_schema {
            Schema::Union(union_schema) => {
                // Flatten and avoid repeating `null`.
                for schema in union_schema.schemas {
                    if !matches!(schema, Schema::Null) {
                        variants.push(schema);
                    }
                }
            }
            Schema::Null => { /* keep just null */ }
            other => variants.push(other),
        }

        Schema::Union(
            UnionSchema::new(variants).expect("Option<T> must produce a valid (non-nested) union"),
        )
    }
}
```
Suggestion
```rust
impl<T> AvroSchemaComponent for Option<T>
where
    T: AvroSchemaComponent,
{
    fn get_schema_in_ctxt(named_schemas: &mut Names, enclosing_namespace: &Namespace) -> Schema {
        let mut variants = vec![Schema::Null, T::get_schema_in_ctxt(named_schemas, enclosing_namespace)];

        Schema::Union(
            UnionSchema::new(variants).expect("Option<T> must produce a valid (non-nested) union"),
        )
    }
}
```
This transformation is wrong.
`Some(Some(Some(None)))` will now deserialize as `None`.
I think it would be better to fallback to one of the enum representations I described in https://github.com/apache/avro-rs/issues/365#issuecomment-3730627801, but that should be done in the PR implementing proper enum support.
I'll start working on that.

avro/tests/avro_schema_component.rs
```diff
@@ -0,0 +1,42 @@
...
+    type VeryOptional = Option<Option<Option<Option<Option<i32>>>>>;
+
+    let schema = VeryOptional::get_schema_in_ctxt(&mut HashMap::default(), &None);
+    match schema {
+        Schema::Union(union_schema) => {
+            assert_eq!(union_schema.variants(), &[Schema::Null, Schema::Int]);
+        }
+        _ => panic!("Unexpected schema type: {schema}"),
+    };
+}
```
Kriskras99, line 33:
```suggestion
#[test]
#[should_panic(expected = "Option<T> must produce a valid (non-nested) union")]
fn avro_rs_394_avro_schema_component_nested_options() {
    use apache_avro::{AvroSchemaComponent, Schema};
    use std::collections::HashMap;

    type VeryOptional = Option<Option<Option<Option<Option<i32>>>>>;

    let _schema = VeryOptional::get_schema_in_ctxt(&mut HashMap::default(), &None);
}
```

avro/src/schema.rs
martin-g: https://doc.rust-lang.org/rustdoc/write-documentation/linking-to-items-by-name.html

> Backticks around the link will be stripped, so [`Option`] will correctly link to Option.

avro/src/schema.rs
martin-g: I don't understand the suggestion. The Github UI shows many removed lines.
E.g. why `impl_schema!(bool, Schema::Boolean);` is being removed ?!

Kriskras99: I don't know what went wrong, I've updated the original comment to just show it manually

avro/src/schema.rs
Kriskras99: It will strip the backticks only for checking where to link. So we can highlight `AvroSchemaComponent` as code and have it link without having to manually specify where it links to.
(Try it out! Run `cargo doc --open` with and without the backticks)

avro/src/schema.rs
Kriskras99: Basically we panic if we get a nested union, but I plan to have the new enum support working before we release 0.22. So the user will never encounter this.

PR #397, opened 2026-01-09
feat(derive): allow overriding the schema for a field and change the schema for `uuid::Uuid`

Kriskras99:
A user can now override the schema used by `#[derive(AvroSchema)]` using `#[avro(with)]`.

If no argument is supplied to `with`, then it will use the same module as specified in `#[serde(with = ".."]`. It is a compile error if the Serde attribute is missing.

If an argument is supplied to `with`, it will call this function with the same arguments as `AvroSchemaComponent::get_schema_in_ctxt`.

I've also replaced the `AvroSchemaComponent` implementation for `uuid::Uuid` with the `Schema::Uuid(UuidSchema::Fixed)` representation. Users who want the old schema can use `#[avro(with)]` to revert back.

User-facing documentation will follow in a separate PR, which also contains an overhaul of the entire Serde/derive documentation of these crates.

### DIFF
--- a/avro_derive/src/lib.rs
+++ b/avro_derive/src/lib.rs
@@ -173,16 +173,27 @@ fn get_data_struct_schema_def(
                 let schema_expr = match field_attrs.with {
                     With::Trait => type_to_schema_expr(&field.ty)?,
                     With::Serde(path) => {
-                        quote! {#path::get_schema_in_ctxt(named_schemas, enclosing_namespace)}
+                        quote! { #path::get_schema_in_ctxt(named_schemas, enclosing_namespace) }
                     }
-                    With::Expr(expr) => match expr {
-                        Expr::Closure(closure) => {
-                            let _inputs = &closure.inputs;
-                            let body = &closure.body;
-                            quote! {#body}
+                    With::Expr(Expr::Closure(closure)) => {
+                        if closure.inputs.is_empty() {
+                            quote! { (#closure)() }
+                        } else {
+                            return Err(vec![syn::Error::new(
+                                field.span(),
+                                "Expected closure with 0 parameters",
+                            )]);
                         }
-                        _ => quote! {#expr(named_schemas, enclosing_namespace)},
-                    },
+                    }
+                    With::Expr(Expr::Path(path)) => {
+                        quote! { #path(named_schemas, enclosing_namespace) }
+                    }
+                    With::Expr(_expr) => {
+                        return Err(vec![syn::Error::new(
+                            field.span(),
+                            "Invalid expression, expected function or closure",
+                        )]);
+                    }
                 };
                 record_field_exprs.push(quote! {
                     schema_fields.push(::apache_avro::schema::RecordField {
--- a/avro_derive/tests/derive.rs
+++ b/avro_derive/tests/derive.rs
@@ -1843,6 +1843,10 @@ fn avro_rs_397_with() {
             {
                 "name":"b",
                 "type":"long"
+            },
+            {
+                "name":"c",
+                "type":"bytes"
             }
         ]
     }
@@ -1872,6 +1876,8 @@ fn avro_rs_397_with() {
         a: String,
         #[avro(with = long_schema)]
         b: i32,
+        #[avro(with = module::get_schema_in_ctxt)]
+        c: String,
     }
 
     assert_eq!(schema, Foo::get_schema());
@@ -1959,6 +1965,38 @@ fn avro_rs_397_uuid() {
 }
 
 #[test]
+fn avro_rs_397_derive_with_expr_lambda() {
+    let schema = r#"
+  {
+    "type":"record",
+    "name":"Foo",
+    "fields": [
+      {
+        "name": "_a",
+        "type": "bytes"
+      },
+      {
+        "name": "_b",
+        "type": "int"
+      }
+    ]
+  }"#;
+
+    let expected_schema = Schema::parse_str(schema).unwrap();
+
+    #[derive(AvroSchema)]
+    struct Foo {
+        #[avro(with = || Schema::Bytes)]
+        _a: String,
+        _b: i32,
+    }
+
+    let derived_schema = Foo::get_schema();
+
+    assert_eq!(expected_schema, derived_schema);
+}
+
+#[test]
 fn avro_rs_401_do_not_match_typename() {
     #[expect(nonstandard_style, reason = "It needs to be exactly this")]
     type f32 = f64;
@@ -2041,36 +2079,4 @@ fn avro_rs_401_supported_type_variants() {
 
     let schema = Schema::parse_str(schema).unwrap();
     assert_eq!(schema, Foo::get_schema());
-}
-
-#[test]
-fn avro_rs_397_derive_with_expr_lambda() {
-    let schema = r#"
-  {
-    "type":"record",
-    "name":"Foo",
-    "fields": [
-      {
-        "name": "_a",
-        "type": "bytes"
-      },
-      {
-        "name": "_b",
-        "type": "int"
-      }
-    ]
-  }"#;
-
-    let expected_schema = Schema::parse_str(schema).unwrap();
-
-    #[derive(AvroSchema)]
-    struct Foo {
-        #[avro(with = || Schema::Bytes)]
-        _a: String,
-        _b: i32,
-    }
-
-    let derived_schema = Foo::get_schema();
-
-    assert_eq!(expected_schema, derived_schema);
 }
--- a/avro_derive/tests/ui/avro_rs_397_with_closure_parameters.rs
+++ b/avro_derive/tests/ui/avro_rs_397_with_closure_parameters.rs
@@ -1,0 +1,27 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+use apache_avro::{AvroSchema, Schema};
+
+#[derive(AvroSchema)]
+struct Foo {
+    #[avro(with = |_named_schemas, _enclosing_namespace| Schema::Bytes)]
+    a: String,
+    b: i32,
+}
+
+pub fn main() {}
--- a/avro_derive/tests/ui/avro_rs_397_with_closure_parameters.stderr
+++ b/avro_derive/tests/ui/avro_rs_397_with_closure_parameters.stderr
@@ -1,0 +1,14 @@
+error: Expected closure with 0 parameters
+  --> tests/ui/avro_rs_397_with_closure_parameters.rs:22:5
+   |
+22 | /     #[avro(with = |_named_schemas, _enclosing_namespace| Schema::Bytes)]
+23 | |     a: String,
+   | |_____________^
+
+warning: unused import: `Schema`
+  --> tests/ui/avro_rs_397_with_closure_parameters.rs:18:31
+   |
+18 | use apache_avro::{AvroSchema, Schema};
+   |                               ^^^^^^
+   |
+   = note: `#[warn(unused_imports)]` (part of `#[warn(unused)]`) on by default
--- a/avro_derive/tests/ui/avro_rs_397_with_expr_string.stderr
+++ b/avro_derive/tests/ui/avro_rs_397_with_expr_string.stderr
@@ -1,8 +1,6 @@
-error[E0618]: expected function, found `&'static str`
-  --> tests/ui/avro_rs_397_with_expr_string.rs:22:19
+error: Invalid expression, expected function or closure
+  --> tests/ui/avro_rs_397_with_expr_string.rs:22:5
    |
-20 | #[derive(AvroSchema)]
-   |          ---------- call expression requires function
-21 | struct Foo {
-22 |     #[avro(with = "Schema::Bytes")]
-   |                   ^^^^^^^^^^^^^^^
+22 | /     #[avro(with = "Schema::Bytes")]
+23 | |     a: String,
+   | |_____________^


avro_derive/src/attributes/avro.rs
```diff
@@ -97,6 +99,21 @@ impl VariantAttributes {
     }
 }
 
+/// How to get the schema for this field.
```
martin-g:
```suggestion
/// How to get the schema for a field.
```

avro_derive/src/attributes/avro.rs
```diff
@@ -137,6 +154,16 @@ pub struct FieldAttributes {
     /// [`serde::FieldAttributes::flatten`]: crate::attributes::serde::FieldAttributes::flatten
     #[darling(default)]
     pub flatten: bool,
+    /// How to get the schema for this field.
```
martin-g:
```suggestion
    /// How to get the schema for a field.
```

avro_derive/src/attributes/avro.rs
```diff
@@ -137,6 +154,16 @@ pub struct FieldAttributes {
...
+    ///
+    /// By default uses `<T as AvroSchemaComponent>::get_schema_in_ctxt`.
+    ///
+    /// When it's provided without an argument (`#[avro(with)]`), it will use the function `get_schema_in_ctxt` defined
+    /// in the same module as the `#[serde(with = "..")]` attribute.
```
martin-g: This is not very clear to me. It sounds like it will use the serde attribute's module.

avro_derive/src/attributes/avro.rs
```diff
@@ -137,6 +154,16 @@ pub struct FieldAttributes {
...
+    ///
+    /// When it's provided with an argument (`#[avro(with = ..)]`), it will use that function.
```
martin-g: Let's try to make the documentation clear enough without assuming the reader knows what `#[serde(with)]` is and how to use it.

I think it would be better to use:
```suggestion
    /// When it's provided with an argument (`#[avro(with = some_fn)]`), it will use that function.
```

Currently with `..` I have to guess that it is the referenced function, and my second guess would be how to refer to it:
 - as a string without a parenthesis: `with = "some_function"`
 - as a string with parenthesis: `with = "some_function()"`
 - or as an expression: `with = some_function`

avro_derive/src/attributes/mod.rs
```diff
@@ -145,13 +146,50 @@ impl VariantOptions {
...
+        match &avro {
+            avro::With::Trait => Ok(Self::Trait),
+            avro::With::Serde => {
+                if let Some(serde) = serde {
+                    let path = Path::from_string(serde).unwrap();
+                    Ok(Self::Serde(path))
+                } else {
+                    Err(syn::Error::new(
+                        span,
+                        "`#[avro(with)]` requires `#[serde(with = \"..\")]` or provide a function to call `#[avro(width = ..)]`",
```
```suggestion
                        "`#[avro(with)]` requires `#[serde(with = \"..\")]` or provide a function to call `#[avro(with = ..)]`",
```

avro_derive/tests/ui/avro_rs_397_with_expr_type.rs
```diff
@@ -0,0 +1,27 @@
...
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+use apache_avro::{AvroSchema, Schema};
+
+#[derive(AvroSchema)]
+struct Foo {
+    #[avro(with = Schema::Bytes)]
```
martin-g, line 22: Let's add one more test with:
```rust
#[avro(with = || Schema::Bytes)]
```
Would inline closure like this work ?!

avro_derive/src/attributes/mod.rs
```diff
@@ -145,13 +146,50 @@ impl VariantOptions {
...
+    fn from_avro_and_serde(
+        avro: &avro::With,
+        serde: &Option<String>,
+        span: Span,
+    ) -> Result<Self, syn::Error> {
+        match &avro {
+            avro::With::Trait => Ok(Self::Trait),
+            avro::With::Serde => {
+                if let Some(serde) = serde {
+                    let path = Path::from_string(serde).unwrap();
```
martin-g: Let's return an Err instead of `.unwrap()`-ing here.

avro/src/schema.rs
```diff
@@ -2765,17 +2767,49 @@ pub mod derive {
...
+    impl AvroSchemaComponent for uuid::Uuid {
+        /// The schema is [`Schema::Uuid`] with the name `uuid`.
+        ///
+        /// The underlying schema is [`Schema::Fixed`] with a size of 16.
+        fn get_schema_in_ctxt(
+            named_schemas: &mut Names,
+            enclosing_namespace: &Namespace,
+        ) -> Schema {
+            let name = Name {
+                name: "uuid".to_string(),
```
martin-g: I have a bad feeling about the hardcoded names we use here: `"uuid"` and `"duration"` above.
They may clash with user-defined schemas with the same names ...
I have no better solution than using some name prefix, e.g. "avro_derived_uuid"

avro/src/schema.rs
martin-g, line 1: We should add Schema::Duration to https://github.com/apache/avro-rs/blob/f3dd04f6db7d2604fdf87d8a003dced94d8b66ea/avro/src/schema.rs#L592 and treat it as a named schema. It is a bug from an earlier PR.

avro_derive/src/attributes/avro.rs
```diff
@@ -97,6 +99,21 @@ impl VariantAttributes {
     }
 }
 
+/// How to get the schema for this field.
+#[derive(Debug, FromMeta, PartialEq, Default)]
+#[darling(from_expr = |expr| Ok(With::Expr(expr.clone())))]
+pub enum With {
+    /// Use `<T as AvroSchemaComponent>::get_schema_with_ctxt`.
```
martin-g:
```suggestion
    /// Use `<T as AvroSchemaComponent>::get_schema_in_ctxt`.
```

avro_derive/src/attributes/avro.rs
```diff
@@ -97,6 +99,21 @@ impl VariantAttributes {
...
+    #[default]
+    #[darling(skip)]
+    Trait,
+    /// Use `module::get_schema_with_ctxt` where the module is defined by Serde's `with` attribute.
```
martin-g:
```suggestion
    /// Use `module::get_schema_in_ctxt` where the module is defined by Serde's `with` attribute.
```

avro_derive/src/attributes/mod.rs
```diff
@@ -145,13 +146,50 @@ impl VariantOptions {
     }
 }
 
+/// How to get the schema for this field or variant.
+#[derive(Debug, PartialEq, Default)]
+pub enum With {
+    /// Use `<T as AvroSchemaComponent>::get_schema_with_ctxt`.
```
```suggestion
    /// Use `<T as AvroSchemaComponent>::get_schema_in_ctxt`.
```

avro_derive/src/attributes/mod.rs
```diff
@@ -145,13 +146,50 @@ impl VariantOptions {
...
+    #[default]
+    Trait,
+    /// Use `module::get_schema_with_ctxt` where the module is defined by Serde's `with` attribute.
```
```suggestion
    /// Use `module::get_schema_in_ctxt` where the module is defined by Serde's `with` attribute.
```

avro/src/schema.rs
Kriskras99: A user will only see these schemas if they derived `AvroSchema` and have a `uuid` or `duration` field. If they have a nameclash somewhere, the can use `rename` on the other uuid struct, or `with` on the field with uuid to override the schema.

So I think having this as a default is fine, because users can (with this PR) override the schema if needed (also if they want the `uuid` `string` version)

martin-g:
> For the `with` attribute the argument has no quotes, matching Bon (https://bon-rs.com/reference/builder/member/with#with). Serde always has quotes (https://serde.rs/field-attrs.html#serialize_with). Which one do we prefer?

I prefer the no-quotes one. It looks cleaner and more type safe to me. Although the quoted one in Serde is equally type safe, I guess.

avro/src/schema.rs
martin-g: Good point!

I should probably just try it myself but it is easier to ask you 😅 

What is the produced schema for this struct ?
```rust
#[derive(AvroSchema)]
struct Foo {
  #[serde(rename="baz")]
  bar: Uuid
}
```

Let's add it as a test case!

avro_derive/src/attributes/avro.rs
Kriskras99: That's correct!

This allows a user to define their overrides for a type in one module, without having to give the path twice:
```rust
#[derive(AvroSchema, Serialize, Deserialize)]
struct Foo {
    #[avro(with)]
    #[serde(with = "bar")]
    field: String,
}

mod bar {
    pub fn serialize<S>(&String, S) -> Result<S::Ok, S::Error> where S: Serializer {
        todo!()
    }

    pub fn deserialize<'de, D>(D) -> Result<String, D::Error> where D: Deserializer<'de> {
        todo!()
    }

    pub fn get_schema_in_ctxt(named_schemas: &mut Names, enclosing_namespace: &Namespace) -> Schema {
        todo!()
    }
}
```

avro_derive/tests/ui/avro_rs_397_with_expr_type.rs
Kriskras99: Unfortunately not, although `bon` does support it. So I could dive into their code and figure out how they do it

martin-g: I think Darling's functionality for this is [Callable](https://github.com/TedDriggs/darling/blob/master/core/src/util/callable.rs)

avro_derive/src/attributes/avro.rs
Kriskras99: So what do you think? `with`'s primary usecase in Serde is being able to use foreign types that don't implement `Serialize`/`Deserialize`.
By having `#[avro(with)]` without an argument use the module that is provided in `#[serde(with = "..")]`, users can also provide `get_schema_in_ctxt` without having to repeat the module declaration (and without us having to support providing a module xor a function as an argument).

Kriskras99: @martin-g I think [test `avro_rs_397_avroschema_with_bytes`](https://github.com/apache/avro-rs/pull/397/changes#diff-7c49e5c1201c2931dcf016b83d362880d27a218e514a62543904c277961ce75bR509) and the changes I added in `bytes.rs`, shows how this can be very useful.

avro_derive/src/attributes/mod.rs
```diff
@@ -145,13 +146,55 @@ impl VariantOptions {
...
+        span: Span,
+    ) -> Result<Self, syn::Error> {
+        match &avro {
+            avro::With::Trait => Ok(Self::Trait),
+            avro::With::Serde => {
+                if let Some(serde) = serde {
+                    let path = Path::from_string(serde).map_err(|err| {
+                        syn::Error::new(
+                            span,
+                            format!("Expected a path for `#[serde(with = \"..\")]: {err:?}"),
```
```suggestion
                            format!("Expected a path for `#[serde(with = \"..\")]`: {err:?}"),
```

avro/src/bytes.rs
```diff
@@ -136,6 +164,29 @@ pub mod serde_avro_fixed {
...
+    };
+
+    /// Returns `Schema::Fixed(N)` named `serde_avro_fixed_{N}`
+    pub fn get_schema_in_ctxt<const N: usize>(
+        named_schemas: &mut Names,
+        enclosing_namespace: &Namespace,
+    ) -> Schema {
+        let name = Name::new(&format!("serde_avro_fixed_{N}"))
+            .expect("Name is valid")
+            .fully_qualified_name(enclosing_namespace);
```
martin-g, line 180: 👍🏻 
We should do the same for Uuid/Duration too: `Name::new().expect().fully_qualified_name()`

Kriskras99: Should I also change what is generated by the `AvroSchema` derive? Then we have it matching everywhere (I'll also make sure to update the docs of `AvroSchemaComponent`)

martin-g: Yes! Sounds good!

avro_derive/src/lib.rs
```diff
@@ -166,7 +170,13 @@ fn get_data_struct_schema_def(
                     None => quote! { None },
                 };
                 let aliases = preserve_vec(field_attrs.alias);
-                let schema_expr = type_to_schema_expr(&field.ty)?;
+                let schema_expr = match field_attrs.with {
+                    With::Trait => type_to_schema_expr(&field.ty)?,
+                    With::Serde(path) => {
+                        quote! {#path::get_schema_in_ctxt(named_schemas, enclosing_namespace)}
+                    }
+                    With::Expr(expr) => quote! {#expr(named_schemas, enclosing_namespace)},
```
martin-g: Is there anything else that mandates the `expr` to be a function ?
I don't see any restrictions in avro.rs and mod.rs to verify that the provided expression is a syn::Expr::Call.

My main question is: What leads to `expected function` in errors like `error[E0618]: expected function, found apache_avro::Schema` ?

I am experimenting with
```diff
                         quote! {#path::get_schema_in_ctxt(named_schemas, enclosing_namespace)}
                     }
-                    With::Expr(expr) => quote! {#expr(named_schemas, enclosing_namespace)},
+                    With::Expr(expr) => match expr {
+                        Expr::Closure(closure) => {
+                            dbg!("==========");
+                            quote! {#closure()}
+                        }
+                        _ => quote! {#expr(named_schemas, enclosing_namespace)},
+                    },
                 };
```
but it seems something earlier makes the check for a `function`.

martin-g: Actually the dbg!() code is executed ! I'm continuing my debug session

martin-g:
```diff
-                    With::Expr(expr) => quote! {#expr(named_schemas, enclosing_namespace)},
+                    With::Expr(expr) => match expr {
+                        Expr::Closure(closure) => {
+                            let inputs = &closure.inputs;
+                            let body = &closure.body;
+                            dbg!("==========", &closure);
+                            dbg!(">>>>>", quote! {#body});
+                            quote! {#body}
+                        }
+                        _ => quote! {#expr(named_schemas, enclosing_namespace)},
+                    },
                 };
```
This seems to work for `|| Schema::Bytes` !

I guess we could make it work also for `|named_schemas, enclosing_namespace| Schema::Bytes`

martin-g: I have pushed the support for lambda/closure.
I think we should not support inputs for now.

~How about supporting also `#[avro(with = Schema::Bytes)]` too ? Seems reasonable to me.~
Using the lambda should be good enough!

Kriskras99: I don't think it's possible to distinguish between a function path `path::to::function` and a constructor `Schema::Bytes`.
Nice work adding the lambda!

PR #398, opened 2026-01-10
feat(derive): Support `#[serde(transparent)]`

Kriskras99:
We only support `#[serde(transparent)]` on structs, as I don't know what it would mean for a plain enum to be transparent (we could maybe emit a `null`).

### DIFF
--- a/avro_derive/src/lib.rs
+++ b/avro_derive/src/lib.rs
@@ -71,7 +71,7 @@ fn derive_avro_schema(input: DeriveInput) -> Result<TokenStream, Vec<syn::Error>
             Ok(create_trait_definition(input.ident, &input.generics, inner))
         }
         syn::Data::Union(_) => Err(vec![syn::Error::new(
-            input.span(),
+            input_span,
             "AvroSchema: derive only works for structs and simple enums",
         )]),
     }
@@ -257,6 +257,7 @@ fn get_transparent_struct_schema_def(
                     )]);
                 }
             }
+
             if let Some((field, attrs)) = found {
                 get_field_schema_expr(&field, attrs.with)
             } else {
--- a/avro_derive/tests/derive.rs
+++ b/avro_derive/tests/derive.rs
@@ -1998,28 +1998,13 @@ fn avro_rs_397_derive_with_expr_lambda() {
 
 #[test]
 fn avro_rs_398_transparent_with_skip() {
-    let schema = Schema::parse_str(
-        r#"
-    {
-        "type":"record",
-        "name":"Foo",
-        "fields": [
-            {
-                "name":"b",
-                "type":"long"
-            }
-        ]
-    }
-    "#,
-    )
-    .unwrap();
-
     fn long_schema(_named_schemas: &mut Names, _enclosing_namespace: &Namespace) -> Schema {
         Schema::Long
     }
 
     #[allow(dead_code)]
     #[derive(AvroSchema)]
+    #[serde(transparent)]
     struct Foo {
         #[serde(skip)]
         a: String,
@@ -2029,7 +2014,7 @@ fn avro_rs_398_transparent_with_skip() {
         c: String,
     }
 
-    assert_eq!(schema, Foo::get_schema());
+    assert_eq!(Schema::Long, Foo::get_schema());
 }
 
 #[test]


avro_derive/src/lib.rs
```diff
@@ -221,21 +235,26 @@ fn get_data_struct_schema_def(
...
 
+/// Generate a schema definition for a enum.
 fn get_data_enum_schema_def(
     full_schema_name: &str,
     doc: Option<String>,
     aliases: Vec<String>,
     rename_all: RenameRule,
-    e: &syn::DataEnum,
+    data_enum: &syn::DataEnum,
     error_span: Span,
```
martin-g: Above you renamed `error_span` to `ident_span`. Should we do the same here too ? For consistency

avro_derive/src/lib.rs
```diff
@@ -264,6 +283,69 @@ fn get_data_enum_schema_def(
...
+                    )]);
+                }
+                let field = fields_named
+                    .named
+                    .first()
+                    .expect("There is exactly one field");
+                let field_attrs = FieldOptions::new(&field.attrs, field.span())?;
+                if field_attrs != FieldOptions::default() {
+                    return Err(vec![syn::Error::new(
+                        input_span,
```
```suggestion
                        field.span(),
```
This would be more accurate, no ?

avro_derive/src/lib.rs
```diff
@@ -264,6 +283,69 @@ fn get_data_enum_schema_def(
...
+                input_span,
+                "AvroSchema derive does not work for tuple structs",
+            )]),
+            syn::Fields::Unit => Err(vec![syn::Error::new(
+                input_span,
+                "AvroSchema derive does not work for unit structs",
+            )]),
+        },
+        syn::Data::Enum(data_enum) => {
+            if data_enum
```
martin-g: Do we need the if/else at all ?
Should we just return an Err with a message like `#[serde(transparent)] is not supported on enums` ?

avro_derive/src/lib.rs
```diff
@@ -264,6 +283,69 @@ fn get_data_enum_schema_def(
...
+            } else {
+                Err(vec![syn::Error::new(
+                    input_span,
+                    "AvroSchema derive does not work for enums with non unit structs",
+                )])
+            }
+        }
+        syn::Data::Union(_) => Err(vec![syn::Error::new(
+            input_span,
+            "AvroSchema derive only works for structs and simple enums",
```
```suggestion
            "AvroSchema derive only works for structs with named fields",
```
We just errored above that enums are not supported, simple or not.

avro_derive/src/attributes/mod.rs
```diff
@@ -89,6 +84,19 @@ impl NamedTypeOptions {
                 "#[avro(rename_all = \"..\")] must match #[serde(rename_all = \"..\")], it's also deprecated. Please use only `#[serde(rename_all = \"..\")]`",
             ));
         }
+        if serde.transparent
+            && (serde.rename.is_some()
+                || avro.namespace.is_some()
```
martin-g, line 93:
```suggestion
                || avro.name.is_some()
                || avro.namespace.is_some()
```
`avro.name` is deprecated but still possible to be used, so maybe we should check it too ?

avro_derive/src/lib.rs
```diff
@@ -264,6 +283,69 @@ fn get_data_enum_schema_def(
...
+                        input_span,
+                        "#[serde(transparent)] is incompatible with all other attributes",
+                    )]);
+                }
+                let ty = &field.ty;
+                Ok(quote! {
+                    #ty::get_schema_in_ctxt(named_schemas, enclosing_namespace)
+                })
+            }
+            syn::Fields::Unnamed(_) => Err(vec![syn::Error::new(
```
martin-g: > #[serde(transparent)]
Serialize and deserialize a newtype struct or a braced struct with one field exactly the same as if its one field were serialized and deserialized by itself. 

Maybe we should support `newtype struct` too ?!

avro_derive/src/lib.rs
```diff
@@ -264,6 +283,69 @@ fn get_data_enum_schema_def(
...
+                    .first()
+                    .expect("There is exactly one field");
+                let field_attrs = FieldOptions::new(&field.attrs, field.span())?;
+                if field_attrs != FieldOptions::default() {
+                    return Err(vec![syn::Error::new(
+                        input_span,
+                        "#[serde(transparent)] is incompatible with all other attributes",
+                    )]);
+                }
+                let ty = &field.ty;
```
martin-g: Do we need to check the type of `ty` before doing the rest ?
E.g. in `#type_to_schema_expr()` we handle Type::Path, Type::Reference and Type::Array.
Here I think we should proceed only when it is Type::Path.

avro_derive/tests/serde.rs
```diff
@@ -314,6 +314,28 @@ mod container_attributes {
             "Invalid field name c",
         );
     }
+
+    #[test]
+    fn avro_rs_398_transparent() {
+        #[derive(Debug, Serialize, Deserialize, AvroSchema, Clone, PartialEq)]
+        #[serde(transparent)]
+        struct Foo {
+            a: String,
```
martin-g, line 323: Let's add a test like:
```rust
struct Foo<'a> {
    a: &'a str,
}
```
It will show whether Type::Reference needs special handling or not.

avro_derive/src/lib.rs
Kriskras99: Yes, but this is a general error as the `#[derive(AvroSchema)]` was put on a `union` which we don't support anyway. This error should be thrown irrespective of any attributes used

avro_derive/src/lib.rs
Kriskras99: I should just use `type_to_schema_expr` directly I think

avro_derive/src/lib.rs
Kriskras99: I was at one point thinking about allowing simple enums with one field to become `null`, but ended up not doing that

avro_derive/src/lib.rs
Kriskras99: It might be better to support `newtype struct` completely then (make it a record with fields named `1`, `2`, ...)

martin-g: You can leave that for a follow-up PR!

avro_derive/src/lib.rs
martin-g: This error is returned by get_transparent_data_schema_def(), so it is special for the `transparent` attribute.
The generic error is at https://github.com/apache/avro-rs/blob/6b9a15f7db96b165cb6839dc9a655fdf1b5b7ddb/avro_derive/src/lib.rs#L91

Kriskras99: Yeah, but `get_transparent_data_schema_def()` branches before that check, so it needs to be done in two locations

avro_derive/tests/derive.rs
```diff
@@ -1996,6 +1996,42 @@ fn avro_rs_397_derive_with_expr_lambda() {
...
+    )
+    .unwrap();
+
+    fn long_schema(_named_schemas: &mut Names, _enclosing_namespace: &Namespace) -> Schema {
+        Schema::Long
+    }
+
+    #[allow(dead_code)]
+    #[derive(AvroSchema)]
+    struct Foo {
```
martin-g, line 2008: There is no usage of `#[serde(transparent)]` here

avro_derive/src/lib.rs
```diff
@@ -34,86 +35,93 @@ use crate::{
...
+                    "AvroSchema: `#[serde(transparent)]` is only supported on structs",
+                )]);
+            }
+            let schema_def =
+                get_data_enum_schema_def(&named_type_options, data_enum, input.ident.span())?;
+            let inner = handle_named_schemas(named_type_options.name, schema_def);
+            Ok(create_trait_definition(input.ident, &input.generics, inner))
+        }
+        syn::Data::Union(_) => Err(vec![syn::Error::new(
+            input.span(),
```
```suggestion
            input_span,
```

avro_derive/src/lib.rs
```diff
@@ -250,23 +237,85 @@ fn get_data_struct_schema_def(
...
+                if attrs.skip {
+                    continue;
+                }
+                if found.replace((field, attrs)).is_some() {
+                    return Err(vec![syn::Error::new(
+                        input_span,
+                        "AvroSchema: #[serde(transparent)] is only allowed on structs with one unskipped field",
+                    )]);
+                }
+            }
```
martin-g, line 259:
```suggestion
            }

```

PR #401, opened 2026-01-13
fix(derive): do not match type by name

Kriskras99:
Users are free to define types matching built-in types. This probably won't happen often for the integer types, but I can see it happening for the `String` type.

### DIFF
--- a/avro_derive/src/lib.rs
+++ b/avro_derive/src/lib.rs
@@ -280,7 +280,7 @@ fn type_to_schema_expr(ty: &Type) -> Result<TokenStream, Vec<syn::Error>> {
         _ => Err(vec![syn::Error::new_spanned(
             ty,
             format!(
-                "AvroSchema: Unexpected type encountered, please open an issue if this kind of type should be supported: {ty:?}"
+                "AvroSchema: Unexpected type encountered! Please open an issue if this kind of type should be supported: {ty:?}"
             ),
         )]),
     }
--- a/avro_derive/tests/ui/avro_rs_401_unknown_type_variant.stderr
+++ b/avro_derive/tests/ui/avro_rs_401_unknown_type_variant.stderr
@@ -1,4 +1,4 @@
-error: AvroSchema: Unexpected type encountered, please open an issue if this kind of type should be supported: Type::BareFn { lifetimes: None, unsafety: None, abi: None, fn_token: Fn, paren_token: Paren, inputs: [], variadic: None, output: ReturnType::Type(RArrow, Type::Path { qself: None, path: Path { leading_colon: None, segments: [PathSegment { ident: Ident { ident: "String", span: #0 bytes(882..888) }, arguments: PathArguments::None }] } }) }
+error: AvroSchema: Unexpected type encountered! Please open an issue if this kind of type should be supported: Type::BareFn { lifetimes: None, unsafety: None, abi: None, fn_token: Fn, paren_token: Paren, inputs: [], variadic: None, output: ReturnType::Type(RArrow, Type::Path { qself: None, path: Path { leading_colon: None, segments: [PathSegment { ident: Ident { ident: "String", span: #0 bytes(882..888) }, arguments: PathArguments::None }] } }) }
   --> tests/ui/avro_rs_401_unknown_type_variant.rs:22:8
    |
 22 |     a: fn() -> String,


avro_derive/src/lib.rs
```diff
@@ -267,34 +267,9 @@ fn get_data_enum_schema_def(
...
-            "bool" => quote! {apache_avro::schema::Schema::Boolean},
-            "i8" | "i16" | "i32" | "u8" | "u16" => quote! {apache_avro::schema::Schema::Int},
-            "u32" | "i64" => quote! {apache_avro::schema::Schema::Long},
-            "f32" => quote! {apache_avro::schema::Schema::Float},
-            "f64" => quote! {apache_avro::schema::Schema::Double},
-            "String" | "str" => quote! {apache_avro::schema::Schema::String},
-            "char" => {
-                return Err(vec![syn::Error::new_spanned(
-                    ty,
-                    "AvroSchema: Cannot guarantee successful deserialization of this type",
```
martin-g, line 282: Maybe we should implement AvroSchemaComponent for `char` and `u64` and panic!() with these error messages ?!

avro_derive/tests/derive.rs
```diff
@@ -1827,3 +1827,31 @@ fn avro_rs_247_serde_flatten_support_with_skip() {
...
+                "name":"field",
+                "type":"double"
+            }
+        ]
+    }
+    "#;
+
+    let schema = Schema::parse_str(schema).unwrap();
+    assert_eq!(schema, Foo::get_schema());
+}
```
martin-g, line 1857: Let's add a test case for `str` too since it now uses AvroSchemaComponent too

avro/src/schema.rs
martin-g, line 2651: Do we also need similar impls for `AvroSchemaComponent for &T / &mut T` ?
To support `Option<&'a str>` for example

avro_derive/src/lib.rs
Kriskras99: I think it would be clearer to not implement it at all. We could add a note to the documentation of the traits (`AvroSchema` and `AvroSchemaComponent`) and the derive why those types aren't implemented.

avro/src/schema.rs
Kriskras99: Yes, that would be a good idea. It would also mean that we can simplify the derive implementation as we no longer have to specifically support references.

avro_derive/src/lib.rs
martin-g: Sounds good!

avro_derive/src/lib.rs
```diff
@@ -266,45 +265,24 @@ fn get_data_enum_schema_def(
...
+            "AvroSchema: derive does not support raw pointers",
+        )]),
+        Type::Tuple(_) => Err(vec![syn::Error::new_spanned(
+            ty,
+            "AvroSchema: derive does not support tuples",
+        )]),
+        _ => Err(vec![syn::Error::new_spanned(
+            ty,
+            format!(
+                "AvroSchema: Unexpected type encountered, please open an issue if this kind of type should be supported: {ty:?}"
```
```suggestion
                "AvroSchema: Unexpected type encountered! Please open an issue if this kind of type should be supported: {ty:?}"
```

avro_derive/tests/ui/avro_rs_401_unknown_type_variant.stderr
```diff
@@ -0,0 +1,5 @@
+error: AvroSchema: Unexpected type encountered, please open an issue if this kind of type should be supported: Type::BareFn { lifetimes: None, unsafety: None, abi: None, fn_token: Fn, paren_token: Paren, inputs: [], variadic: None, output: ReturnType::Type(RArrow, Type::Path { qself: None, path: Path { leading_colon: None, segments: [PathSegment { ident: Ident { ident: "String", span: #0 bytes(882..888) }, arguments: PathArguments::None }] } }) }
```
martin-g:
```suggestion
error: AvroSchema: Unexpected type encountered! Please open an issue if this kind of type should be supported: Type::BareFn { lifetimes: None, unsafety: None, abi: None, fn_token: Fn, paren_token: Paren, inputs: [], variadic: None, output: ReturnType::Type(RArrow, Type::Path { qself: None, path: Path { leading_colon: None, segments: [PathSegment { ident: Ident { ident: "String", span: #0 bytes(882..888) }, arguments: PathArguments::None }] } }) }
```

martin-g (approved):

avro/src/schema.rs
```diff
@@ -979,6 +979,10 @@ pub struct UnionSchema {
 
 impl UnionSchema {
     /// Creates a new UnionSchema from a vector of schemas.
+    ///
+    /// # Errors
+    /// Will return an error if `variants` has duplicate unnamed schemas or if `variants`
```
martin-g:
```suggestion
    /// Will return an error if `schemas` has duplicate unnamed schemas or if `schemas`
```

avro/src/schema.rs
```diff
@@ -1380,6 +1384,15 @@ impl Schema {
         })
     }
 
+    /// Returns a `Schema::Union` with the given variants.
+    ///
+    /// # Errors
+    /// Will return an error if `variants` has duplicate unnamed schemas or if `variants`
```
martin-g:
```suggestion
    /// Will return an error if `schemas` has duplicate unnamed schemas or if `schemas`
```

avro/src/schema.rs
```diff
@@ -1380,6 +1384,15 @@ impl Schema {
...
+    /// contains a union.
+    pub fn union(variants: Vec<Schema>) -> AvroResult<Schema> {
```
martin-g:
```suggestion
    pub fn union(schemas: Vec<Schema>) -> AvroResult<Schema> {
```

avro/src/schema.rs
```diff
@@ -1380,6 +1384,15 @@ impl Schema {
...
+        UnionSchema::new(variants).map(Schema::Union)
```
martin-g:
```suggestion
        UnionSchema::new(schemas).map(Schema::Union)
```

avro_derive/tests/derive.rs
```diff
@@ -1827,3 +1827,88 @@ fn avro_rs_247_serde_flatten_support_with_skip() {
...
+    #[expect(dead_code, reason = "We only check the schema")]
+    #[derive(AvroSchema)]
+    struct Foo<'a> {
+        one: f32,
+        two: std::string::String,
+        three: &'static i32,
+        four: &'a str,
+        five: &'a mut f64,
+        six: [u8; 5],
+        seven: [u8],
```
martin-g, line 1870:
```suggestion
        seven: &'a [u8],
```
to make it usable.

I wonder how does this compile at all.

Kriskras99: The AI is wrong, you can have one unsized type at the end of a struct (**or** it needs to be a reference)

martin-g: And how one sets a value for such field ?

Kriskras99: You have to construct it directly on the heap (the parent type becomes an unsized type), but I don't know the specifics.
As we only need to test the schema generated it's not a problem.

martin-g: I found the answer:
```rust
struct MyData {
    id: u32,          // Sized field (fixed offset)
    // Some other sized fields...
    content: [u8],    // UNSIZED field (must be last)
}

// You can't do this:
// struct Invalid {
//     content: [u8],
//     id: u32,
// }

// You interact via pointers (or references):
let data: &MyData = &MyData { id: 1, content: [10, 20, 30] };
// Or with Box:
let boxed_data: Box<MyData> = Box::new(MyData { id: 2, content: [40, 50] });
```

I learnt something new today!

Kriskras99: Nice!

PR #409, opened 2026-01-19
chore: Extract RecordField related structs to src/schema/record/field.rs

martin-g:
No other functional changes!

### DIFF
--- a/avro/src/schema/mod.rs
+++ b/avro/src/schema/mod.rs
@@ -926,7 +926,7 @@ enum RecordSchemaParseLocation {
 }
 
 #[derive(Default)]
-struct Parser {
+pub(crate) struct Parser {
     input_schemas: HashMap<Name, Value>,
     /// A map of name -> Schema::Ref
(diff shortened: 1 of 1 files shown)


avro/src/schema/mod.rs
```diff
@@ -40,9 +37,11 @@ use std::{
     fmt::Debug,
     hash::Hash,
     io::Read,
-    str::FromStr,
 };
-use strum_macros::{Display, EnumDiscriminants, EnumString};
+use strum_macros::{Display, EnumDiscriminants};
+
+pub mod record_field;
```
Kriskras99:
```suggestion
mod record_field;
```
Let's make the module private for now, as there's nothing there that isn't `pub use` in `schema.rs`.

Unless you plan on deprecating the `pub use` of course.

Kriskras99 (commented):
We could also do this:
```
schema
├── mod.rs
└── record
    ├── mod.rs
    └── field.rs
```
Then we have everything relevant to Record separated in a module.

PR #411, opened 2026-01-19
fix: Calculate the RecordSchema's `lookup` table on build

### LINKED ISSUE #403: `RecordSchema::builder()` requires `fields` and `lookup` to be set
```rust
RecordSchema::builder().name(Name::new("Something").unwrap()).build()
```
This is a compile error.

I think `fields` should be allowed to be empty, and that `lookup` should not be able to be set by the user and shoud instead be derived from `fields`.

martin-g:
Fixes #403

### DIFF
--- a/avro/src/schema/record/schema.rs
+++ b/avro/src/schema/record/schema.rs
@@ -72,10 +72,7 @@ mod tests {
     fn avro_rs_403_record_schema_builder_no_fields() -> TestResult {
         let name = Name::new("TestRecord")?;
 
-        let record_schema = RecordSchema::builder()
-            .name(name.clone())
-            .fields(vec![])
-            .build();
+        let record_schema = RecordSchema::builder().name(name.clone()).build();
 
         assert_eq!(record_schema.name, name);
         assert_eq!(record_schema.aliases, None);
@@ -93,7 +90,6 @@ mod tests {
 
         let record_schema = RecordSchema::builder()
             .name(name.clone())
-            .fields(vec![])
             .aliases(Some(vec!["alias_1".into()]))
             .build();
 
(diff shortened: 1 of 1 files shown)


avro/src/schema/record/schema.rs
```diff
@@ -49,3 +59,105 @@ pub(crate) enum RecordSchemaParseLocation {
...
+    use apache_avro_test_helper::TestResult;
+    use pretty_assertions::assert_eq;
+
+    #[test]
+    fn avro_rs_403_record_schema_builder_no_fields() -> TestResult {
+        let name = Name::new("TestRecord")?;
+
+        let record_schema = RecordSchema::builder()
+            .name(name.clone())
+            .fields(vec![])
```
martin-g: I does not make much sense to have a RecordSchema without fields but it was possible until now, so I left it as is.

Kriskras99: It's actually useful for dealing with Rust enums (https://github.com/apache/avro-rs/issues/365#issuecomment-3730627801)

avro/src/schema/record/schema.rs
```diff
@@ -34,12 +34,22 @@ pub struct RecordSchema {
     pub fields: Vec<RecordField>,
```
Kriskras99, line 35:
```suggestion
    #[builder(default)]
    pub fields: Vec<RecordField>,
```

avro/src/schema/record/schema.rs
```diff
@@ -34,12 +34,22 @@ pub struct RecordSchema {
...
     /// The `lookup` table maps field names to their position in the `Vec`
     /// of `fields`.
+    #[builder(skip = calculate_lookup_table(&fields))]
     pub lookup: BTreeMap<String, usize>,
     /// The custom attributes of the schema
-    #[builder(default = BTreeMap::new())]
+    #[builder(default = Default::default())]
```
Kriskras99:
```suggestion
    #[builder(default)]
```

Kriskras99 (approved):

PR #417, opened 2026-01-20
chore: Extract Name/Namespace/Alias/Aliases to src/schema/name.rs

martin-g:
No functional changes!

### DIFF
--- a/avro/src/schema/mod.rs
+++ b/avro/src/schema/mod.rs
@@ -38,7 +38,7 @@ use std::{
 use strum_macros::{Display, EnumDiscriminants};
 
 mod name;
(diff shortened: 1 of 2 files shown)


martin-g:
I will push it in a separate commit, so we can revert it if it is not a good idea.

Kriskras99 (changes requested):
I think it's better to keep those functions in `mod.rs`. They are small and not `name` specific (it just extracts a field from the enum).

I also think the `Documentation` type alias should stay in `mod.rs`. It's not related to `Name` and is only a type alias, so it's not saving any code either.

Kriskras99 (approved):

PR #420, opened 2026-01-21
fix: Support recursive types for `Schema::independent_canonical_form`

Kriskras99:
The provided test will fail on `main` with a stack overflow.

### DIFF
--- a/avro/src/schema/mod.rs
+++ b/avro/src/schema/mod.rs
@@ -913,15 +913,8 @@ impl Schema {
                     schema.denormalize(schemata, defined_names)?;
                 }
             }
-            Schema::Enum(EnumSchema { name, .. })
-            | Schema::Fixed(FixedSchema { name, .. })
-            | Schema::Decimal(DecimalSchema {
-                inner: InnerDecimalSchema::Fixed(FixedSchema { name, .. }),
-                ..
-            })
-            | Schema::Uuid(UuidSchema::Fixed(FixedSchema { name, .. }))
-            | Schema::Duration(FixedSchema { name, .. }) => {
-                defined_names.insert(name.clone());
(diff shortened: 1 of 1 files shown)


avro/src/schema/mod.rs
```diff
@@ -6453,4 +6476,43 @@ mod tests {
 
         Ok(())
     }
+
+    #[test]
+    fn avro_rs_420_independent_canonical_form() -> TestResult {
+        let record = Schema::Record(
```
martin-g: Any reason not to use a Schema::parse_str() here ? 
The String/JSON representation is easier to read and understand.

avro/src/schema/mod.rs
```diff
@@ -870,36 +870,59 @@ impl Schema {
...
             }
+            Schema::Enum(EnumSchema { name, .. })
+            | Schema::Fixed(FixedSchema { name, .. })
+            | Schema::Decimal(DecimalSchema {
+                inner: InnerDecimalSchema::Fixed(FixedSchema { name, .. }),
+                ..
+            })
+            | Schema::Uuid(UuidSchema::Fixed(FixedSchema { name, .. }))
+            | Schema::Duration(FixedSchema { name, .. }) => {
+                defined_names.insert(name.clone());
```
martin-g: I think this would be more future-proof if redone as:
```rust
schema if schema.is_named() => defined_names.insert(schema.name().clone())
```
This way if we add yet another named schema in the future we will have to add it only in `Schema::is_named()` and `Schema::name()` but no need to add it here

avro/src/schema/mod.rs
martin-g: nit: I think using Schema::parse_str("...") would be more readable than constructing the schemas with code.

PR #423, opened 2026-01-22
fix!: `Name` and `Alias` incorrectly implement `From<&str>`

Kriskras99:
The documentation for `From` mentions the following:

> The conversion is _infallible_: if the conversion can fail, use `TryFrom` instead; don’t provide a `From` impl that panics.

This removes the `From` implementation and implements `TryFrom` instead. It also implements `TryFrom<String>`. I would have prefered to implement `TryFrom<T>` where `T: AsRef<str>` but that's not possible in the current type resolver.

This is a **breaking** change, any user using `Name::into` or `Alias::into` will have to update their code.

### DIFF
--- a/avro/src/schema/name.rs
+++ b/avro/src/schema/name.rs
@@ -15,11 +15,11 @@
 // specific language governing permissions and limitations
 // under the License.
 
-use std::collections::HashMap;
-use std::fmt;
-
 use serde::{Deserialize, Serialize, Serializer};
 use serde_json::{Map, Value};
+use std::collections::HashMap;
+use std::fmt;
+use std::str::FromStr;
 
 use crate::{
     AvroResult, Error, Schema,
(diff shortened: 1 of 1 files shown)


avro/src/schema/record/schema.rs
```diff
@@ -90,11 +90,14 @@ mod tests {
 
         let record_schema = RecordSchema::builder()
             .name(name.clone())
-            .aliases(Some(vec!["alias_1".into()]))
+            .aliases(Some(vec!["alias_1".try_into().expect("Alias is valid")]))
```
martin-g: Can we have them both ?
This very first diff chunk easily convinces me that the From impl is very useful for such usages where the user is certain that it won't fail.

We just have to explain well in the rustdoc that if the string comes from user input and there is a chance that it may be invalid Name then TryFrom should be used.

WDYT ?

martin-g: Or maybe use [FromStr](https://doc.rust-lang.org/stable/std/str/trait.FromStr.html) ?!

Kriskras99: That is not possible because of a generic implementation by Rust:
```rust
impl<T, U> TryFrom<U> for T
where
    U: Into<T>,
```
In my opinion this is important enough to break code, people who don't care can just use `unwrap`/`?`/`expect`.

I did update the commit to use `?` in a lot of places.

Kriskras99: Implementing `FromStr` is a good idea, but I still want to get rid of `From`. It signals to the user that this is an infallible conversion, while it really isn't.

PR #429, opened 2026-01-23
feat: `RecordSchema/RecordField::builder().name()` accepts `Into<String>`

### LINKED ISSUE #419: `RecordSchemaBuilder::name` and `RecordFieldBuilder::name` should accept `Into<String>`
This allows one to do:
```rust
RecordField::builder().name("name").schema(Schema::Null).build()
```
Instead of:
```rust
RecordField::builder().name("name".to_string()).schema(Schema::Null).build()
```

Probably also applicable to other builders we have.

martin-g:
Fixes #419

Usage:

RecordFieldBuilder:
```rust
RecordField::builder().name("str_slice").build();
```

RecordSchemaBuilder:
```rust
RecordSchema::builder().try_name("str_slice")?.build();
```

### DIFF
--- a/avro/src/schema/mod.rs
+++ b/avro/src/schema/mod.rs
@@ -1710,7 +1710,7 @@ mod tests {
 
         let schema_c_expected = Schema::Record(
             RecordSchema::builder()
-                .name(Name::new("C")?)
+                .try_name("C")?
                 .fields(vec![
                     RecordField::builder()
                         .name("field_one".to_string())
--- a/avro/src/schema/union.rs
+++ b/avro/src/schema/union.rs
@@ -151,16 +151,8 @@ mod tests {
     #[test]
     fn avro_rs_402_new_union_schema_duplicate_names() -> TestResult {
         let res = UnionSchema::new(vec![
-            Schema::Record(
-                RecordSchema::builder()
-                    .name("Same_name".try_into()?)
-                    .build(),
-            ),
-            Schema::Record(
-                RecordSchema::builder()
-                    .name("Same_name".try_into()?)
-                    .build(),
-            ),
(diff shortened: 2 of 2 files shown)


avro/src/schema/record/schema.rs
```diff
@@ -42,6 +45,18 @@ pub struct RecordSchema {
...
 
+use record_schema_builder::{IsUnset, SetName, State};
+impl<S: State> RecordSchemaBuilder<S> {
+    /// Try to set a Name from the given string.
+    pub fn try_name(self, name: impl Into<String>) -> AvroResult<RecordSchemaBuilder<SetName<S>>>
+    where
+        <S as State>::Name: IsUnset,
+    {
+        let name = Name::try_from(name.into())?;
+        Ok(self.name(name))
```
Kriskras99:
```suggestion
    pub fn try_name<T>(self, name: T) -> Result<RecordSchemaBuilder<SetName<S>>, <T as TryInto<Name>>::Error>
    where
        <S as State>::Name: IsUnset,
        T: TryInto<Name>,
    {
        let name = name.try_into()?;
        Ok(self.name(name))
```
This makes the function more generic and allows users to convert from their own types.

avro/src/schema/record/schema.rs
```diff
@@ -42,6 +45,18 @@ pub struct RecordSchema {
     pub attributes: BTreeMap<String, Value>,
 }
 
+use record_schema_builder::{IsUnset, SetName, State};
```
Kriskras99: Let's move these to the top of the file, or replace the import by full paths in the `impl`

Kriskras99: Or move the import inside the `try_name` function

PR #430, opened 2026-01-23
Move the `bytes` module to the `serde` module

Kriskras99:
It is polluting the root of the docs and really belongs in the `serde` module.

This PR also makes the `serde` module public. In a next PR I'll move most Serde related documentation to this module, as the root of the docs is currently very overwhelming for new users.

This change is backwards compatible.

### DIFF
--- a/avro/src/serde/with.rs
+++ b/avro/src/serde/with.rs
@@ -284,7 +284,7 @@ pub mod fixed_opt {
 /// Note that `&[u8]` are always serialized as [`Value::Bytes`]. However,
 /// both [`Value::Bytes`] and [`Value::Fixed`] can be deserialized as `&[u8]`.
 ///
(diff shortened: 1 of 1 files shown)


avro/src/serde/with.rs
```diff
@@ -0,0 +1,793 @@
...
+}
+
+/// Efficient (de)serialization of Avro bytes/fixed borrowed values.
+///
+/// This module is intended to be used through the Serde `with` attribute.
+///
+/// Note that `&[u8]` are always serialized as [`Value::Bytes`]. However,
+/// both [`Value::Bytes`] and [`Value::Fixed`] can be deserialized as `&[u8]`.
+///
+/// Use [`apache_avro::serde::slice`] for optional bytes/fixed borrowed values.
```
martin-g:
```suggestion
/// Use [`apache_avro::serde::slice_opt`] for optional bytes/fixed borrowed values.
```

PR #433, opened 2026-01-23
chore: Move `AvroSchema` and `AvroSchemaComponent` to the `serde` module

Kriskras99:
After this, all the Serde stuff is in one place.

Then I'll finish the documentation work.

### DIFF
--- a/avro_derive/tests/derive.rs
+++ b/avro_derive/tests/derive.rs
@@ -19,7 +19,6 @@ use apache_avro::{
     AvroSchema, AvroSchemaComponent, Reader, Schema, Writer, from_value,
     schema::{Alias, EnumSchema, FixedSchema, Name, Names, Namespace, RecordSchema},
 };
-use apache_avro_derive::*;
 use proptest::prelude::*;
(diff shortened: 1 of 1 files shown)


avro_derive/tests/derive.rs
```diff
@@ -17,10 +17,8 @@
 
 use apache_avro::{
     Reader, Schema, Writer, from_value,
-    schema::{
-        Alias, AvroSchema, AvroSchemaComponent, EnumSchema, FixedSchema, Name, Names, Namespace,
-        RecordSchema,
-    },
+    schema::{Alias, EnumSchema, FixedSchema, Name, Names, Namespace, RecordSchema},
+    serde::{AvroSchema, AvroSchemaComponent},
```
martin-g: Those are also exported in the root (`apache_avro::AvroSchema`).

avro/src/serde/derive.rs
```diff
@@ -0,0 +1,450 @@
+use crate::Schema;
```
martin-g, line 18: Missing license

PR #438, opened 2026-01-26
fix: Ensure SpecificSingleObjectWriter writes header on every call

### LINKED ISSUE #431: Question: Clarify intended usage of `SpecificSingleObjectWriter` and `header_written` flag
## Context

I'm using `SpecificSingleObjectWriter` for Avro single object encoding and noticed an inconsistency in behavior between `SpecificSingleObjectWriter::write_ref()` and `GenericSingleObjectWriter::write_value_ref()`.

## Current Behavior

`GenericSingleObjectWriter::write_value_ref()` always writes the header on every call:

```rust
pub fn write_value_ref<W: Write>(&mut self, v: &Value, writer: &mut W) -> AvroResult<usize> {
    let original_length = self.buffer.len();  // header is pre-stored in buffer
    // ...
    write_value_ref_owned_resolved(&self.resolved, v, &mut self.buffer)?;
    writer.write_all(&self.buffer)?;  // writes header + data
    self.buffer.truncate(original_length);  // resets to just header
    Ok(len)
}
```

However, `SpecificSingleObjectWriter` has a `header_written: bool` field and previously only wrote the header on the first call to `write_ref()`:

```rust
pub fn write_ref<W: Write>(&mut self, data: &T, writer: &mut W) -> AvroResult<usize> {
    if !self.header_written {
        bytes_written += writer.write(self.inner.buffer.as_slice())?;
        self.header_written = true;
    }
    // ... serialize data without header on subsequent calls
}
```

## Questions

1. **What is the intended use case for `header_written`?** 
   - Is `SpecificSingleObjectWriter` meant for streaming multiple records after a single header (like the object container format)?
   - Or is it meant for single-object encoding where each message should be independently decodable (requiring header on each message)?

2. **Why the inconsistency with `GenericSingleObjectWriter`?**
   - `GenericSingleObjectWriter::write_value_ref()` writes header every time
   - `SpecificSingleObjectWriter::write_ref()` only wrote header once

## Current State

I've modified `write_ref()` to always write the header (matching `GenericSingleObjectWriter` behavior) and removed the `header_written` field. This allows reusing *one* writer for multiple independent messages without the cost of reconstructing the writer each time.
This is kind of a band aid for - this issue is just for clarification.

flxo:
### Problem

`SpecificSingleObjectWriter` had a bug where it only wrote the single-object encoding header on the first call to `write_ref()`. Subsequent calls would write only the data without the header, making those messages unreadable.

This violated the Avro single-object encoding specification, which requires
that each message be independently decodable with its own header (2-byte marker + 8-byte schema fingerprint). This is critical for use cases like message queues where individual messages are stored or transmitted separately.

### Solution

This PR removes the `header_written` flag from `SpecificSingleObjectWriter` and ensures that every call to `write_ref()` and `write()` writes a complete single-object encoded message (header + data). This matches the behavior of `GenericSingleObjectWriter` and conforms to the Avro specification.

### Changes

- Removed the `header_written` field from `SpecificSingleObjectWriter`
- Modified `write_ref()` to always write the header before writing data
- Updated documentation to clarify that each call produces an independently decodable message
- Added test verification that multiple writes to the same buffer each produce valid, independently decodable messages

### Test Plan

The updated test in `avro/src/writer.rs:1573` verifies that:
- Multiple calls to `write_ref()` produce identical output to calling `write_value()` multiple times
- Each write produces a complete, independently decodable message
- Output matches `GenericSingleObjectWriter` behavior

Fixes #431

### DIFF
--- a/avro/src/writer.rs
+++ b/avro/src/writer.rs
@@ -648,8 +648,8 @@ where
         Ok(bytes_written)
     }
 
-    /// Write the Serialize object to the provided Write object. 
-    /// 
+    /// Write the Serialize object to the provided Write object.
+    ///
     /// Returns the number of bytes written.
     ///
     /// Each call writes a complete single-object encoded message (header + data),


avro/src/writer.rs
```diff
@@ -632,24 +630,27 @@ where
...
-        if !self.header_written {
-            bytes_written += writer
-                .write(self.inner.buffer.as_slice())
-                .map_err(Details::WriteBytes)?;
-            self.header_written = true;
-        }
+        // Always write the header for each message (single object encoding requires
+        // each message to be independently decodable)
+        let mut bytes_written = writer
+            .write(self.inner.buffer.as_slice())
```
```suggestion
            .write_all(self.inner.buffer.as_slice())
```
like GenericSingleObjectWriter

avro/examples/specific_single_object.rs
```diff
@@ -32,24 +33,29 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
-                vec![
-                    195, 1, 166, 59, 243, 49, 82, 230, 8, 161, 54, 6, 102, 111, 111
-                ]
-            );
-        }
-        Err(err) => {
-            panic!("Error during serialization: {err:?}");
+    let reader = SpecificSingleObjectReader::<Test>::new()?;
+
+    for test in repeat(test).take(2) {
```
martin-g:
```suggestion
    for test in repeat_n(test, 2) {
```

avro/examples/specific_single_object.rs
```diff
@@ -17,6 +17,7 @@
 
 use apache_avro::{AvroSchema, SpecificSingleObjectReader, SpecificSingleObjectWriter};
 use serde::{Deserialize, Serialize};
+use std::iter::repeat;
```
martin-g:
```suggestion
use std::iter::repeat_n;
```

PR #439, opened 2026-01-26
fix: Resolve Schema::Ref in SpecificSingleObjectWriter::write_ref()

flxo:
## Summary

Fixes a bug in `SpecificSingleObjectWriter::write_ref()` where `Schema::Ref` types failed to resolve during serialization of recursive schemas.

## Problem

The `write_ref()` method was calling `write_avro_datum_ref()` which created a serializer with an empty names map. This caused `Schema::Ref` types to fail resolution during serialization, preventing recursive schemas from being properly serialized.

## Solution

Modified `SpecificSingleObjectWriter::write_ref()` to:
- Use the resolved schema names from the inner `GenericSingleObjectWriter`'s `ResolvedOwnedSchema`
- Pass these names to the serializer, enabling proper `Schema::Ref` resolution

Also updated `write_avro_datum_ref()` to create a `ResolvedSchema` from the input schema and use its names map instead of an empty map.

## Test Plan

Added `test_specific_single_object_writer_with_schema_ref()` which:
- Defines a recursive `Node` schema with a self-referencing field
- Creates a nested node structure (parent with child)
- Verifies the writer successfully serializes the recursive structure
- Validates the serialized output contains both "parent" and "child" values

## Open questions:

<del>- [x] This change makes `write_avro_datum_ref` use `SchemAwareWriteSerializer` which is probably more expensive than a plain `write_avro_datum_ref`. Shall this be part of the MR?</del>

### DIFF
--- a/avro/src/writer.rs
+++ b/avro/src/writer.rs
@@ -21,7 +21,7 @@ use crate::{
     encode::{encode, encode_internal, encode_to_vec},
     error::Details,
     headers::{HeaderBuilder, RabinFingerprintHeader},
-    schema::{Name, ResolvedOwnedSchema, ResolvedSchema, Schema},
+    schema::{ResolvedOwnedSchema, ResolvedSchema, Schema},
     serde::{AvroSchema, ser_schema::SchemaAwareWriteSerializer},
     types::Value,
 };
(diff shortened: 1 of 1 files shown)


avro/src/writer.rs
```diff
@@ -643,7 +643,18 @@ where
...
         }
 
-        bytes_written += write_avro_datum_ref(&self.schema, data, writer)?;
+        // Use the resolved schema names from the inner GenericSingleObjectWriter
+        // to properly resolve Schema::Ref during serialization
+        let names_ref = self
+            .inner
+            .resolved
+            .get_names()
+            .iter()
```
martin-g, line 651: Why do we need to iterate + map + collect ?
Can't we just use the result of `get_names()` ?

flxo: I we need a `NamesRef` and `get_names` return `Names`:

`ResolvedOwnedSchema::get_names()` returns [Names](https://github.com/apache/avro-rs/blob/6218cd723a02e947c5d799065506a026c92388b5/avro/src/schema/name.rs#L50) which is the "owned" version `pub type Names = HashMap<Name, Schema>;`. 

[SchemaAwareSerializer::new](https://github.com/apache/avro-rs/blob/6218cd723a02e947c5d799065506a026c92388b5/avro/src/serde/ser_schema.rs#L618) takes a `NamesRef` which is `pub type NamesRef<'a> = HashMap<Name, &'a Schema>;`.

I haven't found a way to avoid the nasty map duplication. I decided to keep it in here - it's a onetime thing.

Kriskras99 (commented):
I don't think this is the right way to solve this. To properly fix this, `ResolvedOwnedSchema` needs to be reworked, I've started work on this in a branch and I'll open a draft PR tomorrow

PR #441, opened 2026-01-26
fix: Don't allow serializing `bytes` as a `Uuid::String`

### LINKED ISSUE #440: Schema type for `Uuid` type is overridden by `set_serde_human_readable()` parameter on serialization
Hello team!

Apache Avro Specification states that:

> A `uuid` logical type annotates an Avro `string` or `fixed` of length 16.

However, in the current mainline (#6218cd7) schema requirements for underlying type are ignored on uuids serialisation, and the type depends on `apache_avro::util::set_serde_human_readable()` parameter: when `true` uuids are serialised into strings, when `false` - into bytes.
This behaviour differs from latest released version `rel/release-0.21.0`, where the type depends on the schema.

It seems the behaviour from `rel/release-0.21.0` is more correct. Could it be restored?

Attaching here an integration test for convenience.
```rust
use apache_avro::{Reader, Schema, Uuid, Writer};
use std::error::Error;

#[test]
fn test_uuid_serliazation_deserialization() -> Result<(), Box<dyn Error>> {
    // This line makes the test to pass.
    // Currently it's passigng only with `set_serde_human_readable(true)`
    // It worked correctly both with `true` an `false in `rel/release-0.21.0`.
    // apache_avro::util::set_serde_human_readable(true);

    let raw_schema = r#" 
    {
        "name": "uuid",
        "type": "string",
        "logicalType": "uuid"
    }
    "#;

    let initial_id = Uuid::parse_str("550e8400-e29b-41d4-a716-446655440000")?;

    let mut writer = Writer::new(&schema, Vec::new())?;
    writer.append_ser(&initial_id)?;

    let input = writer.into_inner()?;
    let reader = Reader::with_schema(&schema, &input[..])?;
    
    for value in reader {
        println!("{value:?}");
        assert_eq!(initial_id, apache_avro::from_value::<Uuid>(&value?)?);
    }

    Ok(())
}
```

Kriskras99:
Closes #440 

This would silently write unexpected values

### DIFF
--- a/avro/src/serde/ser_schema.rs
+++ b/avro/src/serde/ser_schema.rs
@@ -1150,6 +1150,9 @@ impl<'s, W: Write> SchemaAwareWriteSerializer<'s, W> {
             Schema::String | Schema::Bytes | Schema::Uuid(UuidSchema::String) => {
                 self.write_bytes(value.as_bytes())
             }
+            Schema::Uuid(UuidSchema::Bytes | UuidSchema::Fixed(_)) => {
+                Err(create_error("Expected bytes but got a string. Did you mean to use `Schema::Uuid(UuidSchema::String)` or `utils::serde_set_human_readable(false)`?".to_string()))
+            }
             Schema::BigDecimal => {
                 // If we get a string for a `BigDecimal` type, expect a display string representation, such as "12.75"
                 let decimal_val =
@@ -1224,10 +1227,13 @@ impl<'s, W: Write> SchemaAwareWriteSerializer<'s, W> {
                     self.write_bytes(value)
                 } else {
                     Err(create_error(format!(
-                        "Expected 16 bytes for `Schema::Uuid(Bytes) but got {} bytes",
+                        "Expected 16 bytes for `Schema::Uuid(Bytes)` but got {} bytes",
                         value.len()
                     )))
                 }
+            }
+            Schema::Uuid(UuidSchema::String) if value.len() == 16 => {
+                Err(create_error("Expected a string, but got 16 bytes. Did you mean to use `Schema::Uuid(UuidSchema::Fixed)` or `utils::serde_set_human_readable(true)`?".to_string()))
             }
             Schema::Fixed(fixed_schema) | Schema::Uuid(UuidSchema::Fixed(fixed_schema)) => {
                 if value.len() == fixed_schema.size {
--- a/avro/tests/serde_human_readable_false.rs
+++ b/avro/tests/serde_human_readable_false.rs
@@ -17,6 +17,7 @@
 
 use apache_avro::{AvroSchema, Schema, SpecificSingleObjectWriter, schema::UuidSchema};
 use apache_avro_test_helper::TestResult;
+use pretty_assertions::assert_eq;
 use serde::{Deserialize, Serialize};
 use uuid::Uuid;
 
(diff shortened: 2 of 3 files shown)


avro/tests/serde_human_readable_false.rs
```diff
@@ -60,3 +61,82 @@ fn avro_rs_53_uuid_with_fixed() -> TestResult {
...
+        inner: Uuid,
+    }
+    let uuid = CustomUuid {
+        inner: Uuid::parse_str("550e8400-e29b-41d4-a716-446655440000")?,
+    };
+    let mut buffer = Vec::new();
+
+    assert!(!apache_avro::util::set_serde_human_readable(false));
+    let mut writer = SpecificSingleObjectWriter::with_capacity(64)?;
+    assert!(writer.write(uuid, &mut buffer).unwrap_err().to_string().contains("Failed to serialize value of type bytes using schema Uuid(String): 55e840e29b41d4a7164466554400. Cause: Expected String, Bytes, Uuid, BigDecimal, Fixed, Duration, Decimal, Ref or Union schema. Got: Uuid"));
```
martin-g: `Expected ..., Uuid, ... schema. Got: Uuid"`
This error message needs improvement.

avro/tests/serde_human_readable_true.rs
```diff
@@ -59,3 +60,77 @@ fn avro_rs_53_uuid_with_string_true() -> TestResult {
...
+    };
+    let mut buffer = Vec::new();
+
+    assert!(apache_avro::util::set_serde_human_readable(true));
+    let mut writer = SpecificSingleObjectWriter::with_capacity(64)?;
+    writer.write(uuid, &mut buffer)?;
+
+    assert_eq!(
+        String::from_utf8_lossy(&buffer),
+        "�\u{1}'G�8�[\u{4}�H550e8400-e29b-41d4-a716-446655440000"
```
martin-g, line 84: nit: maybe assert `.contains(uuid_str)` ?!

avro/tests/serde_human_readable_true.rs
```diff
@@ -59,3 +60,77 @@ fn avro_rs_53_uuid_with_string_true() -> TestResult {
...
+    let uuid = CustomUuid {
+        inner: Uuid::parse_str("550e8400-e29b-41d4-a716-446655440000")?,
+    };
+    let mut buffer = Vec::new();
+
+    assert!(apache_avro::util::set_serde_human_readable(true));
+    let mut writer = SpecificSingleObjectWriter::with_capacity(64)?;
+    assert_eq!(
+        writer.write(uuid, &mut buffer).unwrap_err().to_string(),
+        "Failed to serialize value of type string using schema Uuid(Bytes): 550e8400-e29b-41d4-a716-446655440000. Cause: Expected: Uuid. Got: String"
```
```suggestion
        "Failed to serialize value of type string using schema Uuid(Bytes): 550e8400-e29b-41d4-a716-446655440000. Cause: Expected: Uuid(Bytes). Got: String"
```
Can we make the error message more helpful ?!

avro/src/serde/ser_schema.rs
```diff
@@ -1218,10 +1218,17 @@ impl<'s, W: Write> SchemaAwareWriteSerializer<'s, W> {
...
-            | Schema::Bytes
-            | Schema::Uuid(UuidSchema::Bytes | UuidSchema::String)
-            | Schema::BigDecimal => self.write_bytes(value),
+            Schema::String | Schema::Bytes | Schema::BigDecimal => self.write_bytes(value),
+            Schema::Uuid(UuidSchema::Bytes) => {
+                if value.len() == 16 {
+                    self.write_bytes(value)
+                } else {
+                    Err(create_error(format!(
+                        "Expected 16 bytes for `Schema::Uuid(Bytes) but got {} bytes",
```
```suggestion
                        "Expected 16 bytes for `Schema::Uuid(Bytes)` but got {} bytes",
```

avro/tests/serde_human_readable_true.rs
Kriskras99: The annoying thing about `assert!(.contains())` is that you don't see whats wrong if the test does fail

PR #443, opened 2026-01-27
chore: Less cloning while dealing with ResolvedSchema

martin-g:
Return reference fpr get_schemata()
Use interior mutation for ResolvedSchema::names_ref

### DIFF
--- a/avro/src/schema/resolve.rs
+++ b/avro/src/schema/resolve.rs
@@ -22,21 +22,22 @@ use crate::schema::{
 };
 use crate::{AvroResult, Error, Schema};
 use std::borrow::Borrow;
+use std::cell::{Ref, RefCell};
 use std::collections::HashMap;
 
 #[derive(Debug)]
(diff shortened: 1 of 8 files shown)


avro/src/schema/resolve.rs
```diff
@@ -69,31 +70,31 @@ impl<'s> ResolvedSchema<'s> {
         known_schemata: &'n NamesRef<'n>,
     ) -> AvroResult<Self> {
         let names = HashMap::new();
-        let mut rs = ResolvedSchema {
-            names_ref: names,
+        let rs = ResolvedSchema {
+            names_ref: RefCell::new(names),
             schemata: schemata_to_resolve,
         };
         rs.resolve(rs.get_schemata(), enclosing_namespace, Some(known_schemata))?;
```
martin-g, line 77: This line here requires the interior mutability for `names_ref`. There are both mutable and immutable references to `rs`...
I would be happy to revert the changes related to `names_ref` if possible.

PR #445, opened 2026-01-27
feat!: Rework `SpecificSingleObjectWriter`

Kriskras99:
The first commit changes `ResolvedSchema` and `ResolvedOwnedSchema` to use the same logic for resolving. 
`ResolvedOwnedSchema` now also uses a self-referential struct when resolving the schema, so it does not need to clone all named schemas contained in it.

The second commit contains the rework of `SpecificSingleObjectWriter`

It now resolves the schema and caches it. It also no longer uses `GenericSingleObjectWriter` for the header and does it directly. This removes the need for a buffer.

This is a breaking change:
- `SpecificSingleObjectWriter::with_capcity` has been removed.
- `write_avro_datum_ref` now also takes a `names: &NamesRef` argument

This also fixes a bug in `reader::Block::read_writer_schema` where it unnecessarily parses the schemata twice.

Closes #439

### DIFF
--- a/avro/src/writer.rs
+++ b/avro/src/writer.rs
@@ -624,6 +624,12 @@ where
             _model: PhantomData,
         })
     }
+
+    /// Deprecated. Use [`SpecificSingleObjectWriter::new`] instead.
+    #[deprecated(since = "0.22.0", note = "Use new() instead")]
+    pub fn with_capacity(_buffer_cap: usize) -> AvroResult<Self> {
+        Self::new()
+    }
 }
 
 impl<T> SpecificSingleObjectWriter<T>


avro/tests/serde_human_readable_true.rs
```diff
@@ -127,7 +126,7 @@ fn avro_rs_440_uuid_fixed() -> TestResult {
     let mut buffer = Vec::new();
 
     assert!(apache_avro::util::set_serde_human_readable(true));
-    let mut writer = SpecificSingleObjectWriter::with_capacity(64)?;
```
martin-g, line 130: nit: We can keep `with_capacity()` as a delegate to `new()` and deprecated for a version or two.

avro/src/writer.rs
```diff
@@ -617,44 +630,54 @@ impl<T> SpecificSingleObjectWriter<T>
...
+    /// Returns the number of bytes written.
+    ///
+    /// Each call writes a complete single-object encoded message (header + data),
+    /// making each message independently decodable.
+    pub fn write_value<W: Write>(&self, data: T, writer: &mut W) -> AvroResult<usize> {
+        writer
+            .write_all(&self.header)
+            .map_err(Details::WriteBytes)?;
+        let value: Value = data.into();
+        let bytes = write_value_ref_owned_resolved(&self.resolved, &value, writer)?;
```
martin-g, line 650: If this fails then the writer will contain the already written header.
How about something like:
```rust
let mut buffer = Vec::new();
let bytes = write_value_ref_owned_resolved(&self.resolved, &value, &mut buffer)?;
writer.write_all(&self.header).map_err(Details::WriteBytes)?;
writer.write_all(&buffer).map_err(Details::WriteBytes)?;
Ok(bytes + self.header.len())
```

Kriskras99: `writer.write_all(&buffer)` can also fail partway through.

I don't think this is worth the performance penalty.

PR #447, opened 2026-01-27
feat: Add posibility to append values to writer without validating

### LINKED ISSUE #310: Performance: Deserializing Binary Streams
Hi, I am working on reading Avro from BigQuery Storage API, the response from BigQuery creates a raw Avro binary output, the `Reader` implementation doesn't work since it expects the header to be present, so I turned to use `from_avro_datum` interface, however it killed the performance of the app, upon inspecting this, it turned out that this happens because it's resolving the schema multiple times, see the flamegraph below, but basically `ResolvedSchema::try_from` is `80.26%` of the `from_avro_datum` call.

Can we expose another interface for `from_avro_datum` that takes in the `ResolvedSchema` object instead? Or maybe better, create a binary deserializer, in Java this works by using `GenericDatumReader<>(schema)` and creating a decoder with `DecoderFactory.get().binaryDecoder`

Image

Kriskras99:
Closes #310

### DIFF
--- a/avro/README.md
+++ b/avro/README.md
@@ -214,7 +214,7 @@ record.put("a", 27i64);
 record.put("b", "foo");
 
 // schema validation happens here
-writer.append(record).unwrap();
+writer.append_value(record).unwrap();
 
 // this is how to get back the resulting avro bytecode
 // this performs a flush operation to make sure data has been written, so it can fail
@@ -427,7 +427,7 @@ fn main() -> Result<(), Error> {
     record.put("a", 27i64);
     record.put("b", "foo");
 
-    writer.append(record)?;
+    writer.append_value(record)?;
 
     let test = Test {
         a: 27,
@@ -564,7 +564,7 @@ fn main() -> Result<(), Error> {
     record.put("local_timestamp_nanos", Value::LocalTimestampMicros(6));
     record.put("duration", Duration::new(Months::new(6), Days::new(7), Millis::new(8)));
 
-    writer.append(record)?;
+    writer.append_value(record)?;
 
     let input = writer.into_inner()?;
     let reader = Reader::with_schema(&schema, &input[..])?;


avro/src/writer.rs
```diff
@@ -1805,4 +1828,32 @@ mod tests {
...
+    fn avro_rs_310_append_unvalidated_value() -> TestResult {
+        let schema = Schema::String;
+        let value = Value::Int(1);
+
+        let mut writer = Writer::new(&schema, Vec::new())?;
+        writer.unvalidated_append_value_ref(&value)?;
+        writer.unvalidated_append_value(value)?;
+        let buffer = writer.into_inner()?;
+
+        assert_eq!(&buffer[buffer.len() - 18..buffer.len() - 16], &[2, 2]);
```
martin-g, line 1849: Please add a comment about the maths here - what are the magic numbers 18 and 16.
For easier maintenance in the future.

avro/src/writer.rs
```diff
@@ -195,9 +192,58 @@ impl<'a, W: Write> Writer<'a, W> {
...
+            }
+            .into());
+        }
+        self.unvalidated_append_value_ref(value)
+    }
+
+    /// Append a compatible value to a `Writer`.
+    ///
+    /// This function does **not** validate that the provided value matches the schema. If it does
+    /// not match, the file will contain corrupt data. Use [`Writer::append_value`] to have the
```
martin-g, line 219:
```suggestion
    /// not match, the file will contain corrupt data. Use [`Writer::append`] to have the
```
There is no `append_value()`. There are `append()` and `append_value_ref()`.

Should this method be renamed to `unvalidated_append()`, for consistency ?
Or maybe deprecate `append()` and delegate it to a new `append_value()` ?!

Kriskras99: I went with deprecating `append`

Issue #435, opened 2026-01-24
Issue with schema generation using `T::get_schema()`


Theodlz:
Hi! I have a a top-level struct, for which a couple of its fields are structs themselves, of the same type. When using get_schema(), it looks like the result is a schema for that top-level struct, it contains multiple instances of a type definition (generated by these multiple fields of the same struct type), i.e. an invalid schema. Thereafter, the writer fails to serializes (I am using `append_ser`, with the newly added serde API).

In other - perhaps simpler words - I think there is some issue where the schema inferred from structs that derive `AvroSchema` don't do a great job at figuring out that a type is defined multiple times (in which case it should remove all but the first instance of that type definition).

Kriskras99:
@Theodlz your first issue has been fixed in #448. Your second issue is related to the unreliable enum support in this crate, so I'll close this issue and keep tracking it in #365.

Thank you for reporting the issues!

PR #448, opened 2026-01-28
fix: `flatten` no longer causes duplicate names

Kriskras99:
When a type was used both via `flatten` and directly, the schema generated would contain duplicate names (and schemas). This is because `flatten` would use an empty `named_schemas` to get the record schema. If the existing `named_schemas` was used, `flatten` might get a `Schema::Ref` if the type was already used. Or when `flatten` was used first, if the type was used after it would create a `Schema::Ref` to a schema that does not exist.

This is solved by adding a new function to the `AvroSchemaComponent` that returns the fields directly. To not break code currently implementing this trait, it has a default implementation that will work around the issues above. This default implementation is also used for fields with the `#[avro(with = ||)]` and `#[avro(with = path)]` attributes, as they don't have a way to provide the field directly. Users of `#[avro(with)]` will need to implement `get_record_fields_in_ctxt` in their module.

TODO:
 - [x] Create tests for the logic of the default `get_record_fields_in_ctxt` implementation
 - [x] Test interactions with `transparent`
 - [x] Test interactions with `with`
 - [x] Update the documentation in the branch `feat/documentation`
 
 Fixes one of the issues raised in #435


avro/src/serde/derive.rs
```diff
@@ -82,7 +83,134 @@ pub trait AvroSchema {
...
+    schema_fn: fn(named_schemas: &mut Names, enclosing_namespace: &Namespace) -> Schema,
+) -> Option<Vec<RecordField>> {
+    let mut record = match schema_fn(named_schemas, enclosing_namespace) {
+        Schema::Record(record) => record,
+        Schema::Ref { name } => {
+            // This schema already exists in `named_schemas` so temporarily remove it so we can
+            // get the actual schema.
+            let temp = named_schemas
+                .remove(&name)
+                .expect("Name should exist in `named_schemas` otherwise Ref is invalid");
```
```suggestion
                .expect("Name '{name}' should exist in `named_schemas` otherwise Ref is invalid");
```
and maybe also the keys of `named_schemas`

avro/tests/get_record_fields.rs
```diff
@@ -0,0 +1,82 @@
...
+        _a: i32,
+        _b: String,
+    }
+
+    let mut named_schemas = HashMap::new();
+    let fields =
+        get_record_fields_in_ctxt(&mut named_schemas, &None, Foo::get_schema_in_ctxt).unwrap();
+
+    assert_eq!(fields.len(), 2);
+    assert!(named_schemas.is_empty(), "Name shouldn't have been added");
```
martin-g:
```suggestion
    assert!(named_schemas.is_empty(), "Name shouldn't have been added: {named_schemas:?}");
```
for easier debugging.
also do the same below

avro/tests/get_record_fields.rs
```diff
@@ -0,0 +1,82 @@
...
+
+    // Insert Foo into named_schemas
+    let Schema::Record(_) = Foo::get_schema_in_ctxt(&mut named_schemas, &None) else {
```
martin-g: let's use a `match`, so we could print the non-record schema in the panic message

avro_derive/src/lib.rs
```diff
@@ -146,15 +164,13 @@ fn get_struct_schema_def(
...
-                        if let ::apache_avro::schema::Schema::Record(::apache_avro::schema::RecordSchema { fields, .. }) = #flatten_ty::get_schema() {
-                            for mut field in fields {
-                                field.position = schema_fields.len();
-                                schema_fields.push(field)
-                            }
+                        if let Some(flattened_fields) = #get_record_fields {
+                            schema_fields.extend(flattened_fields)
                         } else {
-                            panic!("Can only flatten RecordSchema, got {:?}", #flatten_ty::get_schema())
+                            panic!("#field does not have any fields to flatten to")
```
```suggestion
                            panic!("{} does not have any fields to flatten to", stringify!(#field))
```

avro_derive/src/lib.rs
```diff
@@ -146,15 +164,13 @@ fn get_struct_schema_def(
...
+                    let get_record_fields =
+                        get_field_get_record_fields_expr(&field, field_attrs.with)?;
                     record_field_exprs.push(quote! {
-                        if let ::apache_avro::schema::Schema::Record(::apache_avro::schema::RecordSchema { fields, .. }) = #flatten_ty::get_schema() {
-                            for mut field in fields {
-                                field.position = schema_fields.len();
-                                schema_fields.push(field)
-                            }
+                        if let Some(flattened_fields) = #get_record_fields {
+                            schema_fields.extend(flattened_fields)
```
martin-g: Here the `RecordField::position` will have to be re-calculated for the flattened fields.
https://github.com/apache/avro-rs/blob/5a553beac6b1858e6f97d110318216cf6ecdb4c6/avro/src/schema/record/field.rs#L55

avro_derive/src/lib.rs
```diff
@@ -302,6 +328,41 @@ fn get_field_schema_expr(field: &Field, with: With) -> Result<TokenStream, Vec<s
...
 }
 
+fn get_field_get_record_fields_expr(
+    field: &Field,
+    with: With,
+) -> Result<TokenStream, Vec<syn::Error>> {
+    match with {
+        With::Trait => Ok(type_to_get_record_fields_expr(&field.ty)?),
+        With::Serde(path) => {
+            Ok(quote! { #path::get_record_fields_in_ctxt(named_schemas, enclosing_namespace) })
```
martin-g: What happens if there is no such function in the module ?
I guess the compilation fails.

Maybe we should add no-op impls (i.e. returning None) for the `with` impls like:
https://github.com/apache/avro-rs/blob/5a553beac6b1858e6f97d110318216cf6ecdb4c6/avro/src/serde/with.rs#L144

avro/src/serde/derive.rs
```diff
@@ -177,50 +329,9 @@ where
             UnionSchema::new(variants).expect("Option<T> must produce a valid (non-nested) union"),
         )
     }
-}
-
-impl<T> AvroSchemaComponent for Map<String, T>
```
martin-g, line 182: This impl is lost.
I don't remember any user asking for it. I am not sure whether it is really needed or not.

avro/src/serde/derive.rs
```diff
@@ -82,7 +83,134 @@ pub trait AvroSchema {
...
+                name: record.name,
+                aliases: record.aliases,
+                doc: record.doc,
+                fields: new_fields,
+                lookup: record.lookup,
+                attributes: record.attributes,
+            };
+
+            let Schema::Ref { name } = std::mem::replace(schema, Schema::Record(new_schema)) else {
+                panic!("Expected only Refs from find_first_ref");
```
martin-g: Let's use a `match` to be able to print the non-Ref schema in case of panic.

avro/src/serde/derive.rs
Kriskras99: `expect` only takes `&str`. I could replace it with `unwrap_or_else(|| panic!(...))`, what do you think?

avro/src/serde/derive.rs
Kriskras99: Forgot to add that to the PR description, here is the reason:
The implementation for `serde_json::Map<String, T>` has been removed. This is **not** a breaking change, as `serde_json::Map` can only be constructed for `<String, Value>` and `Value` does not implement `AvroSchemaComponent`.

avro/src/serde/derive.rs
martin-g: Good!

avro_derive/src/lib.rs
Kriskras99: I have fixed it in the flatten code. An alternative is adding an extra field to `get_record_fields_in_ctxt` that gives the starting position, I think that might be a better option?

avro_derive/src/lib.rs
Kriskras99: I've added the implementations. I really want to avoid people relying on the default implementation, as it is really slow for records, so a compiler error would be better.

avro_derive/src/lib.rs
Kriskras99: I've added my suggestion in the last commit, I think it's cleaner overall

avro/src/serde/with.rs
```diff
@@ -202,18 +212,18 @@ pub mod fixed {
...
-        schema::{FixedSchema, Name, Names, Namespace},
+        schema::{FixedSchema, Name, Names, Namespace, RecordField},
     };
 
     /// Returns `Schema::Fixed(N)` named `serde_avro_fixed_{N}`
     #[expect(clippy::map_entry, reason = "We don't use the value from the map")]
     pub fn get_schema_in_ctxt<const N: usize>(
         named_schemas: &mut Names,
-        enclosing_namespace: &Namespace,
+        enclosing_pace: &Namespace,
```
martin-g: Why ?

avro_derive/src/lib.rs
```diff
@@ -81,14 +94,19 @@ fn derive_avro_schema(input: DeriveInput) -> Result<TokenStream, Vec<syn::Error>
...
     let (impl_generics, ty_generics, where_clause) = generics.split_for_impl();
     quote! {
         #[automatically_derived]
         impl #impl_generics apache_avro::AvroSchemaComponent for #ident #ty_generics #where_clause {
             fn get_schema_in_ctxt(named_schemas: &mut apache_avro::schema::Names, enclosing_namespace: &Option<String>) -> apache_avro::schema::Schema {
-                #implementation
+                #get_schema_impl
+            }
+
+            fn get_record_fields_in_ctxt(mut field_position: usize, named_schemas: &mut apache_avro::schema::Names, enclosing_namespace: &Option<String>) -> Option<std::vec::Vec<apache_avro::schema::RecordField>> {
```
```suggestion
            fn get_record_fields_in_ctxt(mut field_position: usize, named_schemas: &mut apache_avro::schema::Names, enclosing_namespace: &Option<String>) -> Option<::std::vec::Vec<::apache_avro::schema::RecordField>> {
```
I think we need to add such "root" prefix for all types which are used inside `quote!( ... )`.
In a follow-up PR.

avro/src/serde/with.rs
Kriskras99: Ctrrl+F `"_name"` Ctrl+R `"_"`, whoops

avro_derive/src/lib.rs
Kriskras99: That's a good idea

Issue #449, opened 2026-01-31
Value to JSON serialisation is wrong for Unions that are not Null


PookieBuns:
**Edit: I noticed this is somewhat related to #365 so please let me know if I should move my discussion there. Thanks!**

Hi maintainers! First of all, I would love to thank you for your work bringing apache avro into the Rust ecosystem. I am truly grateful. I have been trying to adopt avro more at work, but in its current state, it is quite brittle, especially with respect to using serde with Unions that contain a null variant. I plan on contributing to fix these issues, but before then I would like some input from the maintainers with respect to what we would like to support wrt this crate.

# TLDR
- There are arguably 3 ways to represent avro nullable unions in Rust, some work and some don't, but I'd like input on which ones we'd like to support or if we should take a stance on a single correct way.
- For nullable unions with >3 variants, the rusty way requires `#[serde(untagged)]` to work but it fails on multiple edge cases. Would like input on which plan to address is better direction for this crate

# The Rust Representation Conundrum

Given an schema of 

```
    {
        "name": "MyUnion",
        "type": [
            "null",
            "int"
        ]
    }
```

There are currently 3 arguable ways to represent this with Rust types

## The current `avro-rs` way

```
enum MyUnionNullable {
    Null,
    Int(i32),
}
```

### Advantages

This aligns closest to how avro binary encoding works and is simplest to serialize. Use the variant index from serde to properly encode the branch index then encode the contained value

### Disadvantages

This is not "rusty", as null or missing values are generally represented as `Option<T>`, and by doing this you lose the nice semantics of `Option`.

This also isn't the proper avro json encoding if you choose to serialize it with `serde_json`. 

According to the avro specification,

> The value of a union is encoded in JSON as follows:
>
> - if its type is null, then it is encoded as a JSON null;
> - otherwise it is encoded as a JSON object with one name/value pair whose
  name is the type’s name and whose value is the recursively encoded value.
  For Avro’s named types (record, fixed or enum) the user-specified name is
  used, for other types the type name is used.

Which means that the null branch should be encoded as `null` and the int branch should be encoded as `{ "int": 1 }`, but in this representation the null branch is encoded as `"Null"`.

## The "Rusty" way

```
Option<i32>
```

### Advantages

This is the most intuitive for Rust users

### Disadvantages

This also isn't the proper avro json encoding if you choose to serialize it with `serde_json` because the int branch gets encoded as `1` without the discriminator

## The Anchor On Avro Json Encoding And Serde Json Way

```
[...]

PR #458, opened 2026-02-10
feat: Support (De)Serialization for different representations of Nullable Unions

PookieBuns:
Fixes `The Rust Representation Conundrum` as mentioned in #449, which allows (de)serializing, all `avro-rs`, `rusty`, and `avro_json_encoding_compatible` rust representations to avro. Tested using the provided test suite in the mentioned issue.

**Note: This does not address #365 since this only applies to schema aware serialization.**

I considered using enum variant name as lookup method for picking Union branch when serializing, shown [here](https://github.com/PookieBuns/avro-rs/commit/74172ecf3916ec1c82a5dcbca2c8caa0f8c943a7), but given the current state in which `Value` doesn't hold the name for named types, I realized it's not possible to do the same for deserialization until further work to support names in `Value` has been done.

Therefore, this implementation makes the assumption that if you are using `Option<T>` to represent a nullable union, that means that the `null` is the first variant in the union and serialization / deserialization just bump / decrement the index lookup respectively.

I believe this doesn't actually produce any breaking changes, because the status quo is that tagged Option "Rusty" representations straight up don't work.

Please let me know if you have any questions!

### DIFF
--- a/avro/src/serde/ser_schema/mod.rs
+++ b/avro/src/serde/ser_schema/mod.rs
@@ -1483,7 +1483,7 @@ impl<'s, W: Write> SchemaAwareWriteSerializer<'s, W> {
                 if self.serializing_some {
                     for (i, variant_schema) in union_schema.schemas.iter().enumerate() {
                         match variant_schema {
-                            Schema::Enum(enum_schema) if enum_schema.name.name == name => {
+                            Schema::Enum(enum_schema) if enum_schema.name.name() == name => {
                                 if variant_index as usize >= enum_schema.symbols.len() {
                                     return Err(create_error(format!(
                                         "Variant index out of bounds: {}. The Enum schema has '{}' symbols",
@@ -1497,7 +1497,7 @@ impl<'s, W: Write> SchemaAwareWriteSerializer<'s, W> {
                             Schema::Ref { name: ref_name } => {
                                 let ref_schema = self.get_ref_schema(ref_name)?;
                                 if let Schema::Enum(enum_schema) = ref_schema
-                                    && enum_schema.name.name == name
+                                    && enum_schema.name.name() == name
                                 {
                                     if variant_index as usize >= enum_schema.symbols.len() {
                                         return Err(create_error(format!(
@@ -3805,7 +3805,10 @@ mod tests {
         )?;
         writer.append_ser(&input)?;
         let bytes = writer.into_inner()?;
-        let mut reader = Reader::with_schemata(&schemas[1], schemas.iter().collect(), &bytes[..])?;
+        let mut reader = Reader::builder(&bytes[..])
+            .reader_schema(&schemas[1])
+            .schemata(schemas.iter().collect())
+            .build()?;
         let read_avro_value = reader.next().unwrap()?;
         assert_eq!(
             &read_avro_value, &expected,
--- a/avro/tests/nullable_union.rs
+++ b/avro/tests/nullable_union.rs
@@ -32,7 +32,9 @@ where
     let mut writer = Writer::new(test_case.schema, Vec::new())?;
     writer.append_ser(&test_case.input)?;
     let bytes = writer.into_inner()?;
-    let mut reader = Reader::with_schema(test_case.schema, &bytes[..])?;
+    let mut reader = Reader::builder(&bytes[..])
+        .reader_schema(test_case.schema)
+        .build()?;
     let read_avro_value = reader.next().unwrap()?;
     assert_eq!(
         &read_avro_value, test_case.expected_avro,


avro/src/serde/de.rs
```diff
@@ -638,6 +664,21 @@ impl<'de> de::Deserializer<'de> for Deserializer<'de> {
...
+                if matches!(x.deref(), Value::Union(_, _)) {
+                    Err(de::Error::custom(format!(
+                        "Directly nested union types are not supported. Got Value::Union({i}, {x:?})"
+                    )))
+                } else {
+                    Self::new(x.deref())
+                        .deserialize_byte_buf(visitor)
+                        .map_err(|e| {
+                            de::Error::custom(format!(
+                                "Attempted to deserialize Value::Union({i}, {x:?}) as bytes: {e:?}"
```
martin-g:
```suggestion
                                "Attempted to deserialize Value::Union({i}, {x:?}) as byte_buf: {e:?}"
```

avro/src/serde/de.rs
```diff
@@ -799,9 +850,19 @@ impl<'de> de::Deserializer<'de> for Deserializer<'de> {
...
-                if (*idx as usize) < variants.len() {
+                // if we came here from a some, we need to check if we are deserializing a
+                // non-newtype enum
+                if self.deserializing_some
+                    && let Value::Enum(_index, field) = inner.deref()
+                    && variants.contains(&&**field)
+                {
+                    return visitor.visit_enum(EnumUnitDeserializer::new(field));
+                }
+                // Assume `null` is the first branch if deserializing some so decrement the variant index
```
martin-g, line 861: This assumption is not guaranteed by the spec :-/
We should add some index/length checks to avoid panics/overflows.

avro/src/serde/ser_schema/mod.rs
```diff
@@ -1460,19 +1477,37 @@ impl<'s, W: Write> SchemaAwareWriteSerializer<'s, W> {
...
+                                    )));
+                                }
+                                encode_int(i as i32, &mut self.writer)?;
+                                return encode_int(variant_index as i32, &mut self.writer);
+                            }
+                            _ => { /* skip */ }
+                        }
+                    }
+                }
+                let branch_index = variant_index as usize + usize::from(self.serializing_some);
```
martin-g, line 1517: Again, this assumes that the `null` is the first variant in the union.

avro/tests/nullable_union.rs
```diff
@@ -0,0 +1,982 @@
...
+
+mod nullable_primitive_int {
+    use super::*;
+
+    const NULLABLE_INT_SCHEMA: &str = r#"
+    {
+        "name": "MyUnion",
+        "type": [
+            "null",
+            "int"
```
martin-g: Please add a test where the `null` is not the first variant too.

avro/src/serde/ser_schema.rs
```diff
@@ -1524,17 +1559,21 @@ impl<'s, W: Write> SchemaAwareWriteSerializer<'s, W> {
...
-                    .ok_or_else(|| {
-                        create_error(format!(
-                            "No variant schema at position {variant_index} for {union_schema:?}"
-                        ))
-                    })?;
+                let branch_index = variant_index as usize + usize::from(self.serializing_some);
+                if branch_index >= union_schema.schemas.len() {
+                    return Err(create_error(format!(
+                        "Variant index out of bounds: {}. The union schema has '{}' schemas",
+                        variant_index,
```
```suggestion
                        branch_index,
```

avro/src/serde/ser_schema.rs
```diff
@@ -1703,17 +1742,21 @@ impl<'s, W: Write> SchemaAwareWriteSerializer<'s, W> {
...
-                    .ok_or_else(|| {
-                        create_error(format!(
-                            "Cannot find a variant at position {variant_index} in {union_schema:?}"
-                        ))
-                    })?;
+                let branch_index = variant_index as usize + usize::from(self.serializing_some);
+                if branch_index >= union_schema.schemas.len() {
+                    return Err(create_error(format!(
+                        "Variant index out of bounds: {}. The union schema has '{}' schemas",
+                        variant_index,
```
```suggestion
                        branch_index,
```

avro/src/serde/ser_schema.rs
```diff
@@ -1835,17 +1878,17 @@ impl<'s, W: Write> SchemaAwareWriteSerializer<'s, W> {
...
-                    .ok_or_else(|| {
-                        create_error(format!(
-                            "Cannot find variant at position {variant_index} in {union_schema:?}"
-                        ))
-                    })?;
+                let branch_index = variant_index as usize + usize::from(self.serializing_some);
+                if branch_index >= union_schema.schemas.len() {
+                    return Err(create_error(format!(
+                        "Variant index out of bounds: {}. The union schema has '{}' schemas",
+                        variant_index,
```
```suggestion
                        branch_index,
```

avro/src/serde/ser_schema/mod.rs
```diff
@@ -1460,19 +1477,37 @@ impl<'s, W: Write> SchemaAwareWriteSerializer<'s, W> {
...
+                                if variant_index as usize >= enum_schema.symbols.len() {
+                                    return Err(create_error(format!(
+                                        "Variant index out of bounds: {}. The Enum schema has '{}' symbols",
+                                        variant_index,
+                                        enum_schema.symbols.len()
+                                    )));
+                                }
+                                encode_int(i as i32, &mut self.writer)?;
+                                return encode_int(variant_index as i32, &mut self.writer);
+                            }
```
martin-g, line 1496: This should probably support Schema::Ref too, to refer to Schema::Enum

avro/src/serde/de.rs
```diff
@@ -75,7 +76,17 @@ struct UnionDeserializer<'de> {
...
 impl<'de> Deserializer<'de> {
     pub fn new(input: &'de Value) -> Self {
-        Deserializer { input }
+        Deserializer {
+            input,
+            deserializing_some: false,
+        }
+    }
+
+    pub fn new_deserializing_some(input: &'de Value) -> Self {
```
martin-g:
```suggestion
    fn new_deserializing_some(input: &'de Value) -> Self {
```

avro/src/serde/ser_schema.rs
```diff
@@ -629,6 +630,22 @@ impl<'s, W: Write> SchemaAwareWriteSerializer<'s, W> {
             root_schema: schema,
             names,
             enclosing_namespace,
+            serializing_some: false,
+        }
+    }
+
+    pub fn new_serializing_some(
```
```suggestion
    fn new_serializing_some(
```

avro/tests/nullable_union.rs
```diff
@@ -0,0 +1,982 @@
...
+    fn serialize_rusty_untagged_my_record_b_27() -> AvroResult<()> {
+        test_serialize(TestCase {
+            input: Some(MyUnionUntagged::MyRecordB(MyRecordB { a: 27 })),
+            schema: &schema(),
+            expected_avro: &record_b_variant_expected_avro(27),
+        })
+    }
+
+    #[test]
+    #[ignore]
```
martin-g: Please add a comment why it is ignored

PookieBuns: Updated to instead expect a failure / comment on why it will fail. I'm also open to just removing it if you think its out of scope, since that mod is mostly to demonstrate how untagged enums have edge cases that fail and check that tagged enums do not

avro/src/serde/de.rs
PookieBuns: Updated to use checked_sub to ensure underflow doesn't happen.

Regarding the topic about not being enforced by spec, even though it isn't strictly enforced, it's highly recommended https://avro.apache.org/docs/1.11.1/specification/#unions

> Note that when a default value is specified for a record field whose type is a union, the type of the default value must match the first element of the union. Thus, for unions containing “null”, the “null” is usually listed first, since the default value of such unions is typically null.

For serialization we technically could get around this by resolving union branches using variant name, but for deserialization cannot yet because it's not schema aware and/or Value doesn't hold name information, which is why I chose to use index bumping on serialization too for consistent behavior. This seems like the closest we can get without a significant refactor or feature for deserialization.

For now, the behavior will intentionally throw an error if you try to (de)serialize an Option<T> with null not being the first branch. This doesn't break existing behavior because the status quo is non-functional so although not a perfect solution imo it's a step better than what we have now

avro/src/serde/ser_schema/mod.rs
PookieBuns: Updated to throw error if null isn't first branch, reasoning stated in https://github.com/apache/avro-rs/pull/458#discussion_r2789323892

avro/tests/nullable_union.rs
PookieBuns: Added tests in https://github.com/apache/avro-rs/pull/458/changes/6f691dbbcf2211eb5c112cda90759f7f76c0d44c to ensure null as non-first variant when using Option<T> fails with error message communicating intended behavior

avro/src/serde/ser_schema.rs
PookieBuns: Updated!

avro/src/serde/ser_schema.rs
PookieBuns: Updated!

avro/src/serde/ser_schema/mod.rs
PookieBuns: updated in https://github.com/apache/avro-rs/pull/458/changes/6c58e62db077ad87efb8549e4200a20f5298b465 with test to verify behavior

avro/src/serde/de.rs
```diff
@@ -799,14 +850,26 @@ impl<'de> de::Deserializer<'de> for Deserializer<'de> {
...
+                if self.deserializing_some
+                    && let Value::Enum(_index, field) = inner.deref()
+                    && variants.contains(&&**field)
+                {
+                    return visitor.visit_enum(EnumUnitDeserializer::new(field));
+                }
+                // Assume `null` is the first branch if deserializing some so decrement the variant index
+                let variant_idx = (*idx as usize)
+                    .checked_sub(usize::from(self.deserializing_some))
+                    .ok_or_else(|| Details::GetNull(inner.deref().clone()))?;
```
PookieBuns, line 864: Do we want to add an Error variant here to communicate more clearly that deserializing some when null isn't first isn't supported? Right now it just tells the user Expects null for 0th when deserializing some

PookieBuns:
Updated to use new methods introduced by previous commits

PR #459, opened 2026-02-11
feat: Suggest `boolean` when `bool` is used

### LINKED ISSUE #456: Default parsing does not work for boolean
```rust
#[test]
fn avro_rs_xxx_schema_defaults() -> TestResult {
    let schema = Schema::parse_str(
        r#"{
        "type": "record",
        "name": "defaults",
        "fields": [
            {"name": "boolean", "type": "bool", "default": true}
        ]
    }"#)?;
    Ok(())
}
```
This fails with:
```
      apache_avro::error::Error: Error { details: Unknown primitive type: bool }
    in avro/src/schema/mod.rs, line 5134
    thread: schema::tests::avro_rs_xxx_schema_defaults
```

Kriskras99:
Closes #456

### DIFF
--- a/avro/src/error.rs
+++ b/avro/src/error.rs
@@ -358,7 +358,7 @@ pub enum Details {
     #[error("Unknown primitive type: {0}")]
     ParsePrimitive(String),
 
-    #[error("Unknown primitive type: {0}, did you mean {1}?")]
+    #[error("Unknown primitive type: '{0}'. Did you mean '{1}' ?")]
     ParsePrimitiveSimilar(String, &'static str),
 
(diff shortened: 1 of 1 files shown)


avro/src/error.rs
```diff
@@ -358,6 +358,9 @@ pub enum Details {
     #[error("Unknown primitive type: {0}")]
     ParsePrimitive(String),
 
+    #[error("Unknown primitive type: {0}, did you mean {1}?")]
```
```suggestion
    #[error("Unknown primitive type: '{0}'. Did you mean '{1}' ?")]
```

PR #460, opened 2026-02-11
fix: Enum and Fixed would put `default` in `custom_attributes`

Kriskras99:


avro/src/schema/mod.rs
```diff
@@ -5147,6 +5158,58 @@ mod tests {
             error,
             Details::ParsePrimitiveSimilar("bool".to_string(), "boolean").to_string()
         );
+
+        Ok(())
+    }
+
+    #[test]
+    fn avro_rs_460_fixed_default_in_custom_attributes() -> TestResult {
```
martin-g, line 5128: Why the name of the method is `in_custom_attributes` ?
`ssert_eq!(schema.custom_attributes().unwrap(), &BTreeMap::new());` verifies that it is not there

avro/src/schema/mod.rs
```diff
@@ -5147,6 +5158,58 @@ mod tests {
...
+        }"#,
+        )?;
+
+        assert_eq!(schema.custom_attributes().unwrap(), &BTreeMap::new());
+
+        let json = serde_json::to_string(&schema)?;
+        let schema2 = Schema::parse_str(&json)?;
+
+        let Schema::Enum(enum_schema) = schema2 else {
+            panic!("Expected Schema::Fixed, got {schema2:?}");
```
```suggestion
            panic!("Expected Schema::Enum, got {schema2:?}");
```

avro/src/schema/mod.rs
Kriskras99: That name was missing a `not`, but is now actually fitting.

avro/src/schema/mod.rs
```diff
@@ -5147,6 +5120,55 @@ mod tests {
...
+        assert_eq!(schema2.custom_attributes().unwrap().len(), 1);
+
+        Ok(())
+    }
+
+    #[test]
+    fn avro_rs_460_enum_default_in_custom_attributes() -> TestResult {
+        let schema = Schema::parse_str(
+            r#"{
+            "name": "fixed_with_default",
```
```suggestion
            "name": "enum_with_default",
```

PR #467, opened 2026-02-16
feat!: Support `default` field for array and map

Kriskras99:
One thing I did notice in the spec is that the text says `support a single attribute` while the example does show a `default` field, so the text of the spec should be updated as well.

### DIFF
--- a/avro/src/error.rs
+++ b/avro/src/error.rs
@@ -312,6 +312,9 @@ pub enum Details {
     #[error("`default`'s value type of field {0:?} in {1:?} must be {2:?}")]
     GetDefaultRecordField(String, String, String),
 
+    #[error("JSON number {0} could not be converted into an Avro value as it's too large")]
+    JsonNumberTooLarge(serde_json::Number),
+
     #[error("JSON value {0} claims to be u64 but cannot be converted")]
     GetU64FromJson(serde_json::Number),
 
@@ -342,6 +345,9 @@ pub enum Details {
 
     #[error("Cannot convert i32 to u128: {1}")]
     ConvertI32ToU128(#[source] std::num::TryFromIntError, i32),
+
+    #[error("Cannot convert usize to i64: {1}")]
+    ConvertUsizeToI64(#[source] std::num::TryFromIntError, usize),
 
     #[error("Invalid JSON value for decimal precision/scale integer: {0}")]
     GetPrecisionOrScaleFromJson(serde_json::Number),
--- a/avro/src/schema/mod.rs
+++ b/avro/src/schema/mod.rs
@@ -4729,7 +4729,7 @@ mod tests {
                 assert_eq!(field.name, "birthday");
                 assert_eq!(field.schema, Schema::Date);
                 assert_eq!(
-                    types::Value::from(field.default.clone().unwrap()),
+                    types::Value::try_from(field.default.clone().unwrap())?,
                     types::Value::Int(1681601653)
                 );
             }
--- a/avro/src/schema/parser.rs
+++ b/avro/src/schema/parser.rs
@@ -710,7 +710,8 @@ impl Parser {
             .and_then(|items| self.parse(items, enclosing_namespace))?;
         let default = if let Some(default) = complex.get("default").cloned() {
             if let Value::Array(_) = default {
-                let crate::types::Value::Array(array) = crate::types::Value::from(default) else {
+                let crate::types::Value::Array(array) = crate::types::Value::try_from(default)?
+                else {
                     unreachable!("JsonValue::Array can only become a Value::Array")
                 };
                 // Check that the default type matches the schema type
@@ -747,7 +748,7 @@ impl Parser {
 
         let default = if let Some(default) = complex.get("default").cloned() {
             if let Value::Object(_) = default {
-                let crate::types::Value::Map(map) = crate::types::Value::from(default) else {
+                let crate::types::Value::Map(map) = crate::types::Value::try_from(default)? else {
                     unreachable!("JsonValue::Object can only become a Value::Map")
                 };
                 // Check that the default type matches the schema type
--- a/avro/src/schema/record/field.rs
+++ b/avro/src/schema/record/field.rs
@@ -133,7 +133,7 @@ impl RecordField {
         default: &Option<Value>,
     ) -> AvroResult<()> {
         if let Some(value) = default {
-            let avro_value = types::Value::from(value.clone());
+            let avro_value = types::Value::try_from(value.clone())?;
             match field_schema {
                 Schema::Union(union_schema) => {
                     let schemas = &union_schema.schemas;
--- a/avro/src/types.rs
+++ b/avro/src/types.rs
@@ -156,11 +156,13 @@ impl From<()> for Value {
     }
 }
 
-impl From<usize> for Value {
-    fn from(value: usize) -> Self {
-        i64::try_from(value)
-            .expect("cannot convert usize to i64")
-            .into()
+impl TryFrom<usize> for Value {
+    type Error = Error;
+
+    fn try_from(value: usize) -> Result<Self, Self::Error> {
+        Ok(i64::try_from(value)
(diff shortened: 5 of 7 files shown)


avro/src/error.rs
```diff
@@ -431,9 +431,21 @@ pub enum Details {
...
+    #[error("Default value for an array must be an array! Got: {0}")]
+    ArrayDefaultWrongType(serde_json::Value),
+
+    #[error("Default value for an array must be an array of {0}! Found: {1:?}")]
+    ArrayDefaultWrongInnerType(Schema, Value),
+
+    #[error("Default value for a map must be a object! Got: {0}")]
+    MapDefaultWrongType(serde_json::Value),
+
+    #[error("Default value for a map must be a object with (String, {0})! Found: (String, {1:?})")]
```
```suggestion
    #[error("Default value for a map must be an object with (String, {0})! Found: (String, {1:?})")]
```

avro/src/error.rs
```diff
@@ -431,9 +431,21 @@ pub enum Details {
...
+    #[error("Default value for an enum must be a string! Got: {0}")]
     EnumDefaultWrongType(serde_json::Value),
 
+    #[error("Default value for an array must be an array! Got: {0}")]
+    ArrayDefaultWrongType(serde_json::Value),
+
+    #[error("Default value for an array must be an array of {0}! Found: {1:?}")]
+    ArrayDefaultWrongInnerType(Schema, Value),
+
+    #[error("Default value for a map must be a object! Got: {0}")]
```
```suggestion
    #[error("Default value for a map must be an object! Got: {0}")]
```

avro/src/schema/mod.rs
```diff
@@ -3952,19 +3982,17 @@ mod tests {
...
+
         let result = Schema::parse_str(schema_str);
         assert!(result.is_err());
         let err = result
             .map_err(|e| e.to_string())
             .err()
             .unwrap_or_else(|| "unexpected".to_string());
-        assert_eq!(expected, err);
+        assert_eq!(
+            r#"Default value for a map must be a object! Got: "invalid""#,
```
```suggestion
            r#"Default value for a map must be an object! Got: "invalid""#,
```

avro/src/schema/mod.rs
```diff
@@ -5171,4 +5209,180 @@ mod tests {
...
+            "type": "map",
+            "values": "string",
+            "default": {"foo": true}
+        }"#,
+        )
+        .unwrap_err();
+
+        assert_eq!(
+            err.to_string(),
+            "Default value for a map must be a object with (String, String)! Found: (String, Boolean(true))"
```
```suggestion
            "Default value for a map must be an object with (String, String)! Found: (String, Boolean(true))"
```

avro/src/schema/mod.rs
```diff
@@ -5171,4 +5209,180 @@ mod tests {
...
+            "type": "map",
+            "values": "string",
+            "default": {"foo": "bar", "spam": true}
+        }"#,
+        )
+        .unwrap_err();
+
+        assert_eq!(
+            err.to_string(),
+            "Default value for a map must be a object with (String, String)! Found: (String, Boolean(true))"
```
```suggestion
            "Default value for a map must be an object with (String, String)! Found: (String, Boolean(true))"
```

avro/src/schema/mod.rs
```diff
@@ -5171,4 +5209,180 @@ mod tests {
...
+
+        let Schema::Array(array) = schema else {
+            panic!("Expected Schema::Array, got {schema:?}");
+        };
+
+        assert_eq!(array.attributes, BTreeMap::new());
+        assert_eq!(array.default, Some(Vec::new()));
+
+        let json = serde_json::to_string(&Schema::Array(array))?;
+        json.contains(r#""default":[]"#);
```
```suggestion
        assert!(json.contains(r#""default":[]"#));
```

avro/src/schema/parser.rs
```diff
@@ -704,16 +704,31 @@ impl Parser {
...
-                )
-            })
+            .and_then(|items| self.parse(items, enclosing_namespace))?;
+        let default = if let Some(default) = complex.get("default").cloned() {
+            if let Value::Array(_) = default {
+                let crate::types::Value::Array(array) = crate::types::Value::from(default) else {
+                    unreachable!("JsonValue::Array can only become a Value::Array")
+                };
+                // Check that the default type matches the schema type
+                if let Some(value) = array.iter().find(|v| !v.validate(&items)) {
```
martin-g: If the `items` contain Schema::Ref then you will need to pass the `schemata`, i.e. use `Value::validate_internal()`

avro/src/schema/parser.rs
```diff
@@ -722,16 +737,33 @@ impl Parser {
...
-            })
+            .and_then(|types| self.parse(types, enclosing_namespace))?;
+
+        let default = if let Some(default) = complex.get("default").cloned() {
+            if let Value::Object(_) = default {
+                let crate::types::Value::Map(map) = crate::types::Value::from(default) else {
+                    unreachable!("JsonValue::Object can only become a Value::Map")
+                };
+                // Check that the default type matches the schema type
+                if let Some(value) = map.values().find(|v| !v.validate(&types)) {
```
martin-g: If the `values` contain Schema::Ref then you will need to pass the `schemata`, i.e. use `Value::validate_internal()`

avro/src/schema/parser.rs
```diff
@@ -722,16 +737,33 @@ impl Parser {
...
-                Schema::map_with_attributes(
-                    items,
-                    self.get_custom_attributes(complex, vec!["values"]),
-                )
-            })
+            .and_then(|types| self.parse(types, enclosing_namespace))?;
+
+        let default = if let Some(default) = complex.get("default").cloned() {
+            if let Value::Object(_) = default {
+                let crate::types::Value::Map(map) = crate::types::Value::from(default) else {
```
martin-g: Not caused by this PR:
This could be exploited to trigger the panic at https://github.com/apache/avro-rs/blob/af56163614c4b45fb0d8cfaec997ed07a75e34ec/avro/src/types.rs#L286
I think we should get rid of this panic!() and log an error/warning + return `JsonValue::Double(f64::MAX)` (a best effort).

avro/src/schema/mod.rs
```diff
@@ -5171,4 +5209,180 @@ mod tests {
...
+        assert_eq!(
+            array.default,
+            Some(vec![
+                Value::String("foo".into()),
+                Value::String("bar".into())
+            ])
+        );
+
+        let json = serde_json::to_string(&Schema::Array(array))?;
+        json.contains(r#""default":["foo","bar"]"#);
```
```suggestion
        assert!(json.contains(r#""default":["foo","bar"]"#));
```

avro/src/schema/mod.rs
```diff
@@ -5171,4 +5209,180 @@ mod tests {
...
+
+        let Schema::Map(map) = schema else {
+            panic!("Expected Schema::Map, got {schema:?}");
+        };
+
+        assert_eq!(map.attributes, BTreeMap::new());
+        assert_eq!(map.default, Some(HashMap::new()));
+
+        let json = serde_json::to_string(&Schema::Map(map))?;
+        json.contains(r#""default":{}"#);
```
```suggestion
        assert!(json.contains(r#""default":{}"#));
```

avro/src/schema/mod.rs
```diff
@@ -5171,4 +5209,180 @@ mod tests {
...
+            panic!("Expected Schema::Map, got {schema:?}");
+        };
+
+        let mut hashmap = HashMap::new();
+        hashmap.insert("foo".to_string(), Value::String("bar".into()));
+        assert_eq!(map.attributes, BTreeMap::new());
+        assert_eq!(map.default, Some(hashmap));
+
+        let json = serde_json::to_string(&Schema::Map(map))?;
+        json.contains(r#""default":{"foo":"bar"}"#);
```
```suggestion
        assert!(json.contains(r#""default":{"foo":"bar"}"#));
```

avro/src/schema/parser.rs
Kriskras99: Or remove the `From` implementation and make it a `TryFrom`, as `From` is for lossless conversions which cannot panic.

martin-g: Agreed!

Kriskras99: Fixed!

PR #469, opened 2026-02-18
feat: Add support for resetting a Writer back by a clearable buffer

Kriskras99:
This means the schema doesn't have to be resolved every time a new "file" is started.

I created the `Clearable` trait so users can also use their own buffer types (like [`Bytes`](https://docs.rs/bytes/1.11.1/bytes/struct.Bytes.html)).

### DIFF
--- a/avro/src/writer.rs
+++ b/avro/src/writer.rs
@@ -557,8 +557,8 @@ impl Clearable for Vec<u8> {
 impl<'a, W: Clearable + Write> Writer<'a, W> {
     /// Reset the writer.
     ///
-    /// This will clear the underlying writer and the internal buffer.
-    /// It will also clear any user metadata added.
+    /// This will clear the underlying writer, the internal buffer, and the user metadata.
+    /// It will also generate a new sync marker.
     ///
     /// # Example
     /// ```
@@ -598,6 +598,7 @@ impl<'a, W: Clearable + Write> Writer<'a, W> {
         self.has_header = false;
         self.num_values = 0;
         self.user_metadata.clear();
+        self.marker = generate_sync_marker();
     }
 }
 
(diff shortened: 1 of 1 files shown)


avro/src/writer.rs
```diff
@@ -542,6 +542,60 @@ impl<'a, W: Write> Writer<'a, W> {
...
 
+/// A buffer that can be cleared.
+pub trait Clearable {
+    /// Clear the buffer, keeping the capacity.
+    fn clear(&mut self);
+}
+
+impl Clearable for Vec<u8> {
+    fn clear(&mut self) {
+        self.clear();
```
martin-g: This works but it looks like a recursion. Maybe:

```suggestion
        Vec::clear(self);
```
to make it more explicit.

avro/src/writer.rs
```diff
@@ -542,6 +542,60 @@ impl<'a, W: Write> Writer<'a, W> {
...
+    /// // Flush the buffer and only then do something with buffer
+    /// writer.flush()?;
+    /// send(writer.get_ref());
+    ///
+    /// // Reset the writer
+    /// writer.reset();
+    ///
+    /// // Write some values again
+    /// for value in values {
+    ///     writer.append(value)?;
```
```suggestion
    ///     writer.append_value(value)?;
```

avro/src/writer.rs
```diff
@@ -542,6 +542,60 @@ impl<'a, W: Write> Writer<'a, W> {
...
+    /// }
+    ///
+    /// # Ok::<(), Error>(())
+    /// ```
```
martin-g, line 589:
```suggestion
    /// ```
    ///
    /// # Warning
    /// Any data that has been appended but not yet flushed will be silently
    /// discarded. Call [`flush`](Writer::flush) before `reset()` if you need
    /// to preserve in-flight records.
```

avro/src/writer.rs
```diff
@@ -542,6 +542,60 @@ impl<'a, W: Write> Writer<'a, W> {
...
+    pub fn reset(&mut self) {
+        self.buffer.clear();
+        self.writer.clear();
+        self.has_header = false;
```
martin-g, line 598: Do we need a new `marker` here ?!

avro/src/writer.rs
```diff
@@ -542,6 +542,60 @@ impl<'a, W: Write> Writer<'a, W> {
     }
 }
 
+/// A buffer that can be cleared.
+pub trait Clearable {
+    /// Clear the buffer, keeping the capacity.
```
```suggestion
    /// Clear the buffer.
```
There is no need to mention the `capacity` here. We don't really require it.

avro/src/writer.rs
martin-g: I think it is OK to reuse the marker.
Since it is random any consumer of the Avro bytes should not expect that it is unique.

PR #470, opened 2026-02-19
feat: Make `Schema::denormalize` public

Kriskras99:
This should help fix https://github.com/lerouxrgd/rsgen-avro/issues/92

I've not applied the `with_capacity(schemata.len())` optimisation that `independant_canonical_form` had, as that capacity will be wrong in 99% of usecases. This should allow the capacity to grow in more predictable ways.

### DIFF
--- a/avro/src/schema/mod.rs
+++ b/avro/src/schema/mod.rs
@@ -398,7 +398,7 @@ impl Schema {
     /// Returns the [Parsing Canonical Form] of `self` that is self contained (not dependent on
     /// any definitions in `schemata`)
     ///
-    /// If your require a self contained schema including `default` and `doc` attributes, see [`denormalize`][Schema::denormalize].
+    /// If you require a self contained schema including `default` and `doc` attributes, see [`denormalize`][Schema::denormalize].
     ///
     /// [Parsing Canonical Form]:
     /// https://avro.apache.org/docs/++version++/specification/#parsing-canonical-form-for-schemas


avro/src/schema/mod.rs
```diff
@@ -398,11 +398,13 @@ impl Schema {
     /// Returns the [Parsing Canonical Form] of `self` that is self contained (not dependent on
     /// any definitions in `schemata`)
     ///
+    /// If your require a self contained schema including `default` and `doc` attributes, see [`denormalize`][Schema::denormalize].
```
```suggestion
    /// If you require a self contained schema including `default` and `doc` attributes, see [`denormalize`][Schema::denormalize].
```

avro/src/schema/mod.rs
```diff
@@ -688,7 +690,18 @@ impl Schema {
...
-    fn denormalize(
+    /// Remove all external references from the schema.
+    ///
+    /// `schemata` must contain all externally referenced schemas.
+    ///
+    /// # Errors
+    /// Will return a [`Details::SchemaResolutionError`] if it fails to find
+    /// a referenced schema. This will put the schema in a partly denormalized state.
+    pub fn denormalize(&mut self, schemata: &[Schema]) -> AvroResult<()> {
+        self.denormalize_inner(schemata, &mut HashSet::new())
```
martin-g, line 701:
```suggestion
        let mut candidate = self.clone();
        candidate.denormalize_inner(schemata, &mut HashSet::new())?;
        *self = candidate;
        Ok(())
```
to prevent `This will put the schema in a partly denormalized state.`

Kriskras99: It is fine if the schema is partly denormalized isn't it? It is still a valid schema, some (but not all) references have just been inlined.

Kriskras99: And if users really don't want that, they can do the clone themselves.

PR #472, opened 2026-02-19
feat!: Use bon builder style instead of several methods

martin-g:
Instead of having several methods like: `Schema::array(Schema)` and `Schema::array_with_attributes(Schema, BTreeMap)` we could use Bon's function builders with start_fn:
```
Schema::array(Schema).call()
Schema::array(Schema).attributes(BTreeMap).call()
```

The only annoying part is the `.call()`

### DIFF
--- a/avro/src/schema/builders.rs
+++ b/avro/src/schema/builders.rs
@@ -15,11 +15,12 @@
 // specific language governing permissions and limitations
 // under the License.
 
-use crate::Schema;
 use crate::schema::{
     Alias, ArraySchema, EnumSchema, FixedSchema, MapSchema, Name, RecordField, RecordSchema,
+    UnionSchema,
 };
 use crate::types::Value;
+use crate::{AvroResult, Schema};
 use bon::bon;
 use serde_json::Value as JsonValue;
 use std::collections::{BTreeMap, HashMap};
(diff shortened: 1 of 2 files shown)


WARP.md
Kriskras99, line 1: I don't think we should commit files for specific tooling. We could add a `LLM.md` or something similar.

martin-g: It was not intentional!

avro/src/reader.rs
```diff
@@ -337,57 +338,39 @@ pub struct Reader<'a, R> {
...
     ///
     /// **NOTE** The avro header is going to be read automatically upon creation of the `Reader`.
-    pub fn with_schemata(
-        schema: &'a Schema,
-        schemata: Vec<&'a Schema>,
-        reader: R,
+    #[builder(finish_fn = build)]
+    pub fn builder(
+        #[builder(start_fn)] reader: R,
+        schema: Option<&'a Schema>,
```
Kriskras99: Maybe rename this to `reader_schema`, so it's clearer that this is not the writer schema

avro/src/reader.rs
```diff
@@ -337,57 +338,40 @@ pub struct Reader<'a, R> {
...
     /// **NOTE** The avro header is going to be read automatically upon creation of the `Reader`.
-    pub fn with_schemata(
-        schema: &'a Schema,
-        schemata: Vec<&'a Schema>,
-        reader: R,
+    #[builder(finish_fn = build)]
+    pub fn builder(
+        #[builder(start_fn)] reader: R,
+        reader_schema: Option<&'a Schema>,
+        schemata: Option<Vec<&'a Schema>>,
```
martin-g, line 360: Before it was not possible to set only `schemata` without `reader_schema`.
Now it is possible but all tests pass. I am not sure whether this could lead to any problems.

Kriskras99: I think this is fine. Then `schemata` just needs to contain any references in the `writer_schema`.

Kriskras99:
The `*datum*` functions would also be a good fit, although it would be better of those functions are removed in favour of a `DatumReader`/`DatumWriter` struct that can actually cache the resolved schemas.

We can do that in a separate PR.

Kriskras99 (approved):

PR #477, opened 2026-02-20
feat: Allow types to provide default values

### LINKED ISSUE #434: support for serde_as: skip_serializing_none
Hi! Thank you so much for the work you've put in the serde integration. We used to have a custom crate to overwrite some of the behavior of older versions of this package (to respect some of the serde macros), but I gave a go at using what's now on main and it works great!!!

I only have one comment. it's not uncommon to have large structs, and for these to have a bunch of optional fields. When that happens, one might want to not serialize the None fields (say you are writing documents to a database, outputting data as JSON for an API, ...). There, folks often use the `serde_as` crate, as such:
```rust
#[serde_as]
#[skip_serializing_none]
#[derive(Deserialize, Serialize)]
struct Test {
  a: i64,
  b: Option<String>
}
```

here with just one optional fields it's just overkill, but when you have a lot of them it's very elegant to use that "struct-wide" `#[skip_serializing_none]` rather than having to set a skip serialize attribute on all optional fields. Unfortunately, that doesn't seem to work at the moment (previous version of the code actually didn't complain about these things, likely because it ignored serde stuff altogether?). Instead, it shows me this error:
```
#[serde(skip_serializing)]` and `#[serde(skip_serializing_if)]` require `#[avro(default = "..")]
```
That is solved by, as stated, adding a `#[avro(default = "null")` on all the optional fields of the struct.

But, it would be a bit more elegant if it could just automatically infer that the default should be null/None for these optional fields, so we could still use `serde_as + skip_serializing_none`

Kriskras99:
`AvroSchemaComponent` is extended with a function `field_default` which will be called when deriving a record to set the default value for a field. The default implementation is to return `None`, which means no default.

On the derive side, it is now possible to specify a default for a type using `#[avro(default = "..")]`. It is also possible to disable setting a default for a field with `#[avro(default = false)]`.

This enables users to use `#[serde(skip_serializing{_if})]` on most fields without having to provide a default value.

Closes #434

### DIFF
--- a/avro_derive/tests/derive.rs
+++ b/avro_derive/tests/derive.rs
@@ -2571,7 +2571,7 @@ fn avro_rs_476_skip_serializing_fielddefault_trait_none() {
         Err(e) => match e.into_details() {
             apache_avro::error::Details::MissingDefaultForSkippedField { field_name, .. }
                 if field_name == "_y" => {}
-            d => panic!("Unexpected error: {:?}", d),
+            d => panic!("Unexpected error: {d:?}"),
         },
     }
 }


avro/src/serde/derive.rs
```diff
@@ -562,6 +614,13 @@ where
     ) -> Option<Vec<RecordField>> {
         None
     }
+
+    /// If `T` has a field default, this will return an array of with that default. Otherwise there is no default.
```
```suggestion
    /// If `T` has a field default, this will return an array of elements with that default. Otherwise there is no default.
```

avro/src/serde/derive.rs
```diff
@@ -562,6 +614,13 @@ where
...
+    fn field_default() -> Option<serde_json::Value> {
+        T::field_default().map(|default| {
+            serde_json::Value::Array(std::array::from_fn::<_, N, _>(|_| default.clone()).to_vec())
```
```suggestion
            serde_json::Value::Array(vec![default; N])
```
Does this work ?
Or it needs to call `.clone()` explicitly ?

avro_derive/src/attributes/mod.rs
```diff
@@ -116,12 +118,29 @@ impl NamedTypeOptions {
...
         let doc = avro.doc.or_else(|| extract_rustdoc(attributes));
 
+        let default = match avro.default {
+            None => quote! { None },
+            Some(default_value) => {
+                let _: serde_json::Value =
+                    serde_json::from_str(&default_value[..]).map_err(|e| {
+                        vec![syn::Error::new(
+                            ident.span(),
+                            format!("Invalid avro default json: \n{e}"),
```
```suggestion
                            format!("Invalid Avro `default` JSON: \n{e}"),
```

avro/src/serde/derive.rs
```diff
@@ -81,6 +85,10 @@ use std::collections::{HashMap, HashSet};
 ///
 ///    Set the `doc` attribute of the schema. Defaults to the documentation of the type.
 ///
+///  - `#[avro(default = r#"{"field": 42, "other": "Spam"}"#)]`
```
martin-g, line 84: I am not sure this is a good idea.
Using such container level attribute with stringified map looks fragile. If a field is renamed then its string key here may become obsolete/stale.
The field level attribute looks better to me. More type safe.

avro/src/serde/derive.rs
```diff
@@ -113,11 +121,14 @@ use std::collections::{HashMap, HashSet};
 ///
 ///    Set the `doc` attribute of the field. Defaults to the documentation of the field.
 ///
-///  - `#[avro(default = "null")]`
+///  - `#[avro(default = "null")]` or `#[avro(default = false)]`
 ///
-///    Set the `default` attribute of the field.
+///    Control the `default` attribute of the field. When not used, it will use [`AvroSchemaComponent::field_default`]
+///    to get the default value for a type. This default value can be overriden by providing a JSON string.
+///    To remove the `default` attribute for a field, set `default` to `false`.
```
martin-g: nit: What if one needs to set the default to `false` explicitly ?
Can we use some marker struct/constant instead ?

avro/src/serde/derive.rs
```diff
@@ -479,6 +515,10 @@ where
 
 macro_rules! impl_schema (
     ($type:ty, $variant_constructor:expr) => (
+        impl_schema!($type, $variant_constructor, <$type as Default>::default());
```
martin-g: Is there a reason to use `Default::default()` by default ?
IMO the default should be `None` and only the app developer should explicitly sets non-None defaults.

avro_derive/src/attributes/mod.rs
```diff
@@ -116,12 +118,29 @@ impl NamedTypeOptions {
...
+                        )]
+                    })?;
+                quote! {
+                    Some(serde_json::from_str(#default_value).expect(format!("Invalid JSON: {:?}", #default_value).as_str()))
```
martin-g: Why the second parsing is needed here ?

avro_derive/src/attributes/mod.rs
```diff
@@ -116,12 +118,29 @@ impl NamedTypeOptions {
 
         let doc = avro.doc.or_else(|| extract_rustdoc(attributes));
 
+        let default = match avro.default {
+            None => quote! { None },
+            Some(default_value) => {
+                let _: serde_json::Value =
```
martin-g, line 124:
```suggestion
                let json: serde_json::Value =
```

avro_derive/src/attributes/mod.rs
```suggestion
                    Some(#json)
```
This does not work ?!

avro_derive/src/attributes/mod.rs
```diff
@@ -274,11 +320,11 @@ impl FieldOptions {
         }
         if ((serde.skip_serializing && !serde.skip_deserializing)
             || serde.skip_serializing_if.is_some())
-            && avro.default.is_none()
+            && avro.default == FieldDefault::Disabled
```
martin-g, line 323: What happens if `avro.default` is `FieldDefault::Trait` but `field_default()` returns `None` ?

avro_derive/tests/derive.rs
```diff
@@ -2377,3 +2378,203 @@ fn avro_rs_448_flatten_field_positions() {
...
+        _o: Vec<bool>,
+        #[avro(default = "[1,2,3,4,5]")]
+        _p: [u8; 5],
+        #[avro(
+            default = r#"[{"_field": true},{"_field": false},{"_field": true},{"_field": false},{"_field": true}]"#
+        )]
+        _p_alt: [Spam; 5],
+        #[avro(default = r#"{"A": "B"}"#)]
+        _q: HashMap<String, String>,
+        #[avro(default = "42.0")]
```
martin-g, line 2516: Since this is an Option it will generate a Union schema with Null being the first variant, so the default must be `null`.

Another way is to detect that the default type is not Null and generate a Union schema with a proper first variant that matches the default's type. But this would be more complex.

avro/src/serde/derive.rs
```diff
@@ -113,11 +121,14 @@ use std::collections::{HashMap, HashSet};
 ///
 ///    Set the `doc` attribute of the field. Defaults to the documentation of the field.
 ///
-///  - `#[avro(default = "null")]`
+///  - `#[avro(default = "null")]` or `#[avro(default = false)]`
 ///
-///    Set the `default` attribute of the field.
+///    Control the `default` attribute of the field. When not used, it will use [`AvroSchemaComponent::field_default`]
+///    to get the default value for a type. This default value can be overriden by providing a JSON string.
```
```suggestion
///    to get the default value for a type. This default value can be overridden by providing a JSON string.
```

avro_derive/tests/derive.rs
Kriskras99: The default does not need to be a `null`, according to spec:
```
Default values for union fields correspond to the first schema that matches in the union.
```

avro_derive/src/attributes/mod.rs
Kriskras99: Nope. The first time we parse is during the build. We do that so we can check the JSON is at least valid.
The second time we parse is when we create the `Schema` at runtime (see the `quote!` block), which can't reuse the build time parsing because the original parsing does not exist at runtime.
If you do try this, you get:
```
the trait bound `Value: darling::ToTokens` is not satisfied [E0277]
```

avro/src/serde/derive.rs
Kriskras99: If a user wants to set the JSON value `false` it would be `#[avro(default = "false")]`.
I clearly need to improve the documentation.

avro/src/serde/derive.rs
Kriskras99: The top-level attribute is only used for implementing `AvroSchemaComponent::field_default`, it is not used to set defaults for the individual fields.

avro_derive/src/attributes/mod.rs
Kriskras99: We unfortunately cannot detect that at compile time, so that will produce an error at runtime.

Kriskras99 (commented):
My goal for this PR is to allow users to have defaults for fields with as little effort as possible.
That's why I implemented `AvroSchemaComponent::field_default()` as `Some(Default::default())` for all types we implement.
I can understand if some users don't want to have a default for a field, which is why I added `#[avro(default = false)]` to disable getting a default for that field. This allows for usecases like in #434

avro_derive/tests/derive.rs
martin-g: Hm. You are right that the spec does not say what I said ...
But I remember this rule from somewhere. Maybe an issue in JIRA or in the mailing lists...
Anyway, not a stopper!

avro/src/serde/derive.rs
```diff
@@ -609,6 +678,10 @@ where
     ) -> Option<Vec<RecordField>> {
         None
     }
+
+    fn field_default() -> Option<serde_json::Value> {
+        Some(serde_json::Value::Null)
```
martin-g, line 658: This should be enough for #434 !

But I still think we should return `None` by default for anything else, instead of `Some(JsonValue::from(Default::default()))`.

Imagine the developer uses `#[serde(skip_serializing_if=123)]` (or any other custom value). Then they should face an error that a default should be specified. Currently we would default to 0 (for a numeric type) while the user will actually need `123`. So, the user will realize the problem sooner or later and (s)he will still need to provide `#[avro(default = "123")]` anyway.

avro_derive/src/attributes/mod.rs
martin-g: If `None` is the default result from `field_default()` (as I suggest) then this won't work at all ... :-/
Let me think some more on this.

Kriskras99:
I've removed the `field_default` implementation for all types except `Option<T>`. It's easier to add new implementations later than removing existing implementations.

avro/src/serde/derive.rs
```diff
@@ -562,6 +614,11 @@ where
     ) -> Option<Vec<RecordField>> {
         None
     }
+
+    /// If `T` has a field default, this will return an array of elements with that default. Otherwise there is no default.
```
```suggestion
```

Kriskras99:
Looks good!

PR #479, opened 2026-02-20
fix!: Stricter schema parsing

### LINKED ISSUE #476: `type.enum` is valid while it should fail, good is `type.type.enum` according to spec
Hi, a schema like this is parsed correctly by avro-rs, this tests fails (it should err):

```rs
fn test_wrong_enum() {
    let schema_str = r#"{
        "type": "record",
        "name": "ExampleEnum",
        "namespace": "com.schema",
        "fields": [
            {
            "name": "wrong_enum",
            "type": "enum",
            "symbols": ["INSERT", "UPDATE"]
            }
        ]
    }"#;
    let result = Schema::parse_str(schema_str);
    assert!(result.is_err());
}
```

While tools like `avro-tools` fail to create the enums with this schema.

The problem is that `enum` should be defined inside `enum`, this would be a good looking example.
Notice the `type.type.enum`:

```json
{
  "type": "record",
  "name": "ExampleEnum",
  "namespace": "com.schema",
  "fields": [
    {
      "name": "wrong_enum",
      "type": {
        "name": "WrongEnum",
        "type": "enum",
        "symbols": ["INSERT", "UPDATE"]
      }
    }
  ]
}
```

I couldn't find any example in the specification of `type.enum`, and given that `avro-tools` fails to generate the java code, I assume `type.enum` is wrong. Let me know if that's not the case.

Best.

Kriskras99:
This will break code, as it no longer accepts invalid schemas.

Closes #476

### DIFF
--- a/avro/src/error.rs
+++ b/avro/src/error.rs
@@ -405,7 +405,7 @@ pub enum Details {
     #[error("No `type` in complex type")]
     GetComplexTypeField,
 
-    #[error("No `type` in in record field")]
+    #[error("No `type` in record field")]
     GetRecordFieldTypeField,
 
     #[error("No `fields` in record")]
@@ -433,7 +433,7 @@ pub enum Details {
     InvalidNamespace(String, &'static str),
 
     #[error(
-        "Invalid schema: There is no type called '{0}', if you meant to define a non-primitive schema, it should be defined inside `type` attribute. Please review the specification"
+        "Invalid schema: There is no type called '{0}', if you meant to define a non-primitive schema, it should be defined inside `type` attribute."
     )]
     InvalidSchemaRecord(String),
 
--- a/avro/src/schema/mod.rs
+++ b/avro/src/schema/mod.rs
@@ -1055,7 +1055,7 @@ fn pcf_array(arr: &[JsonValue], defined_names: &mut HashSet<String>) -> String {
 }
 
 fn pcf_string(s: &str) -> String {
-    format!("\"{s}\"")
+    format!(r#""{s}""#)
 }
 
 const RESERVED_FIELDS: &[&str] = &[
@@ -1101,9 +1101,9 @@ mod tests {
 
     #[test]
     fn test_primitive_schema() -> TestResult {
-        assert_eq!(Schema::Null, Schema::parse_str("\"null\"")?);
-        assert_eq!(Schema::Int, Schema::parse_str("\"int\"")?);
-        assert_eq!(Schema::Double, Schema::parse_str("\"double\"")?);
+        assert_eq!(Schema::Null, Schema::parse_str(r#""null""#)?);
+        assert_eq!(Schema::Int, Schema::parse_str(r#""int""#)?);
+        assert_eq!(Schema::Double, Schema::parse_str(r#""double""#)?);
         Ok(())
     }
 
@@ -1326,7 +1326,7 @@ mod tests {
             Ok(_) => unreachable!("Expected an error that the name is already defined"),
             Err(e) => assert_eq!(
                 e.to_string(),
-                "Two schemas with the same fullname were given: \"A\""
+                r#"Two schemas with the same fullname were given: "A""#
             ),
         }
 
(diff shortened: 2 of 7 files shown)


avro/src/schema/mod.rs
```diff
@@ -3815,20 +3799,10 @@ mod tests {
...
-        let result = Schema::parse_str(schema_str);
-        assert!(result.is_err());
-        let err = result
-            .map_err(|e| e.to_string())
-            .err()
-            .unwrap_or_else(|| "unexpected".to_string());
-        assert_eq!(expected, err);
+        assert_eq!(
+            Schema::parse_str(schema_str).unwrap_err().to_string(),
+            "`default`'s value type of field `f1` in `ns.record1` must be a `{\"name\":\"ns.record2\",\"type\":\"record\",\"fields\":[{\"name\":\"f1_1\",\"type\":\"int\"}]}`. Got: String(\"invalid\")"
```
martin-g: nit: Let's use raw string to make a bit more readable. `r#"..."#`
Same for the other similar assertions.

avro/src/schema/mod.rs
```diff
@@ -5438,4 +5401,26 @@ mod tests {
...
+                "name": "wrong_enum",
+                "type": "enum",
+                "symbols": ["INSERT", "UPDATE"]
+                }
+            ]
+        }"#;
+        let result = Schema::parse_str(schema_str).unwrap_err();
+        assert_eq!(
+            result.to_string(),
+            "Invalid schema: There is no type called 'enum', if you meant to define a non-primitive schema, it should be defined inside `type` attribute. Please review the specification"
```
```suggestion
            "Invalid schema: There is no type called 'enum', if you meant to define a non-primitive schema, it should be defined inside `type` attribute. Please consult with the specification"
```

avro/src/error.rs
```diff
@@ -405,6 +405,9 @@ pub enum Details {
     #[error("No `type` in complex type")]
     GetComplexTypeField,
 
+    #[error("No `type` in in record field")]
```
```suggestion
    #[error("No `type` in record field")]
```

avro/src/schema/record/field.rs
```diff
@@ -81,12 +80,14 @@ impl RecordField {
...
-            field,
-            &enclosing_record.namespace,
-            RecordSchemaParseLocation::FromField,
-        )?;
+        let ty = field.get("type").ok_or(Details::GetRecordFieldTypeField)?;
+        let schema = parser.parse(ty, &enclosing_record.namespace)?;
+
+        if let Some(logical_type) = field.get("logicalType") {
+            warn!(
+                "Ignored the {enclosing_record}.logicalType property (`{logical_type}`). It should probably be nested instide the `type` for the field"
```
martin-g:
```suggestion
                "Ignored the {enclosing_record}.logicalType property (`{logical_type}`). It should probably be nested inside the `type` for the field"
```

avro/src/schema/record/field.rs
martin-g: This needs reworking too.
`logicalType`, `symbols`, `size`, `items` and `values` cannot be at that level anymore.

avro/src/schema/record/field.rs
```diff
@@ -162,10 +163,14 @@ impl RecordField {
...
 
                     if !resolved {
+                        let schemata = names.values().cloned().collect::<Vec<_>>();
                         return Err(Details::GetDefaultRecordField(
                             field_name.to_string(),
                             record_name.to_string(),
-                            field_schema.canonical_form(),
+                            field_schema
+                                .independent_canonical_form(&schemata)
+                                .unwrap_or("Unknown".to_string()),
```
martin-g: How about falling back to the canonical form instead of this "Unknown" ?

martin-g (commented):
Here are some comments on the proposed changes in this PR.

But as I commented at https://github.com/apache/avro-rs/issues/476#issuecomment-3943071557 the Java and Python SDKs do support parsing such flat schemas too.

avro/src/schema/mod.rs
Kriskras99: I've removed the entire `Please review the specification`/`Please consult with the specification` bit

avro/src/schema/record/field.rs
Kriskras99: I've also removed the `position` field, as it is not in the spec or used in the code.

avro/src/schema/record/field.rs
Kriskras99: Great idea!

martin-g (approved):

PR #486, opened 2026-02-25
fix!: Small things

Kriskras99:
These are a bunch of small fixes that don't really deserve their own PRs.

- fix: Use `#[track_caller]` for `assert{,_not}_logged` to show the caller source location when panicking
- fix: Use `Write::write_all` in `utils::encode_variable`
- fix: Change default of `SER_BYTES_TYPE` to `BytesType::Unset` and set if for every serialize/deserialize call
  - This is not used yet, but will be used in the improvements to `SchemaAwareWriteSerializer` that will also add enum support
- fix: Implement `AvroSchemaComponent` for `()`
- fix!: `Alias::name` and `Alias::namespace` should return references, user can clone if needed
  - This is a breaking change, but easy to resolve by calling `.clone()` on the return value of the functions
- fix: `deserialize_big_decimal` doesn't need a `Vec<u8>`
  - This is not used yet, but will also be used in the improvements to `SchemaAwareWriteSerializer`

### DIFF
--- a/avro/src/serde/derive.rs
+++ b/avro/src/serde/derive.rs
@@ -946,7 +946,7 @@ mod tests {
     }
 
     #[test]
-    fn avro_rs_xxx_unit() -> TestResult {
+    fn avro_rs_486_unit() -> TestResult {
         let schema = <()>::get_schema();
         assert_eq!(schema, Schema::Null);
 


avro/src/serde/with.rs
```diff
@@ -184,13 +189,15 @@ pub mod bytes_opt {
...
     {
+        let _guard = super::BytesTypeGuard::set(BytesType::Bytes);
         serde_bytes::serialize(bytes, serializer)
     }
 
     pub fn deserialize<'de, D>(deserializer: D) -> Result<Option<Vec<u8>>, D::Error>
     where
         D: Deserializer<'de>,
     {
+        let _guard = super::BytesTypeGuard::set(BytesType::Bytes);
```
martin-g, line 200: This is not needed for `deserialize`.
It is used only in avro/src/serde/ser.rs > serialize_bytes() (line 183)

Same below.

Kriskras99: Yes, that's correct. To implement enum support I'll also be adding a `SchemaAwareReadDeserializer` which does need it.

PR #489, opened 2026-02-26
feat!: Add a `UnionSchemaBuilder`

Kriskras99:
This also fixes a issue with the original `new` implementation where it would insert named types in the `variant_index` and then `find_schema_with_known_schemata` would use the fast path without checking the schema.

`find_schema_with_known_schemata` has also been simplified to use `known_schemata` directly instead of rebuilding it with the current schema, as this would cause duplicate schema errors after the incorrect fast path was removed.

Another issue fixed is that logical types cannot be added if their base type is already in the union, and the same in reverse.

The `UnionSchemaBuilder::variant_ignore_duplicates` and `UnionSchemaBuilder::contains` are needed for `avro_derive` to implement full support for enums.

This is a breaking change as it no longer allows certain schemas (that shouldn't have been allowed in the first place).

### DIFF
--- a/avro/src/schema/union.rs
+++ b/avro/src/schema/union.rs
@@ -236,7 +236,7 @@ impl PartialEq for UnionSchema {
 #[derive(Default, Debug)]
 pub struct UnionSchemaBuilder {
     schemas: Vec<Schema>,
-    names: BTreeMap<Name, usize>,
+    names: HashMap<Name, usize>,
     variant_index: BTreeMap<SchemaKind, usize>,
 }
 
--- a/avro/src/schema/name.rs
+++ b/avro/src/schema/name.rs
@@ -38,7 +38,7 @@ use crate::{
 ///
 /// More information about schema names can be found in the
 /// [Avro specification](https://avro.apache.org/docs/++version++/specification/#names)
-#[derive(Clone, Debug, Hash, PartialEq, Eq, PartialOrd, Ord)]
+#[derive(Clone, Debug, Hash, PartialEq, Eq)]
 pub struct Name {
     pub name: String,
     pub namespace: Namespace,


avro/src/serde/derive.rs
Kriskras99, line 1: I've accidently added the fix in #486 in here, but we can just leave it in

avro/src/schema/union.rs
```diff
@@ -15,24 +15,32 @@
...
-    // to support that. A simple solution is to also keep a mapping of the names used.
+    /// The indexes of unnamed types.
+    ///
+    /// Logical types have been reduced to their inner type.
+    /// Used to provide constant time finding of the
+    /// schema index given an unnamed type. Must only contain unnamed types.
     variant_index: BTreeMap<SchemaKind, usize>,
+    /// The indexes of named types.
+    ///
+    /// The names self aren't saved as they aren't used.
```
martin-g:
```suggestion
    /// The names themselves aren't saved as they aren't used.
```

avro/src/schema/union.rs
```diff
@@ -15,24 +15,32 @@
...
     /// The schemas that make up this union
     pub(crate) schemas: Vec<Schema>,
-    // Used to ensure uniqueness of schema inputs, and provide constant time finding of the
-    // schema index given a value.
-    // **NOTE** that this approach does not work for named types, and will have to be modified
-    // to support that. A simple solution is to also keep a mapping of the names used.
+    /// The indexes of unnamed types.
+    ///
+    /// Logical types have been reduced to their inner type.
+    /// Used to provide constant time finding of the
```
martin-g: > constant time finding 

I wonder why this old doc says that. BTreeMap's search is O(log n), not constant.

avro/src/schema/union.rs
```diff
@@ -93,57 +92,301 @@ impl UnionSchema {
...
+            } else {
+                self.variant_index
+                    .insert(SchemaKind::Map, self.schemas.len());
+                self.schemas.push(schema);
+            }
+        } else if let Schema::Array(_) = &schema {
+            if let Some(index) = self.variant_index.get(&SchemaKind::Array).copied() {
+                if self.schemas[index] != schema {
+                    return Err(
+                        Details::GetUnionDuplicateMap(self.schemas.remove(index), schema).into(),
```
martin-g:
```suggestion
                        Details::GetUnionDuplicateArray(self.schemas.remove(index), schema).into(),
```

avro/src/schema/union.rs
```diff
@@ -93,57 +92,301 @@ impl UnionSchema {
...
+            }
+        }
+    }
+
+    /// Create the `UnionSchema`.
+    pub fn build(mut self) -> UnionSchema {
+        self.schemas.shrink_to_fit();
+        UnionSchema {
+            variant_index: self.variant_index,
+            named_index: self.names.into_values().collect(),
```
martin-g: Hm.
`names` is a BTreeMap, i.e. the keys are sorted by Name.
`into_values().collect()` will return a `Vec<usize>` that is not ordered by the variant's insertion order.
The iteration at line 128 (https://github.com/apache/avro-rs/pull/489/changes#diff-4d9db0a667065166534f4c7e90caaa27e60109325303dee2730f12182877a9f8R128) may return a Schema that resolves well (i.e. has same type) but has a different name.

avro/src/schema/union.rs
```diff
@@ -93,57 +92,301 @@ impl UnionSchema {
...
-            let mut collected_names: HashMap<Name, &Schema> = known_schemata
-                .map(|names| {
-                    names
-                        .iter()
-                        .map(|(name, schema)| (name.clone(), schema.borrow()))
-                        .collect()
+        let ValueSchemaKind { unnamed, named } = Self::value_to_base_schemakind(value);
+        // Unnamed schema types can be looked up directly using the variant_index
+        let unnamed = unnamed
+            .and_then(|kind| self.variant_index.get(&kind).copied())
```
martin-g, line 98: AFAIS there is no "slow path" anymore for unnamed schemas.
That means "type promotion" is no more supported.
Value::Int won't resolve with Schema::Long anymore.

avro/src/serde/derive.rs
martin-g: Let's add a test for this.

avro/src/schema/union.rs
```diff
@@ -93,57 +92,301 @@ impl UnionSchema {
...
+            // Every named type needs to be checked against a value until one matches
+
+            let known_schemata_if_none = HashMap::new();
+            let known_schemata = known_schemata.unwrap_or(&known_schemata_if_none);
+
+            self.named_index
+                .iter()
+                .copied()
+                .map(|i| (i, &self.schemas[i]))
+                .filter(|(_i, s)| s.discriminant() == kind || s.discriminant() == SchemaKind::Ref)
```
martin-g: Shouldn't this check against the base SchemaKind ?
```suggestion
                .filter(|(_i, s)| {
                    s.discriminant() == SchemaKind::Ref
                        || UnionSchemaBuilder::schema_to_base_schemakind(s) == kind
                })
```

avro/src/schema/union.rs
```diff
@@ -93,57 +92,301 @@ impl UnionSchema {
...
+                }
+            } else {
+                self.names.insert(name.clone(), self.schemas.len());
+                self.schemas.push(schema);
+            }
+        } else if let Schema::Map(_) = &schema {
+            if let Some(index) = self.variant_index.get(&SchemaKind::Map).copied() {
+                if self.schemas[index] != schema {
+                    return Err(
+                        Details::GetUnionDuplicateMap(self.schemas.remove(index), schema).into(),
```
martin-g: What is the reason to remove the old schema here ?
If the idea is to make it possible to attempt again then the indices (`names` and `variant_index`) should be updated too.

avro/src/schema/union.rs
```diff
@@ -93,57 +92,301 @@ impl UnionSchema {
...
+        Self::default()
+    }
+
+    #[doc(hidden)]
+    /// This is not a public API, it should only be used by `avro_derive`
+    ///
+    /// Add a variant to this union, if it already exists ignore it.
+    ///
+    /// # Errors
+    /// Will return a [`Details::GetUnionDuplicateMap`] or [`Details::GetUnionDuplicateArray`] if
```
martin-g, line 257: What is the reason Map and Array are specially handled here ?

Kriskras99: Because they are the only two unnamed types that cannot be differentiated through schemakind alone, as you also need to check the inner type.
It can also be quite confusing if the error is just `Duplicate Map found` without explaining that the inner type is different.

avro/src/schema/union.rs
Kriskras99: Leftover optimisation from when the builder would take `self` (and thus when returning an error would stop to exist).

avro/src/schema/union.rs
Kriskras99: Yes, good find! Fixed and added test.

avro/src/schema/union.rs
Kriskras99: I've added a slow path that only activates if both `named` and `unnamed` are `None`.
Also added a test.

avro/src/schema/union.rs
Kriskras99: Good point, I now sort the `named_index` before storing it.

avro/src/schema/union.rs
Kriskras99: I've removed that comment. I do think the `BTreeMap` is a good fit for this problem, as for small unions, they will actually be stored in an array so cache locality is very good.

avro/src/serde/derive.rs
Kriskras99: I think this is actually wrong. `Some(())` would be `None` after a roundtrip through Serde.

Kriskras99: I've removed the fix, and added tests to make sure that it panics.
We unfortunately cannot detect this at compile time.

avro/src/schema/union.rs
```diff
@@ -93,57 +90,320 @@ impl UnionSchema {
...
+                false
+            }
+        }
+    }
+
+    /// Create the `UnionSchema`.
+    pub fn build(mut self) -> UnionSchema {
+        self.schemas.shrink_to_fit();
+        let mut named_index: Vec<_> = self.names.into_values().collect();
+        named_index.sort();
```
martin-g, line 358: 👍🏻 

nit: With this solution we don't really need BTreeMap (and Ord+PartialOrd for Name). A HashMap would work too

Kriskras99: I don't have a strong preference between `BTreeMap` and `HashMap`. I used a `BTreeMap` because the `variant_index` already used one.
Do you want to switch to a `HashMap`?

martin-g: My only concern is the `Name` struct. Now is has auto-derived Ord and PartialOrd impls.
I haven't checked what is the derived implementation but I have the feeling that it is based on the fields' order, i.e. first the `name`s are compared and if they match then the `namespace`s are compared.
Intuitively, this is wrong. IMO the ordering should use the full name, i.e. "namespace.name"

Kriskras99: We could implement `Ord` manually, but the documentation says:
```If you implement it manually, you should manually implement all four traits, based on the implementation of Ord```
Which is too much work of too little gain. I've replaced the `BTreeMap` with a `HashMap`.

PR #491, opened 2026-02-27
fix!: Remove unused fields in `RecordField` (`order`, `position`)

### LINKED ISSUE #490: Remove `RecordField` `position` and `order`
Both fields aren't really used in the code, and make the existing code more complex.
`order` has in its documentation that it is not used.
`position` is only used in the `SchemaAwareWriteSerializer` where the value could also have been passed in as a function argument from the parent function.

Needing to fill the `position` fields make generating code in `avro_derive` significantly complex. So removing it would simplify a lot of code (we would also need to generate less code which makes `T::get_schema()` faster at runtime).

@martin-g what do you think?

Kriskras99:
Also changes the type of `aliases` from `Option<Vec<String>>` to `Vec<String>`.

I've changed almost all the places where we instantiate `RecordField` to use the builder syntax which is less verbose.
I've also removed two tests that were testing the `position` field.

Closes #490

### DIFF
--- a/avro/src/schema/record/field.rs
+++ b/avro/src/schema/record/field.rs
@@ -69,7 +69,7 @@ impl Debug for RecordField {
             debug.field("custom_attributes", &self.custom_attributes);
         }
         if self.doc.is_none()
-            || !self.aliases.is_empty()
+            || self.aliases.is_empty()
             || self.default.is_none()
             || self.custom_attributes.is_empty()
         {
--- a/avro/src/serde/derive.rs
+++ b/avro/src/serde/derive.rs
@@ -191,7 +191,7 @@ use std::collections::{HashMap, HashSet};
 /// 1. In combination with `#[serde(with = "path::to::module)]`
 ///
 ///    To get the schema, it will call the functions `fn get_schema_in_ctxt(&mut HashSet<Name>, &Namespace) -> Schema`
-///    and `fn get_record_fields_in_ctxt(usize, &mut HashSet<Name>, &Namespace) -> Option<Vec<RecordField>>` in the module provided
+///    and `fn get_record_fields_in_ctxt(&mut HashSet<Name>, &Namespace) -> Option<Vec<RecordField>>` in the module provided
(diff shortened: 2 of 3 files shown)


avro/src/schema/record/field.rs
```diff
@@ -67,38 +58,32 @@ impl Debug for RecordField {
...
-        // This field is ignored as it currently has no effect
-        // debug.field("order", &self.order);
-        debug.field("position", &self.position);
         if !self.custom_attributes.is_empty() {
             debug.field("custom_attributes", &self.custom_attributes);
         }
-        // As we are always skipping self.order, always show the ..
-        debug.finish_non_exhaustive()
+        if self.doc.is_none()
+            || !self.aliases.is_empty()
```
martin-g:
```suggestion
            || self.aliases.is_empty()
```

avro/src/serde/derive.rs
```suggestion
///    and `fn get_record_fields_in_ctxt(&mut HashSet<Name>, &Namespace) -> Option<Vec<RecordField>>` in the module provided
```

avro/src/serde/ser_schema/mod.rs
```diff
@@ -373,52 +378,45 @@ impl<W: Write> ser::SerializeStruct for SchemaAwareWriteSerializeStruct<'_, '_,
...
-                Details::SerializeRecordFieldWithSchema {
-                    field_name: key.to_string(),
-                    record_schema: Schema::Record(self.record_schema.clone()),
-                    error: Box::new(e),
-                }
-                .into()
-            }),
-            None => Err(Details::FieldName(String::from(key)).into()),
+        if let Some(position) = self.record_schema.lookup.get(key).copied() {
+            let field = &self.record_schema.fields[position];
```
martin-g: Is it always safe to make this index lookup ?
The previous version uses `.get()`

Kriskras99: Nope, reverted to use `.get()`.

I do think it would be good to make the `fields` and `lookup` fields of `RecordSchema` private and improve the builder.

PR #493, opened 2026-02-28
feat!: Rework `Name` to be more performant

### LINKED ISSUE #492: Why do the `*Validator` traits have the `regex` function?
`SchemaNameValidator`, `SchemaNamespaceValidator`, `EnumSymbolNameValidator` and `RecordFieldNameValidator` all include a `regex` function:
```rust
fn regex(&self) -> &'static Regex {
    ...
}
```
What's the usecase for this? The `validate` functions cover all usecases I can think of.

Kriskras99:
This is achieved by:
- Making `Name` one allocation instead of two
- Returning references as much as possible to reduce cloning
  - `.namespace()` returns a `NamespaceRef<'_>` which is a `Option<&'_ str>`
  - `fully_qualified_name` returns a `Cow<'_, Name>` which reduces allocations when the name already has a namespace or there is enclosing namespace.
- Allow efficient construction of the name when there is an enclosing namespace
- `Display` and `fullname(None)` can now return the full name directly

This required changing the `SchemaNameValidator` trait which now needs to return the index of the starting byte of the name. Included in this change is providing default implementations for `*Validator::validate` functions. This allows users to easily implement the trait by just returning a different regex.

Fixes #492

### DIFF
--- a/avro/src/error.rs
+++ b/avro/src/error.rs
@@ -639,7 +639,7 @@ pub enum Details {
     MapNoKey,
 
     #[error(
-        "The implementation of `SchemaNameValidator` is incorrect, it returned an out-of-bounds index or provided a regex that did not capture a group named `name`"
+        "The implementation of `SchemaNameValidator` is incorrect! It returned an out-of-bounds index or provided a regex that did not capture a group named `name`"
     )]
     InvalidSchemaNameValidatorImplementation,
 }


avro/src/validator.rs
```diff
@@ -77,25 +86,26 @@ pub trait SchemaNameValidator: Send + Sync {
...
             .captures(schema_name)
             .ok_or_else(|| Details::InvalidSchemaName(schema_name.to_string(), regex.as_str()))?;
-        Ok((
-            caps["name"].to_string(),
-            caps.name("namespace").map(|s| s.as_str().to_string()),
-        ))
+        Ok(caps
+            .name("name")
+            .expect("Regex has no group named `name`")
+            .start())
```
martin-g, line 103: return an error when there is no `name` group instead of panicking
```suggestion
        caps.name("name")
            .map(|m| m.start())
            .ok_or_else(|| Details::InvalidSchemaName(schema_name.to_string(), regex.as_str()).into())
```

avro/src/schema/name.rs
```diff
@@ -126,22 +149,39 @@ impl Name {
...
+
+    /// Create an empty name.
+    ///
+    /// This name is invalid and should never be used anywhere! The only valid use is filling
+    /// a `Name` field that will not be used.
+    ///
+    /// Using this name will cause a panic.
+    pub(crate) fn invalid_empty_name() -> Self {
+        Self {
+            namespace_and_name: String::new(),
```
martin-g, line 186:
```suggestion
            namespace_and_name: String::from("invalid_empty_name"),
```
for easier debugging if a Name created with this helper method is ever used as Debug/Display target

avro/src/schema/name.rs
```diff
@@ -52,72 +57,90 @@ pub type Names = HashMap<Name, Schema>;
...
+        enclosing_namespace: NamespaceRef,
+    ) -> AvroResult<Self> {
+        // Having both `Into<String>` and `AsRef<str>` allows optimal use in both of these cases:
+        // - `name` is a `String`. We can reuse the allocation when `enclosing_namespace` is `None`
+        //   or `name` already has a namespace.
+        // - `name` is a `str`. With only `Into<String` we need an extra allocation in the case `name`
+        //   doesn't have namespace and `enclosing_namespace` is `Some`. Having `AsRef<str>` allows
+        //   skipping that allocation.
+        let name_ref = name.as_ref();
+        let index_of_name = validate_schema_name(name_ref)?;
```
martin-g, line 83: A custom validator may return an index that would lead to out of bounds error later. Maybe we should add a check to prevent this here.

avro/src/schema/name.rs
Kriskras99: Instead of this, I modified the `Debug` implementation to show `Name(Invalid name!)` instead. I also made `Display` panic when an invalid name is used.
As users cannot create this type, this panic should never happen.

avro/src/error.rs
```diff
@@ -637,6 +637,11 @@ pub enum Details {
 
     #[error("No key for value when serializing a map")]
     MapNoKey,
+
+    #[error(
+        "The implementation of `SchemaNameValidator` is incorrect, it returned an out-of-bounds index or provided a regex that did not capture a group named `name`"
```
```suggestion
        "The implementation of `SchemaNameValidator` is incorrect! It returned an out-of-bounds index or provided a regex that did not capture a group named `name`"
```

PR #496, opened 2026-03-03
feat: Replace `from_avro_datum*` functions with `GenericDatumReader`

Kriskras99:
The `from_avro_datum*` functions have been deprecated and point to `GenericDatumReader`.

I've also made the `reader` module public so that users need to import `GenericDatumReader` from `reader::datum`. This should further discourage users from using this.

I also want to this for the `to_avro_datum*` functions, but I'll first make a PR making `writer.rs` a directory module.

### DIFF
--- a/avro/src/reader/datum.rs
+++ b/avro/src/reader/datum.rs
@@ -34,7 +34,7 @@ pub struct GenericDatumReader<'s> {
 
 #[bon]
 impl<'s> GenericDatumReader<'s> {
-    /// Build a [`DatumReader`].
+    /// Build a [`GenericDatumReader`].
     ///
     /// This is most likely not what you need. Most users should use [`Reader`][crate::Reader],
     /// [`GenericSingleObjectReader`][crate::GenericSingleObjectReader], or
@@ -144,7 +144,7 @@ impl<'s> GenericDatumReader<'s> {
 /// **NOTE** This function has a quite small niche of usage and does NOT take care of reading the
 /// header and consecutive data blocks; use [`Reader`](struct.Reader.html) if you don't know what
 /// you are doing, instead.
-#[deprecated(since = "0.22.0", note = "Use `DatumReader` instead")]
+#[deprecated(since = "0.22.0", note = "Use `GenericDatumReader` instead")]
 pub fn from_avro_datum<R: Read>(
     writer_schema: &Schema,
     reader: &mut R,
@@ -173,7 +173,7 @@ pub fn from_avro_datum<R: Read>(
 /// schemata to resolve any dependencies.
 ///
 /// When a reader `Schema` is provided, schema resolution will also be performed.
-#[deprecated(since = "0.22.0", note = "Use `DatumReader` instead")]
+#[deprecated(since = "0.22.0", note = "Use `GenericDatumReader` instead")]
 pub fn from_avro_datum_schemata<R: Read>(
     writer_schema: &Schema,
     writer_schemata: Vec<&Schema>,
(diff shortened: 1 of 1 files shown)


avro/src/reader/datum.rs
```diff
@@ -0,0 +1,351 @@
...
+///    .read_value(reader)
+/// ```
+///
+/// Decode a `Value` from raw Avro data.
+///
+/// If the writer schema is incomplete, i.e. contains `Schema::Ref`s then it will use the provided
+/// schemata to resolve any dependencies.
+///
+/// When a reader `Schema` is provided, schema resolution will also be performed.
+#[deprecated(since = "0.22.0", note = "Use `DatumReader` instead")]
```
martin-g:
```suggestion
#[deprecated(since = "0.22.0", note = "Use `GenericDatumReader` instead")]
```

avro/src/reader/datum.rs
```diff
@@ -0,0 +1,351 @@
...
```
martin-g:
```suggestion
#[deprecated(since = "0.22.0", note = "Use `GenericDatumReader` instead")]
```

avro/src/reader/datum.rs
```diff
@@ -0,0 +1,351 @@
...
+/// [`SpecificSingleObjectReader`][crate::SpecificSingleObjectReader] instead.
+pub struct GenericDatumReader<'s> {
+    writer: &'s Schema,
+    resolved: ResolvedSchema<'s>,
+    reader: Option<(&'s Schema, ResolvedSchema<'s>)>,
+}
+
+#[bon]
+impl<'s> GenericDatumReader<'s> {
+    /// Build a [`DatumReader`].
```
martin-g:
```suggestion
    /// Build a [`GenericDatumReader`].
```

avro/src/reader/datum.rs
```diff
@@ -0,0 +1,351 @@
...
+///
+/// Decode a `Value` encoded in Avro format given its `Schema` and anything implementing `io::Read`
+/// to read from.
+///
+/// In case a reader `Schema` is provided, schema resolution will also be performed.
+///
+/// **NOTE** This function has a quite small niche of usage and does NOT take care of reading the
+/// header and consecutive data blocks; use [`Reader`](struct.Reader.html) if you don't know what
+/// you are doing, instead.
+#[deprecated(since = "0.22.0", note = "Use `DatumReader` instead")]
```
martin-g:
```suggestion
#[deprecated(since = "0.22.0", note = "Use `GenericDatumReader` instead")]
```

avro/src/reader/datum.rs
```diff
@@ -0,0 +1,351 @@
...
+    writer_schema: &Schema,
+    writer_schemata: Vec<&Schema>,
+    reader: &mut R,
+    reader_schema: Option<&Schema>,
+    reader_schemata: Vec<&Schema>,
+) -> AvroResult<Value> {
+    GenericDatumReader::builder(writer_schema)
+        .writer_schemata(writer_schemata)?
+        .maybe_reader_schema(reader_schema)
+        .reader_schemata(reader_schemata)?
```
martin-g, line 219: Is it valid to call `.reader_schemata(reader_schemata)?` here if `reader_schema` is None ?
https://github.com/apache/avro-rs/pull/496/changes#diff-503ff4f8f9b911572b84be5ea91a1d2127a77cc40931bc89d3ff105283916f3cR110 has an `IsSet` constraint but I am sure this helps here after the call to `.maybe_reader_schema(None)`

Kriskras99: It's not a problem, the value will just be ignored. I put the extra bound on the `reader_schemata` function just to make it easier for users to do the right thing.

PR #499, opened 2026-03-03
feat: Replace `to_datum*` functions with `GenericDatumWriter`

Kriskras99:
This is the sibling to #496

### DIFF
--- a/avro/src/types.rs
+++ b/avro/src/types.rs
@@ -380,10 +380,8 @@ impl Value {
         let rs = ResolvedSchema::try_from(schemata.clone())
             .expect("Schemata didn't successfully resolve");
         let schemata_len = schemata.len();
-        schemata.iter().any(|schema| {
-            let enclosing_namespace = schema.namespace();
-
-            match self.validate_internal(schema, rs.get_names(), enclosing_namespace) {
+        schemata.iter().any(
+            |schema| match self.validate_internal(schema, rs.get_names(), None) {
                 Some(reason) => {
                     let log_message =
                         format!("Invalid value: {self:?} for schema: {schema:?}. Reason: {reason}");
@@ -395,8 +393,8 @@ impl Value {
                     false
                 }
                 None => true,
-            }
-        })
+            },
+        )
     }
 
     fn accumulate(accumulator: Option<String>, other: Option<String>) -> Option<String> {


avro/src/writer/datum.rs
```diff
@@ -15,28 +15,123 @@
...
+        Ok(vec)
+    }
+
+    pub fn write_ser<W: Write, T: Serialize>(
+        &self,
+        writer: &mut W,
+        value: &T,
+    ) -> AvroResult<usize> {
+        let mut serializer =
+            SchemaAwareWriteSerializer::new(writer, self.schema, self.resolved.get_names(), None);
```
martin-g, line 132: At many places we pass `None` as enclosing namespace. 
We should use the schema's namespace instead.
The issue is that their types are different: Namespace (`Option<String>`) vs. NamespaceRef (`Option<&str>`)

avro/src/writer/datum.rs
```diff
@@ -15,28 +15,123 @@
...
+        writer: &mut W,
+        value: V,
+    ) -> AvroResult<usize> {
+        let value = value.into();
+        self.write_value_ref(writer, &value)
+    }
+
+    pub fn write_value_ref<W: Write>(&self, writer: &mut W, value: &Value) -> AvroResult<usize> {
+        if self.validate
+            && self.resolved.get_schemata().iter().all(|s| {
```
martin-g: Why the value is validated against all schemas ?
It should validate against `self.schema` and pass the Names of the resolved schemas.

avro/src/writer/datum.rs
```diff
@@ -15,28 +15,123 @@
...
+        encode_internal(value, self.schema, self.resolved.get_names(), None, writer)
+    }
+
+    /// Write a value to a [`Vec`].
+    pub fn write_value_to_vec<V: Into<Value>>(&self, value: V) -> AvroResult<Vec<u8>> {
+        let mut vec = Vec::new();
+        self.write_value(&mut vec, value)?;
+        Ok(vec)
+    }
+
```
martin-g, line 124: Let's add some docs for the pub methods.

avro/src/writer/datum.rs
Kriskras99: I mistakenly looked at `Value::validate_schemata`.

avro/src/writer/datum.rs
Kriskras99: I don't think that's right. In my mental model, the `enclosing_namespace` is the namespace of the parent schema of the schema currently being used.

Looking at the code for `SchemaAwareWriteSerializer` and `Value::validate_internal` this is exactly how it is used.

Kriskras99: I've added a fix for `Value::validate_schemata` that does not set the `enclosing_namespace`.

PR #504, opened 2026-03-05
chore: Detect MSRV builds and run derive UI tests

martin-g:
Idea to detect MSRV builds:
- use `std::env::var("CARGO_PKG_RUST_VERSION")` to read the MSRV
- use `rust_version::version().unwrap().to_string();`to read the rustc version
- export `msrv` cfg if both match
- use `cfg!(msrv)` to build conditionally

The UI tests are re-generated with `TRYBUILD=overwrite`. Their output is less informative than `nightly` though.

### DIFF
--- a/avro_derive/tests/ui.rs
+++ b/avro_derive/tests/ui.rs
@@ -18,7 +18,7 @@
 /// These tests only run on nightly as the output can change per compiler version.
 ///
 /// See <https://github.com/dtolnay/trybuild/issues/84>
-#[rustversion::attr(not(nightly), ignore)]
+#[cfg_attr(not(msrv), ignore)]
 #[test]
 fn ui() {
     let t = trybuild::TestCases::new();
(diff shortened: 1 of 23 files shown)


avro_derive/tests/ui.rs
```diff
@@ -18,7 +18,7 @@
 /// These tests only run on nightly as the output can change per compiler version.
 ///
 /// See <https://github.com/dtolnay/trybuild/issues/84>
-#[rustversion::attr(not(nightly), ignore)]
+#[cfg_attr(not(msrv), ignore)]
```
Kriskras99, line 21: This does mean that a developer will have to install the MSRV to run the UI tests, I think it's more likely that developers will have nightly installed (or even enabled by default).

martin-g: Right!
This is inconvenient!
Let's stick with nightly!

PR #509, opened 2026-03-07
fix: Remove `MapSchema` and `ArraySchema` `default` fields

Kriskras99:
Partially reverts #467 as it was accidentally added to spec and removed again.

### DIFF
--- a/avro/src/schema/mod.rs
+++ b/avro/src/schema/mod.rs
@@ -5096,11 +5096,10 @@ mod tests {
         )?;
 
         let Schema::Record(record) = schema else {
-            panic!("Expect Schema::Record for {schema:?}");
+            panic!("Expected Schema::Record, got {schema:?}");
         };
-        println!("{:?}", record);
         let Schema::Map(map) = &record.fields[0].schema else {
-            panic!("Expect Schema::Map for first field of {record:?}");
+            panic!("Expected Schema::Map for first field of {record:?}");
         };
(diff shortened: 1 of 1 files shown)


avro/src/schema/mod.rs
```diff
@@ -5102,289 +5061,62 @@ mod tests {
...
         }"#,
         )?;
 
         let Schema::Record(record) = schema else {
-            panic!("Expected Schema::Record, got {schema:?}");
+            panic!("Expect Schema::Record for {schema:?}");
         };
-        let Schema::Map(map) = &record.fields[1].schema else {
-            panic!("Expected Schema::Map, got {:?}", record.fields[1].schema);
+        println!("{:?}", record);
```
martin-g: debug leftover
```suggestion
```

avro/src/schema/mod.rs
```diff
@@ -5102,289 +5061,62 @@ mod tests {
...
-                }
+                {"name": "a", "type": { "type": "map", "values": "string", "default": null }},
+                {"name": "b", "type": { "type": "array", "items": "string", "default": null }}
             ]
         }"#,
         )?;
 
         let Schema::Record(record) = schema else {
-            panic!("Expected Schema::Record, got {schema:?}");
+            panic!("Expect Schema::Record for {schema:?}");
```
martin-g: The old message look better to me.

PR #512, opened 2026-03-20
feat!: Support enums and tuples in `SchemaAwareSerializer` and implement `SchemaAwareDeserializer`

Kriskras99:
I've done my best to make this PR as small as possible, but the serializer and deserializer implementations need to match. So I can't make it smaller than this. Every commit builds, but only the last commit is guaranteed to pass tests.

This is a breaking change, as this makes the serializer as strict as possible (checking names). An array is now also serialized as a tuple (like Serde does) unless `apache_avro::serde::array` is used.

Other stuff that happens in this PR:
- Document the mapping between the Serde and Avro data models
- `human_writable` is now configurable per reader and writer instance
- `Array` and `Map` serialisation will not write the block size in bytes unless configured to do so
  - This means they don't need to buffer the data, causing performance improvements
- Fix the `AvroSchemaComponent` implementation of `std::time::Duration`
  - The original implementation didn't work as the `Serialize` implementation for this type is a struct
- Fix the `AvroSchemaComponent` for `bigdecimal::BigDecimal`
  - The `Serialize` implementation for `BigDecimal` is a string, which had a workaround in the serializer to make it serialize to `Schema::BigDecimal`. Now the default is `Schema::String` which can be overwritten using `apache_avro::serde::bigdecimal` 

This PR does not include support in `avro_schema_derive`, that will come in a separate PR. I also have a deserializer implementation in the works that can resolve schemas, which should allow use the completely deprecate `from_value` and `to_value`.


avro/src/serde/deser_schema/mod.rs
```diff
@@ -0,0 +1,1688 @@
...
+
+    fn deserialize_u128<V>(mut self, visitor: V) -> Result<V::Value, Self::Error>
+    where
+        V: Visitor<'de>,
+    {
+        match self.schema {
+            Schema::Fixed(fixed) if fixed.size == 16 && fixed.name.name() == "u128" => {
+                visitor.visit_u128(u128::from_le_bytes(self.read_array()?))
+            }
+            Schema::Union(union) => self.with_union(union)?.deserialize_i128(visitor),
```
martin-g:
```suggestion
            Schema::Union(union) => self.with_union(union)?.deserialize_u128(visitor),
```
copy/paste error ?

avro/src/serde/ser_schema/block.rs
```diff
@@ -0,0 +1,361 @@
...
+impl<'s, 'w, W: Write, S: Borrow<Schema>> SerializeSeq for BufferedBlockSerializer<'s, 'w, W, S> {
+    type Ok = usize;
+    type Error = Error;
+
+    fn serialize_element<T>(&mut self, value: &T) -> Result<(), Self::Error>
+    where
+        T: ?Sized + Serialize,
+    {
+        value.serialize(SchemaAwareSerializer::new(
+            self.writer,
```
martin-g: Shouldn't this use the buffer ?
```suggestion
            &mut self.buffer,
```

avro/src/serde/deser_schema/tuple.rs
```diff
@@ -0,0 +1,126 @@
...
+                self.config,
+            )?)
+            .map(Some)
+        } else {
+            Ok(None)
+        }
+    }
+
+    fn size_hint(&self) -> Option<usize> {
+        todo!()
```
martin-g:
```suggestion
        Some(self.schema.fields.len().saturating_sub(self.current_field))
```

avro/src/reader/block.rs
```diff
@@ -205,6 +215,43 @@ impl<'r, R: Read> Block<'r, R> {
...
+            if self.is_empty() {
+                return Ok(None);
+            }
+        }
+
+        let mut block_bytes = &self.buf[self.buf_idx..];
+        let b_original = block_bytes.len();
+
+        let item = if read_schema.is_some() {
+            todo!("Schema aware deserialisation does not resolve schemas yet");
```
martin-g: Is this a todo for the current PR or for a follow-up ?

avro/src/reader/mod.rs
```diff
@@ -113,6 +125,10 @@ impl<'a, R: Read> Reader<'a, R> {
 
         self.block.read_next(read_schema)
     }
+
+    fn read_next_deser<T: DeserializeOwned>(&mut self) -> AvroResult<Option<T>> {
+        self.block.read_next_deser(self.reader_schema)
```
martin-g, line 140: Should this also use `if self.should_resolve_schema {}` as ::read_next() does above ?

avro/src/reader/datum.rs
```diff
@@ -124,6 +137,68 @@ impl<'s> GenericDatumReader<'s> {
...
+
+    /// Read a Avro datum from the reader.
+    ///
+    /// # Panics
+    /// Will panic if a reader schema has been configured, this is a WIP.
+    pub fn read_deser<T: DeserializeOwned>(&self, reader: &mut impl Read) -> AvroResult<T> {
+        // `reader` is `impl Read` instead of a generic on the function like T so it's easier to
+        // specify the type wanted (`read_deser<String>` vs `read_deser<String, _>`)
+        if let Some((_, _)) = &self.reader {
+            todo!("Schema aware deserialisation does not resolve schemas yet");
```
martin-g: Is this a todo for the current PR or for a follow-up ?

avro/src/documentation/serde_data_model_to_avro.rs
```diff
@@ -0,0 +1,116 @@
...
+//! - **byte array** => [`Schema::Bytes`] or [`Schema::Fixed`]
+//! - **option** => [`Schema::Union([Schema::Null, _])`](crate::schema::Schema::Union)
+//! - **unit** => [`Schema::Null`]
+//! - **unit struct** => [`Schema::Record`] with the unqualified name equal to the name of the struct and zero fields
+//! - **unit variant** => See [Enums](##Enums)
+//! - **newtype struct** => [`Schema::Record`] with the unqualified name equal to the name of the struct and one field
+//! - **newtype variant** => See [Enums](##Enums)
+//! - **seq** => [`Schema::Array`]
+//! - **tuple**
+//!     - For tuples with one element, the schema will be the schema the only element
```
martin-g:
```suggestion
//!     - For tuples with one element, the schema will be the schema of the only element
```

avro/src/reader/block.rs
```diff
@@ -205,6 +215,43 @@ impl<'r, R: Read> Block<'r, R> {
...
+            T::deserialize(SchemaAwareDeserializer::new(
+                &mut block_bytes,
+                &self.writer_schema,
+                config,
+            )?)?
+        };
+
+        if b_original != 0 && b_original == block_bytes.len() {
+            // No bytes were read, return an error to avoid an infinite loop
+            return Err(Details::ReadBlock.into());
```
martin-g, line 249: Let's add some more context information to this error. 
As a user I won't be sure what to do.

avro/src/documentation/serde_data_model_to_avro.rs
```diff
@@ -0,0 +1,116 @@
...
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+//! # Mapping the Serde data model to the Avro data model
+//!
+//! When manually mapping Rust types to an Avro schema, or the reverse, it is important to understand
+//! how the different data models are mapped. When mapping from an Avro schema to Rust types,
+//! see [the documentation for the reverse](super::serde_data_model_to_avro).
```
martin-g: this seems to point to the current module/document
```suggestion
//! see [the documentation for the reverse](super::avro_data_model_to_serde).
```

avro/src/serde/ser_schema/union.rs
```diff
@@ -0,0 +1,566 @@
...
+        self.checked_write_int("u16", i32::from(v))
+    }
+
+    fn serialize_u32(mut self, v: u32) -> Result<Self::Ok, Self::Error> {
+        self.checked_write_long("u32", i64::from(v))
+    }
+
+    fn serialize_u64(mut self, v: u64) -> Result<Self::Ok, Self::Error> {
+        match self.union.find_named_schema("u64", self.config.names)? {
+            Some((index, Schema::Fixed(FixedSchema { size: 16, .. }))) => {
```
martin-g:
```suggestion
            Some((index, Schema::Fixed(FixedSchema { size: 8, .. }))) => {
```

avro/src/error.rs
```diff
@@ -156,7 +157,6 @@ pub enum Details {
     #[error("Union index {index} out of bounds: {num_variants}")]
     GetUnionVariant { index: i64, num_variants: usize },
 
-    #[deprecated(since = "0.20.0", note = "This error variant is not generated anymore")]
     #[error("Enum symbol index out of bounds: {num_variants}")]
```
martin-g: let's add `index` to the error message

avro/src/serde/ser_schema/tuple.rs
```diff
@@ -0,0 +1,283 @@
...
+            self.writer,
+            schema,
+            self.config,
+        )?)?;
+        self.field_position += 1;
+        Ok(())
+    }
+
+    fn end(self) -> Result<Self::Ok, Self::Error> {
+        assert_eq!(self.field_position, self.schema.fields.len());
```
martin-g: Why assert_eq!(), i.e. a possible panic ? 
Why not an Err or at least a debug_assert_eq!()?

avro/src/documentation/avro_data_model_to_serde.rs
```diff
@@ -0,0 +1,54 @@
...
+// specific language governing permissions and limitations
+// under the License.
+
+//! # Mapping the Avro data model to the Serde data model
+//!
+//! When manually mapping an Avro schema to Rust types it is important to understand
+//! how the different data models are mapped. When mapping from Rust types to an Avro schema,
+//! see [the documentation for the reverse](super::serde_data_model_to_avro).
+//!
+//! Only the the mapping as defined here is supported. Any other behavior might change in a minor version.
```
martin-g:
```suggestion
//! Only the mapping as defined here is supported. Any other behavior might change in a minor version.
```

avro/src/documentation/avro_data_model_to_serde.rs
```diff
@@ -0,0 +1,54 @@
...
+//! - `double`: [`f64`]
+//! - `bytes`: [`Vec<u8>`](std::vec::Vec) (or any type that uses [`Serialize::serialize_bytes`](serde::Serialize), [`Deserialize::deserialize_bytes`](serde::Deserialize), [`Deserialize::deserialize_byte_buf`](serde::Deserialize))
+//!     - It is required to use [`apache_avro::serde::bytes`] as otherwise Serde will (de)serialize a `Vec` as an array of integers instead.
+//! - `string`: [`String`] (or any type that uses [`Serialize::serialize_str`](serde::Serialize), [`Deserialize::deserialize_str`](serde::Deserialize), [`Deserialize::deserialize_string`](serde::Deserialize))
+//!
+//! ## Complex Types
+//! - `records`: A struct with the same name and fields or a tuple with the same fields.
+//!     - Extra fields can be added to the struct if they are marked with `#[serde(skip)]`
+//! - `enums`: A enum with the same name and unit variants for every symbol
+//!     - The index of the symbol most match the index of the variant
```
martin-g:
```suggestion
//!     - The index of the symbol must match the index of the variant
```

avro/src/schema/mod.rs
```diff
@@ -784,6 +785,69 @@ impl Schema {
...
+                inner: InnerDecimalSchema::Fixed(FixedSchema { name, .. }),
+                ..
+            })
+            | Schema::Uuid(UuidSchema::Fixed(FixedSchema { name, .. }))
+            | Schema::Duration(FixedSchema { name, .. })
+            | Schema::Ref { name } => {
+                let name: String = name
+                    .to_string()
+                    .chars()
+                    .map(|c| if c.is_ascii_alphanumeric() { c } else { '_' })
```
martin-g, line 847: IMO `_` should be cared specially here.
Currently two distinct names as `a.b.c` and `a_b.c` will lead to `r5_a_b_c`.

avro/src/schema/union.rs
```diff
@@ -74,7 +81,116 @@ impl UnionSchema {
...
+                _ => {}
+            }
+        }
+        Ok(None)
+    }
+
+    /// Find a [`Schema::Record`] with `n` fields.
+    ///
+    /// Will use `names` to resolve references.
+    pub(crate) fn find_record_with_n_fields<'s>(
```
martin-g, line 170: What is the idea here ?
If a union schema has several record schemas with the same number of fields but with different types of these fields then the first one wins.
Why do we need this ?

avro/src/serde/deser_schema/map.rs
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    fn read_block_header(reader: &mut R) -> Result<Option<u32>, Error> {
+        let remaining = zag_i32(reader)?;
+        if remaining < 0 {
+            // If the block size is negative the next number is the size of the block in bytes
+            let _bytes = zag_i64(reader)?;
+        }
+        if remaining == 0 {
+            // If the block size is zero the array is finished
```
martin-g:
```suggestion
            // If the block size is zero the map is finished
```

avro/src/serde/mod.rs
```diff
@@ -27,7 +27,8 @@
 //! details on how to change the generated schema.
 //!
 //! Alternatively, you can write your own schema. If you go down this path, it is recommended you start with
-//! the schema derived by [`AvroSchema`] and then modify it to fit your needs.
+//! the schema derived by [`AvroSchema`] and then modify it to fit your needs. For more information on mapping
+//! between Sere and Avro see [Avro to Serde](crate::documentation::avro_data_model_to_serde) and [Serde to Avro](crate::documentation::serde_data_model_to_avro).
```
martin-g:
```suggestion
//! between Serde and Avro see [Avro to Serde](crate::documentation::avro_data_model_to_serde) and [Serde to Avro](crate::documentation::serde_data_model_to_avro).
```

avro/src/serde/ser_schema/block.rs
```diff
@@ -0,0 +1,361 @@
...
+impl<'s, 'w, W: Write, S: Borrow<Schema>> SerializeMap for BufferedBlockSerializer<'s, 'w, W, S> {
+    type Ok = usize;
+    type Error = Error;
+
+    fn serialize_key<T>(&mut self, key: &T) -> Result<(), Self::Error>
+    where
+        T: ?Sized + Serialize,
+    {
+        key.serialize(SchemaAwareSerializer::new(
+            self.writer,
```
martin-g: Shouldn't this use the buffer ?
```suggestion
            &mut self.buffer,
```

avro/src/serde/ser_schema/record/field_default.rs
```diff
@@ -0,0 +1,219 @@
...
+                .values()
+                .all(|value| Self::recursive_type_check(value, &map.types)),
+            (Value::Array(a), Schema::Array(array)) => a
+                .iter()
+                .all(|value| Self::recursive_type_check(value, &array.items)),
+            (_, Schema::Union(union)) => union
+                .variants()
+                .iter()
+                .any(|variant| Self::recursive_type_check(value, variant)),
+            _ => false,
```
martin-g, line 111: What about the logical types with inner type, e.g. Uuid, Decimal & Duration) ?

avro/src/serde/ser_schema/record/field_default.rs
```diff
@@ -0,0 +1,219 @@
...
+                Schema::Long
+                | Schema::TimeMicros
+                | Schema::TimestampMillis
+                | Schema::TimestampMicros
+                | Schema::TimestampNanos
+                | Schema::LocalTimestampMillis
+                | Schema::LocalTimestampMicros
+                | Schema::LocalTimestampNanos,
+            ) if n.is_i64() => true,
+            (Value::Number(n), Schema::Float | Schema::Double) if n.is_f64() => true,
```
martin-g:
```suggestion
            (Value::Number(n), Schema::Float | Schema::Double) if n.as_f64().is_some() => true,
```
to support also integers as defaults for floating point schema types

avro/src/serde/ser_schema/record/field_default.rs
```diff
@@ -0,0 +1,219 @@
...
+                | Schema::TimestampMicros
+                | Schema::TimestampNanos
+                | Schema::LocalTimestampMillis
+                | Schema::LocalTimestampMicros
+                | Schema::LocalTimestampNanos,
+            ) if n.is_i64() => {
+                let long = n.as_i64().unwrap();
+                serializer.serialize_i64(long)
+            }
+            (Value::Number(n), Schema::Float) if n.is_f64() => {
```
martin-g:
```suggestion
            (Value::Number(n), Schema::Float) if n.as_f64().is_some() => {
```

avro/src/serde/ser_schema/record/field_default.rs
```diff
@@ -0,0 +1,219 @@
...
+                let float = n.as_f64().unwrap();
+                serializer.serialize_f32(float as f32)
+            }
+            (Value::Number(n), Schema::Double) if n.is_f64() => {
```
martin-g:
```suggestion
            (Value::Number(n), Schema::Double) if n.as_f64().is_some() => {
```

avro/src/serde/ser_schema/record/field_default.rs
```diff
@@ -0,0 +1,219 @@
...
+                    variant_index as u32,
+                    SERIALIZING_SCHEMA_DEFAULT,
+                )
+            }
+            // This abuses the support for flattened fields, which are also serialized as a map.
+            (Value::Object(o), Schema::Record(record)) => serializer.collect_map(
+                o.iter()
+                    .enumerate()
+                    .map(|(i, (k, v))| (k, Self::new(v, &record.fields[i].schema))),
+            ),
```
martin-g:
```suggestion
            (Value::Object(o), Schema::Record(record)) => serializer.collect_map(
                record.fields.iter().filter_map(|field| {
                    o.get(&field.name)
                        .map(|value| (&field.name, Self::new(value, &field.schema)))
                }),
            ),
```
to serialize the defaults in the schema order

avro/src/serde/ser_schema/record/mod.rs
```diff
@@ -0,0 +1,270 @@
...
+            Ordering::Less => {
+                // Another field needs to be written first, so cache this field
+                let mut bytes = Vec::new();
+                value
+                    .serialize(SchemaAwareSerializer::new(
+                        &mut bytes,
+                        &field.schema,
+                        self.config,
+                    )?)
+                    .map_err(|e| self.field_error(self.field_position, e))?;
```
martin-g:
```suggestion
                    .map_err(|e| self.field_error(position, e))?;
```
Use the requested position for the error, not the other field's position

avro/src/serde/ser_schema/record/mod.rs
```diff
@@ -0,0 +1,270 @@
...
+        }
+    }
+
+    fn serialize_value<T>(&mut self, value: &T) -> Result<(), Self::Error>
+    where
+        T: ?Sized + Serialize,
+    {
+        self.serialize_next_field(
+            self.map_position
+                .expect("serialze_value called without calling serialize_key"),
```
martin-g:
```suggestion
                .expect("serialize_value called without calling serialize_key"),
```

avro/src/writer/datum.rs
```diff
@@ -59,6 +62,19 @@ impl<'s> GenericDatumWriter<'s> {
...
         validate: bool,
+        /// At what block size to start a new block (for arrays and maps).
+        ///
+        /// This is a minimum value, the block size will always be larger than this except for the last
+        /// block.
+        ///
+        /// When set to `None` all values will be written in a single block. This can be faster as no
+        /// intermediate buffer is used, but seeking through written data will be slower.
+        target_block_size: Option<usize>,
+        /// Should [`Serialize`] implementations pick a human readable represenation.
```
martin-g:
```suggestion
        /// Should [`Serialize`] implementations pick a human readable representation.
```

avro/src/writer/mod.rs
```diff
@@ -67,6 +70,19 @@ impl<'a, W: Write> Writer<'a, W> {
         #[builder(default = false)]
         has_header: bool,
         #[builder(default)] user_metadata: HashMap<String, Value>,
+        /// Should [`Serialize`] implementations pick a human readable represenation.
```
martin-g:
```suggestion
        /// Should [`Serialize`] implementations pick a human readable representation.
```

avro/src/writer/single_object.rs
```diff
@@ -86,12 +89,35 @@ impl GenericSingleObjectWriter {
...
+        with = |schema: Schema| -> Result<_, Error> { ResolvedOwnedSchema::new(schema) },
+        default = ResolvedOwnedSchema::new(T::get_schema()).expect("AvroSchema implementation should create valid schemas")
+    )]
     resolved: ResolvedOwnedSchema,
+    #[builder(
+        default = RabinFingerprintHeader::from_schema(resolved.get_root_schema()).build_header(),
+        with = |header_builder: impl HeaderBuilder| header_builder.build_header(),
+    )]
     header: Vec<u8>,
+    /// Should [`Serialize`] implementations pick a human readable represenation.
```
martin-g:
```suggestion
    /// Should [`Serialize`] implementations pick a human readable representation.
```

avro/src/documentation/avro_data_model_to_serde.rs
```diff
@@ -0,0 +1,54 @@
...
+//! - `arrays`: [`Vec`] (or any type that uses [`Serialize::serialize_seq`](serde::Serialize), [`Deserialize::deserialize_seq`](serde::Deserialize))
+//!     - `[T; N]` is (de)serialized as a tuple by Serde, to (de)serialize them as an `array` use [`apache_avro::serde::array`]
+//! - `maps`: [`HashMap<String, _>`](std::collections::HashMap) (or any type that uses [`Serialize::serialize_map`](serde::Serialize), [`Deserialize::deserialize_map`](serde::Deserialize))
+//! - `unions`: A enum with a variant for each variant
```
martin-g:
```suggestion
//! - `unions`: An enum with a variant for each union variant
```

avro/src/documentation/serde_data_model_to_avro.rs
```diff
@@ -0,0 +1,116 @@
...
+//! `tag` attribute. If a field is not used by all variants it must have a `default` set.
+//!
+//! ### Adjacently tagged
+//! This enum representation is used by Serde if the attributes `#[serde(tag = "...", content = "...")]` are used.
+//! It maps to a [`Schema::Record`] with two fields. One field must be named as the value of the `tag`
+//! attribute and use the [`Schema::Enum`] schema. The other field must be named as the value of the
+//! `content` tag and use the [`Schema::Union`] schema.
+//!
+//! ### Untagged
+//! This enum representation is ued by Serde if the attribute `#[serde(untagged)]` is used. It maps
```
martin-g:
```suggestion
//! This enum representation is used by Serde if the attribute `#[serde(untagged)]` is used. It maps
```

avro/src/reader/block.rs
Kriskras99: Follow-up, I've changed it to a panic and a comment. I've also documented on the builder that setting a reader schema is currently incompatible with the deserialisation interface.

avro/src/reader/block.rs
Kriskras99: I've changed the error description from `unable to read block` to `Did not read any bytes, block is corrupt`.
The only way that this can be triggered if the schema is `Schema::Null` and the block actually has data.

avro/src/reader/datum.rs
Kriskras99: Follow-up, I've changed it to a panic and a comment. It's also documented as a panic on `read_deser`

avro/src/schema/mod.rs
Kriskras99: Every `_` is now expended to two `_` and I've added some tests

avro/src/schema/union.rs
Kriskras99: It's used in the serializer for the `tuple` and `map` cases. We don't have a set name for a tuple, so we just use the first record with the same amount of fields.
For `serialize_map` it's used for records with flattened fields, because we don't get the name. This is an improvement on the previous serialize implementation that would just pick the first map/record/reference.

avro/src/serde/deser_schema/mod.rs
Kriskras99: Unfortunately yes

avro/src/serde/ser_schema/record/field_default.rs
Kriskras99: Added

avro/src/serde/ser_schema/record/field_default.rs
Kriskras99: Nice!

avro/src/serde/ser_schema/tuple.rs
Kriskras99: Converted to an error, originally chose a panic because this can only happen for invalid Serialize implementations.

avro/src/serde/deser_schema/mod.rs
```diff
@@ -0,0 +1,1818 @@
...
+        };
+        Ok(self)
+    }
+
+    /// Read the union and create a new deserializer with the existing reader and config.
+    ///
+    /// This will resolve the read schema if it is a reference.
+    fn with_union(self, schema: &'s UnionSchema) -> Result<Self, Error> {
+        let index = zag_i32(self.reader)?;
+        let variant = schema.get_variant(index as usize)?;
```
martin-g:
```suggestion
        let index = zag_i32(self.reader)?;
        let index = usize::try_from(index).map_err(|e| Details::ConvertI32ToUsize(e, index))?;
        let variant = schema.get_variant(index)?;
```

avro/src/serde/ser_schema/block.rs
```diff
@@ -0,0 +1,361 @@
...
+impl<'s, 'w, W: Write, S: Borrow<Schema>> SerializeMap for BufferedBlockSerializer<'s, 'w, W, S> {
+    type Ok = usize;
+    type Error = Error;
+
+    fn serialize_key<T>(&mut self, key: &T) -> Result<(), Self::Error>
+    where
+        T: ?Sized + Serialize,
+    {
+        key.serialize(SchemaAwareSerializer::new(
+            &mut self.writer,
```
martin-g:
```suggestion
            &mut self.buffer,
```
?

avro/src/serde/ser_schema/tuple.rs
```diff
@@ -0,0 +1,290 @@
...
+
+impl<'s, 'w, W: Write, S: Borrow<Schema>> SerializeTuple for ManyTupleSerializer<'s, 'w, W, S> {
+    type Ok = usize;
+    type Error = Error;
+
+    fn serialize_element<T>(&mut self, value: &T) -> Result<(), Self::Error>
+    where
+        T: ?Sized + Serialize,
+    {
+        let schema = &self.schema.fields[self.field_position].schema;
```
martin-g: Let's return an Err if `self.field_position` is out of bounds

avro/src/serde/ser_schema/union.rs
```diff
@@ -0,0 +1,566 @@
...
+    ) -> Result<usize, Error> {
+        if let Some(index) = self.union.index_of_schema_kind(SchemaKind::Int) {
+            let mut bytes_written = zig_i32(index as i32, &mut *self.writer)?;
+            bytes_written += zig_i32(v, &mut *self.writer)?;
+            Ok(bytes_written)
+        } else {
+            Err(self.error(
+                original_ty,
+                "Expected Schema::Int | Schema::Date | Schema::TimeMillis in variants",
+            ))?
```
martin-g:
```suggestion
            ))
```

avro/src/serde/ser_schema/union.rs
```diff
@@ -0,0 +1,566 @@
...
+    type SerializeStruct = RecordSerializer<'s, 'w, W, S>;
+    type SerializeStructVariant = RecordSerializer<'s, 'w, W, S>;
+
+    fn serialize_bool(mut self, v: bool) -> Result<Self::Ok, Self::Error> {
+        if let Some(index) = self.union.index_of_schema_kind(SchemaKind::Boolean) {
+            let mut bytes_written = zig_i32(index as i32, &mut *self.writer)?;
+            bytes_written += self.write_array([u8::from(v)])?;
+            Ok(bytes_written)
+        } else {
+            Err(self.error("bool", "Expected Schema::Boolean in variants"))?
```
martin-g:
```suggestion
            Err(self.error("bool", "Expected Schema::Boolean in variants"))
```

avro/src/serde/ser_schema/union.rs
```diff
@@ -0,0 +1,566 @@
...
+        }
+    }
+
+    fn serialize_f32(mut self, v: f32) -> Result<Self::Ok, Self::Error> {
+        if let Some(index) = self.union.index_of_schema_kind(SchemaKind::Float) {
+            let mut bytes_written = zig_i32(index as i32, &mut *self.writer)?;
+            bytes_written += self.write_array(v.to_le_bytes())?;
+            Ok(bytes_written)
+        } else {
+            Err(self.error("f32", "Expected Schema::Float in variants"))?
```
martin-g:
```suggestion
            Err(self.error("f32", "Expected Schema::Float in variants"))
```

avro/src/serde/ser_schema/union.rs
```diff
@@ -0,0 +1,566 @@
...
+        }
+    }
+
+    fn serialize_f64(mut self, v: f64) -> Result<Self::Ok, Self::Error> {
+        if let Some(index) = self.union.index_of_schema_kind(SchemaKind::Double) {
+            let mut bytes_written = zig_i32(index as i32, &mut *self.writer)?;
+            bytes_written += self.write_array(v.to_le_bytes())?;
+            Ok(bytes_written)
+        } else {
+            Err(self.error("f64", "Expected Schema::Double in variants"))?
```
martin-g:
```suggestion
            Err(self.error("f64", "Expected Schema::Double in variants"))
```

avro/src/serde/ser_schema/union.rs
```diff
@@ -0,0 +1,566 @@
...
+        }
+    }
+
+    fn serialize_char(mut self, v: char) -> Result<Self::Ok, Self::Error> {
+        if let Some(index) = self.union.index_of_schema_kind(SchemaKind::String) {
+            let mut bytes_written = zig_i32(index as i32, &mut *self.writer)?;
+            bytes_written += self.write_bytes_with_len(v.to_string().as_bytes())?;
+            Ok(bytes_written)
+        } else {
+            Err(self.error("char", "Expected Schema::String in variants"))?
```
martin-g:
```suggestion
            Err(self.error("char", "Expected Schema::String in variants"))
```

avro/src/serde/ser_schema/union.rs
```diff
@@ -0,0 +1,566 @@
...
+        }
+    }
+
+    fn serialize_str(mut self, v: &str) -> Result<Self::Ok, Self::Error> {
+        if let Some(index) = self.union.index_of_schema_kind(SchemaKind::String) {
+            let mut bytes_written = zig_i32(index as i32, &mut *self.writer)?;
+            bytes_written += self.write_bytes_with_len(v.as_bytes())?;
+            Ok(bytes_written)
+        } else {
+            Err(self.error("str", "Expected Schema::String in variants"))?
```
martin-g:
```suggestion
            Err(self.error("str", "Expected Schema::String in variants"))
```

avro/src/schema/mod.rs
```diff
@@ -5118,4 +5183,29 @@ mod tests {
...
+        let four = Schema::Ref {
+            name: "a._b._c".parse()?,
+        };
+
+        assert_ne!(one, two);
+        assert_ne!(one, three);
+        assert_ne!(one, four);
+        assert_ne!(two, three);
+        assert_ne!(two, four);
+        assert_ne!(three, four);
```
```suggestion
        assert_ne!(one.unique_normalized_name(), two.unique_normalized_name());
        assert_ne!(one.unique_normalized_name(), three.unique_normalized_name());
        assert_ne!(one.unique_normalized_name(), four.unique_normalized_name());
        assert_ne!(two.unique_normalized_name(), three.unique_normalized_name());
        assert_ne!(two.unique_normalized_name(), four.unique_normalized_name());
        assert_ne!(three.unique_normalized_name(), four.unique_normalized_name());
```

avro/src/serde/ser_schema/block.rs
```diff
@@ -0,0 +1,361 @@
...
+            items_in_buffer: 0,
+            target_block_size,
+        }
+    }
+
+    /// Write a block including the items and size header.
+    fn write_block(&mut self) -> Result<(), Error> {
+        // Write the header, the negative item count indicates that the next value is the size of the
+        // block in bytes
+        self.bytes_written += encode_int(0 - (self.items_in_buffer as i32), &mut *self.writer)?;
```
martin-g: Why `items_in_buffer` is `usize` if we assume that it won't be bigger than i32 ?

avro/src/error.rs
```diff
@@ -553,32 +555,51 @@ pub enum Details {
...
+        schema: RecordSchema,
+    },
+
+    #[error("Failed to serialize field '{field_name}' of record {record_schema:?}: {error}")]
     SerializeRecordFieldWithSchema {
         field_name: String,
-        record_schema: Schema,
-        error: Box<Error>,
+        record_schema: RecordSchema,
+        error: String,
```
martin-g, line 581: Why the source/cause is a String now ?

avro_derive/src/attributes/mod.rs
```diff
@@ -340,6 +340,12 @@ impl FieldOptions {
             return Err(errors);
         }
 
+        // TODO: Implement a better way to do this (maybe if user specifies `#[avro(with)]` also use that for the default)
+        // Disable getting the field default, if the schema is not retrieved from the field type
+        if with != With::Trait && avro.default == FieldDefault::Trait {
+            avro.default = FieldDefault::Disabled;
```
martin-g, line 333: Should this check be made before the `skip_serializing` check at line 321 ?

avro/src/documentation/avro_data_model_to_serde.rs
```diff
@@ -0,0 +1,54 @@
...
+//! Only the mapping as defined here is supported. Any other behavior might change in a minor version.
+//!
+//! ## Primitive Types
+//! - `null`: `()`
+//! - `boolean`: [`bool`]
+//! - `int`: [`i32`] (or [`i16`], [`i8`], [`u16`], [`u8`])
+//! - `long`: [`i64`] (or [`u32`])
+//! - `float`: [`f32`]
+//! - `double`: [`f64`]
+//! - `bytes`: [`Vec<u8>`](std::vec::Vec) (or any type that uses [`Serialize::serialize_bytes`](serde::Serialize), [`Deserialize::deserialize_bytes`](serde::Deserialize), [`Deserialize::deserialize_byte_buf`](serde::Deserialize))
```
martin-g:
```suggestion
//! - `bytes`: [`Vec<u8>`](std::vec::Vec) (or any type that uses [`Serializer::serialize_bytes`](serde::Serializer), [`Deserializer::deserialize_bytes`](serde::Deserializer), [`Deserializer::deserialize_byte_buf`](serde::Deserializer))
```

avro/src/documentation/avro_data_model_to_serde.rs
```diff
@@ -0,0 +1,54 @@
...
+//!     - It is required to use [`apache_avro::serde::bytes`] as otherwise Serde will (de)serialize a `Vec` as an array of integers instead.
+//! - `string`: [`String`] (or any type that uses [`Serialize::serialize_str`](serde::Serialize), [`Deserialize::deserialize_str`](serde::Deserialize), [`Deserialize::deserialize_string`](serde::Deserialize))
```
martin-g:
```suggestion
//! - `string`: [`String`] (or any type that uses [`Serializer::serialize_str`](serde::Serializer), [`Deserializer::deserialize_str`](serde::Deserializer), [`Deserializer::deserialize_string`](serde::Deserializer))
```

avro/src/documentation/avro_data_model_to_serde.rs
```diff
@@ -0,0 +1,54 @@
...
+//!
+//! ## Complex Types
+//! - `records`: A struct with the same name and fields or a tuple with the same fields.
+//!     - Extra fields can be added to the struct if they are marked with `#[serde(skip)]`
+//! - `enums`: A enum with the same name and unit variants for every symbol
+//!     - The index of the symbol must match the index of the variant
+//! - `arrays`: [`Vec`] (or any type that uses [`Serialize::serialize_seq`](serde::Serialize), [`Deserialize::deserialize_seq`](serde::Deserialize))
```
martin-g:
```suggestion
//! - `arrays`: [`Vec`] (or any type that uses [`Serializer::serialize_seq`](serde::Serializer), [`Deserializer::deserialize_seq`](serde::Deserializer))
```

avro/src/documentation/avro_data_model_to_serde.rs
```diff
@@ -0,0 +1,54 @@
...
+//!     - `[T; N]` is (de)serialized as a tuple by Serde, to (de)serialize them as an `array` use [`apache_avro::serde::array`]
+//! - `maps`: [`HashMap<String, _>`](std::collections::HashMap) (or any type that uses [`Serialize::serialize_map`](serde::Serialize), [`Deserialize::deserialize_map`](serde::Deserialize))
```
martin-g:
```suggestion
//! - `maps`: [`HashMap<String, _>`](std::collections::HashMap) (or any type that uses [`Serializer::serialize_map`](serde::Serializer), [`Deserializer::deserialize_map`](serde::Deserializer))
```

avro/src/documentation/avro_data_model_to_serde.rs
```diff
@@ -0,0 +1,54 @@
...
+//! - `unions`: An enum with a variant for each union variant
+//!     - The index of the union variant must match the enum variant
+//!     - A `null` can be a unit variant or a newtype variant with a unit type
+//!     - All other variants must be newtype variants, struct variants, or tuple variants.
+//! - `fixed`: [`Vec<u8>`](std::vec::Vec) (or any type that uses [`Serialize::serialize_bytes`](serde::Serialize), [`Deserialize::deserialize_bytes`](serde::Deserialize), [`Deserialize::deserialize_byte_buf`](serde::Deserialize))
```
martin-g:
```suggestion
//! - `fixed`: [`Vec<u8>`](std::vec::Vec) (or any type that uses [`Serializer::serialize_bytes`](serde::Serializer), [`Deserializer::deserialize_bytes`](serde::Deserializer), [`Deserializer::deserialize_byte_buf`](serde::Deserializer))
```

avro/src/documentation/avro_data_model_to_serde.rs
```diff
@@ -0,0 +1,54 @@
...
+//!     - It is required to use [`apache_avro::serde::bytes`] as otherwise Serde will (de)serialize a `Vec` as an array of integers instead.
```
martin-g:
```suggestion
//!     - It is required to use [`apache_avro::serde::fixed`] as otherwise Serde will (de)serialize a `Vec` as an array of integers instead.
```

avro/src/documentation/serde_data_model_to_avro.rs
```diff
@@ -0,0 +1,116 @@
...
+//! - **tuple**
+//!     - For tuples with one element, the schema will be the schema of the only element
+//!     - For tuples with more than one element, the schema will be a [`Schema::Record`] with as many fields as there are elements.
+//!       The schema must have the attribute `org.apache.avro.rust.tuple` with the value set to `true`.
+//!     - **Note:** Serde (de)serializes arrays (`[T; N]`) as tuples. To (de)serialize an array as a
+//!       [`Schema::Array`] use [`apache_avro::serde::array`].
+//! - **tuple_struct** => [`Schema::Record`] with the unqualified name equal to the name of the struct and as many fields as there are elements
+//!     - **Note:** Tuple structs with 0 or 1 elements will also be (de)serialized as a [`Schema::Record`]. This
+//!       is different from normal tuples`.
+//! - **tuple_variant** => See [Enums](##Enums)
```
martin-g:
```suggestion
//! - **tuple_variant** => See [Enums](#enums)
```

avro/src/documentation/serde_data_model_to_avro.rs
```diff
@@ -0,0 +1,116 @@
...
+//! - **map** => [`Schema::Map`]
+//!     - **Note:** the key type of the map will be (de)serialized as a [`Schema::String`]
+//! - **struct** => [`Schema::Record`]
+//! - **struct_variant** => See [Enums](##Enums)
```
martin-g:
```suggestion
//! - **struct_variant** => See [Enums](#enums)
```

avro/src/documentation/serde_data_model_to_avro.rs
```diff
@@ -0,0 +1,116 @@
...
+//!     - `f32` => [`Schema::Float`]
+//!     - `f64` => [`Schema::Double`]
+//!     - `char` => [`Schema::String`]
+//!         - Only one character allowed, deserializer will return an error for strings with more than one character.
+//! - **string** => [`Schema::String`]
+//! - **byte array** => [`Schema::Bytes`] or [`Schema::Fixed`]
+//! - **option** => [`Schema::Union([Schema::Null, _])`](crate::schema::Schema::Union)
+//! - **unit** => [`Schema::Null`]
+//! - **unit struct** => [`Schema::Record`] with the unqualified name equal to the name of the struct and zero fields
+//! - **unit variant** => See [Enums](##Enums)
```
martin-g:
```suggestion
//! - **unit variant** => See [Enums](#enums)
```

avro/src/documentation/serde_data_model_to_avro.rs
```diff
@@ -0,0 +1,116 @@
...
+//! - **newtype struct** => [`Schema::Record`] with the unqualified name equal to the name of the struct and one field
+//! - **newtype variant** => See [Enums](##Enums)
```
martin-g:
```suggestion
//! - **newtype variant** => See [Enums](#enums)
```

avro/src/schema/mod.rs
martin-g: Both `a._b` and `a_.b` would still produce `a___b`, but I also don't see a way to avoid all duplicates. 
And I doubt someone will use such names.

avro/src/serde/derive.rs
```diff
@@ -630,27 +612,35 @@ where
...
+                        .build(),
+                    // Nanos is an u32
+                    RecordField::builder()
+                        .name("nanos")
+                        .schema(Schema::Long)
+                        .build(),
+                ])
+                .build()
         }
     }
```
martin-g, line 639: `get_record_fields_in_ctxt()` should `Some(vec![records fields])` too

avro/src/serde/deser_schema/array.rs
```diff
@@ -0,0 +1,102 @@
...
+    util::{zag_i32, zag_i64},
+};
+
+/// Deserialize sequences from an Avro array.
+pub struct ArrayDeserializer<'s, 'r, R: Read, S: Borrow<Schema>> {
+    reader: &'r mut R,
+    schema: &'s Schema,
+    config: Config<'s, S>,
+    /// Track where we are in reading the array
+    remaining: Option<u32>,
```
martin-g: https://avro.apache.org/docs/1.12.0/specification/#complex-types-1 says:
```
Arrays are encoded as a series of blocks. Each block consists of a long count value, followed by that many array items. 
```
"**long** count".
Also see https://github.com/apache/avro-rs/blob/6e6da17ade56087ed10b2064ef6817d90196f6a8/avro/src/encode.rs#L238

```suggestion
    remaining: Option<u64>,
```

avro/src/serde/deser_schema/map.rs
```diff
@@ -0,0 +1,150 @@
...
+    schema::MapSchema,
+    serde::deser_schema::{Config, SchemaAwareDeserializer},
+    util::{zag_i32, zag_i64},
+};
+
+pub struct MapDeserializer<'s, 'r, R: Read, S: Borrow<Schema>> {
+    reader: &'r mut R,
+    schema: &'s Schema,
+    config: Config<'s, S>,
+    remaining: Option<u32>,
```
martin-g:
```suggestion
    remaining: Option<u64>,
```
https://avro.apache.org/docs/1.12.0/specification/#complex-types-1 says:
```
Maps are encoded as a series of blocks. Each block consists of a long count value, ...
```
Note: **long**
Also see https://github.com/apache/avro-rs/blob/6e6da17ade56087ed10b2064ef6817d90196f6a8/avro/src/encode.rs#L264

avro/src/serde/deser_schema/record.rs
```diff
@@ -0,0 +1,95 @@
...
+            ))
+            .map(Some)
+        }
+    }
+
+    fn next_value_seed<V>(&mut self, seed: V) -> Result<V::Value, Self::Error>
+    where
+        V: DeserializeSeed<'de>,
+    {
+        let schema = &self.schema.fields[self.current_field].schema;
```
martin-g, line 68: Let's check whether `self.current_field` is not out of bounds.

avro/src/serde/deser_schema/record.rs
```diff
@@ -0,0 +1,95 @@
...
+        let value = if let Schema::Record(record) = schema
```
martin-g, line 74: What about `Schema::Ref` ?

avro/src/serde/ser_schema/union.rs
```diff
@@ -0,0 +1,566 @@
...
+    }
+
+    fn serialize_unit_variant(
+        self,
+        name: &'static str,
+        variant_index: u32,
+        variant: &'static str,
+    ) -> Result<Self::Ok, Self::Error> {
+        match self.union.find_named_schema(name, self.config.names)? {
+            Some((index, Schema::Enum(schema))) if schema.symbols[variant_index as usize] == variant => {
```
martin-g:
```suggestion
            Some((index, Schema::Enum(schema))) if schema.symbols.get(variant_index as usize) == Some(&variant.to_string()) => {
```

avro/src/serde/with.rs
```diff
@@ -503,11 +500,326 @@ pub mod slice_opt {
...
+        S: Serializer,
+        T: Serialize,
+    {
+        value.as_slice().serialize(serializer)
+    }
+
+    pub fn deserialize<'de, const N: usize, D, T>(deserializer: D) -> Result<[T; N], D::Error>
+    where
+        D: Deserializer<'de>,
+        T: DeserializeOwned,
```
martin-g:
```suggestion
        T: Deserialize<'de>,
```
Does this work too ?
array_opt below is OK with it.

avro/src/schema/mod.rs
Kriskras99: True, the only thing I can think of is hashing the names but that will make the names very ugly.

avro/src/serde/deser_schema/record.rs
Kriskras99: It is checked in `next_key_seed` and the Serde docs say we're allowed to panic if `next_value_seed` is called without `next_key_seed`: https://docs.rs/serde/latest/serde/de/trait.MapAccess.html#tymethod.next_value_seed.

avro/src/error.rs
Kriskras99: So I can create a more optimal error string. When an error happens in the following chain: record -> field -> record -> field. The schema will be printed 4 times. This makes the error message almost useless. My modified error (as created in `RecordSerializer::field_error`) will only print the top schema and then just print the actual errors.

avro_derive/src/attributes/mod.rs
Kriskras99: Yes, I also have a better check in the next PR (with the support for deriving schemas for all enums)

avro/src/serde/deser_schema/mod.rs
```diff
@@ -0,0 +1,1817 @@
...
+    fn deserialize_option<V>(self, visitor: V) -> Result<V::Value, Self::Error>
+    where
+        V: Visitor<'de>,
+    {
+        if let Schema::Union(union) = self.schema
+            && union.variants().len() == 2
+            && union.is_nullable()
+        {
+            let index = zag_i32(self.reader)?;
+            let schema = union.get_variant(index as usize)?;
```
martin-g:
```suggestion
            let index = usize::try_from(index).map_err(|e| Details::ConvertI32ToUsize(e, index))?;
            let schema = union.get_variant(index)?;
```

avro/src/serde/ser_schema/record/field_default.rs
```diff
@@ -0,0 +1,237 @@
...
+            }
+            (
+                Value::String(s),
+                Schema::Bytes
+                | Schema::Fixed(_)
+                | Schema::Uuid(UuidSchema::Bytes | UuidSchema::Fixed(_))
+                | Schema::BigDecimal
+                | Schema::Decimal(_)
+                | Schema::Duration(_)
+                | Schema::Date,
```
martin-g: Why Date is here ?

avro/src/documentation/serde_data_model_to_avro.rs
```diff
@@ -0,0 +1,116 @@
...
+//! - **seq** => [`Schema::Array`]
+//! - **tuple**
+//!     - For tuples with one element, the schema will be the schema of the only element
+//!     - For tuples with more than one element, the schema will be a [`Schema::Record`] with as many fields as there are elements.
+//!       The schema must have the attribute `org.apache.avro.rust.tuple` with the value set to `true`.
+//!     - **Note:** Serde (de)serializes arrays (`[T; N]`) as tuples. To (de)serialize an array as a
+//!       [`Schema::Array`] use [`apache_avro::serde::array`].
+//! - **tuple_struct** => [`Schema::Record`] with the unqualified name equal to the name of the struct and as many fields as there are elements
+//!     - **Note:** Tuple structs with 0 or 1 elements will also be (de)serialized as a [`Schema::Record`]. This
+//!       is different from normal tuples`.
```
martin-g:
```suggestion
//!       is different from normal tuples.
```

avro/src/serde/deser_schema/mod.rs
```diff
@@ -0,0 +1,1817 @@
...
+        name: &'static str,
+        len: usize,
+        visitor: V,
+    ) -> Result<V::Value, Self::Error>
+    where
+        V: Visitor<'de>,
+    {
+        match self.schema {
+            Schema::Record(record) if record.name.name() == name && record.fields.len() == len => {
+                visitor.visit_map(RecordDeserializer::new(self.reader, record, self.config))
```
martin-g: Shouldn't this be `visit_seq()` ?

avro/src/serde/deser_schema/mod.rs
```diff
@@ -0,0 +1,1817 @@
...
+        )?;
+
+        let value1 = TestTupleExternalEnum {
+            a: TupleExternalEnum::Val1(1.0, 2.0),
+        };
+
+        assert_roundtrip(value1, &schema, Vec::new())?;
+
+        let value2 = TestTupleExternalEnum {
+            a: TupleExternalEnum::Val1(2.0, 1.0),
```
martin-g: Val2 ?

avro/src/serde/deser_schema/tuple.rs
```diff
@@ -0,0 +1,126 @@
...
+{
+    type Error = Error;
+
+    fn next_element_seed<T>(&mut self, seed: T) -> Result<Option<T::Value>, Self::Error>
+    where
+        T: DeserializeSeed<'de>,
+    {
+        if self.current_field < self.schema.fields.len() {
+            let schema = &self.schema.fields[self.current_field].schema;
+            self.current_field += 1;
```
martin-g: Let's move the incrementation after `seed.deserialize()`, so that it is not incremented if deserialize fails for any reason

avro/src/serde/ser_schema/record/field_default.rs
Kriskras99: Confused it with Schema::Duration

avro/src/serde/deser_schema/mod.rs
Kriskras99: Yes, it shouldn't even use RecordDeserializer but ManyTupleDeserializer.

avro/src/reader/block.rs
```diff
@@ -205,6 +215,44 @@ impl<'r, R: Read> Block<'r, R> {
         Ok(Some(item))
     }
 
+    pub(super) fn read_next_deser<T: DeserializeOwned>(
+        &mut self,
+        read_schema: Option<&Schema>,
```
martin-g:
```suggestion
        reader_schema: Option<&Schema>,
```
for consistency

avro/src/documentation/serde_data_model_to_avro.rs
```diff
@@ -0,0 +1,116 @@
...
+//!
+//! ### Externally tagged
+//! This is the default enum representation for Serde. It can be mapped in three ways to the Avro data model.
+//! For all three options it is important that the enum index matches the Avro index.
+//! - As a [`Schema::Enum`], this is only possible for enums with only unit variants.
+//! - As a [`Schema::Union`] with a [`Schema::Record`] for every variant:
+//!     - **unit_variant** => [`Schema::Record`] named as the variant and with no fields.
+//!     - **newtype_variant** => [`Schema::Record`] named as the variant and with one field.
+//!       The schema must have the attribute `org.apache.avro.rust.union_of_records` with the value set to `true`.
+//!     - **tuple_variant** => [`Schema::Record`] named as the variant and with as many fields as there are element.
```
martin-g:
```suggestion
//!     - **tuple_variant** => [`Schema::Record`] named as the variant and with as many fields as there are elements.
```

avro/src/reader/single_object.rs
```diff
@@ -315,11 +354,11 @@ mod tests {
...
-        )
-        .expect("failed to build reader");
+        let generic_reader = GenericSingleObjectReader::builder()
+            .schema(TestSingleObjectReader::get_schema())
+            .header(header_builder.build_header())
+            .build()
+            .expect("failed to build reader");
         let data_to_read: Vec<u8> = vec![
             3, 0, 178, 241, 207, 0, 4, 52, 1, 62, 67, 154, 18, 94, 184, 72, 90, 95,
         ];
```
martin-g, line 357: The `matches!(read_result, Err(Details::ReadBytes(_)));` below should probably be wrapped with `assert!(...)`
GitHub UI does not allow me to add a suggestion

avro/src/reader/single_object.rs
```diff
@@ -68,6 +79,26 @@ impl GenericSingleObjectReader {
             Err(io_error) => Err(Details::ReadHeader(io_error).into()),
         }
     }
+
+    pub fn read_deser<T: DeserializeOwned>(&self, reader: &mut impl Read) -> AvroResult<T> {
+        let mut header = vec![0; self.expected_header.len()];
```
martin-g, line 85: The header reading and validation logic is the same in `read_value()` and `read_deser()`. It could be extracted to a helper function

avro/src/schema/union.rs
```diff
@@ -74,7 +81,116 @@ impl UnionSchema {
...
+    /// Will use `names` to resolve references.
+    pub(crate) fn find_named_schema<'s>(
+        &'s self,
+        name: &str,
+        names: &'s HashMap<Name, impl Borrow<Schema>>,
+    ) -> Result<Option<(usize, &'s Schema)>, Error> {
+        for index in self.named_index.iter().copied() {
+            let schema = &self.schemas[index];
+            if let Some(schema_name) = schema.name()
+                && schema_name.name() == name
```
martin-g, line 115: Shouldn't this use the fully qualified names ?

avro/src/serde/with.rs
```diff
@@ -503,11 +502,323 @@ pub mod slice_opt {
...
+    }
+}
+
+/// (De)serialize an optional Rust array (`Option<[T; N]>`) as an Avro `Schema::Union([Schema::Null, Schema::Array])`.
+///
+/// This module is intended to be used through the Serde `with` attribute.
+///
+/// Use [`apache_avro::serde::array`] for non-optional array values.
+///
+/// When used with different serialization formats, this is equivalent to [`serde_bytes`].
```
martin-g: Is this copy/pasted ? It sounds unrelated here.

avro/src/schema/union.rs
Kriskras99: No, it is used for matching Rust type names to schemas which don't include the namespace. I've updated the documentation to make that more explicit.

avro/src/reader/single_object.rs
Kriskras99: This simplifies the code significantly, good suggestion!

avro/src/serde/with.rs
Kriskras99: Yes, sorry

avro/src/reader/single_object.rs
Kriskras99: Adding a assert makes the test fail, I've reworked the test to actually test being able to deserialize a value wih a different header format.

avro/src/reader/single_object.rs
```diff
@@ -315,19 +345,26 @@ mod tests {
...
-            TestSingleObjectReader::get_schema(),
-            header_builder,
-        )
-        .expect("failed to build reader");
+        let generic_reader = GenericSingleObjectReader::builder()
+            .schema(TestSingleObjectReader::get_schema())
+            .header(header_builder.build_header())
+            .build()
+            .expect("failed to build reader");
+        // First 18 bytes are the header, than it's a varint 0, double 0 and an empty list (0)
```
martin-g:
```suggestion
        // First 18 bytes are the header, then it's a varint 0, double 0 and an empty list (0)
```

martin-g (approved):

Kriskras99:
Thank you so much for the time and effort you put into reviewing this PR!

kevinjqliu:
Hey @Kriskras99 thanks for adding this! This is exactly what I'm looking for 😄 

It works really well for the upstream `iceberg-rust` manifest reader, since we already know the target Rust types and don’t need to go through `Value` first. Nice to be able to keep using `apache-avro` here without the extra conversion step.

Kriskras99:
@kevinjqliu that's great to hear! Please do let me know if you run into any issues or find anything lacking in the documentation.

Issue #517, opened 2026-03-25
Support flattening `Option<T>` into the `UnionSchema` of `T`


Kriskras99:
Original issues #449 and #458 by @PookieBuns 

Currently the `SchemaAware(De)Serializer` implementations only support one representation of `Option<T>`: `["null", T::get_schema()]` (order of `null` and `T` does not matter). This means that the following schema:
```json
[
    "null",
    "int",
    "string"
]
```
can only map to the following enum:
```rust
enum Example {
    Null,
    Int(i32),
    String(String),
}
```

It is requested that the following `Option` would also match this schema:
```rust
enum Example {
    Int(i32),
    String(String),
}
Option<Example>
```

This would require the following changes to the serializer:
1. In `serialize_some` and `serialize_none`: Only require that the current schema is a `Schema::Union` with a `Schema::Null` variant
2. In `serialize_some` there are two cases to keep in mind:
  - If there are only two variants, serialize using the current implementation. 
  - If there are more variants, serialize `T` using the current union schema but provide a flag to the serializer that the `null` variant cannot be used. It would be preferable to modify the union schema/create a new union schema so that the `null` variant does not exist, but as we need a reference to the schema that would mean using a `Cow<Schema>` everywhere.
3. In `get_resolved_union_variant` the logic needs to be modified to correct the index:
  - If the index is before the `null` variant, it can be used as is
  - If the index is at or after the `null` variant, the index needs to be incremented by one
4. `UnionSerializer` would also need to be changed to ignore the null variant when the flag is set

This would require the following changes to the deserializer:
1. In `deserialize_option`:
  - Only require that the current schema is a `Schema::Union` with a `Schema::Null` variant
  - If a `null` is read, then the current logic can be used
  - Otherwise, store the index that was read and `visit_some` with the current deserializer
2. In `with_union` reading the index must be skipped if the index is already set.
3. In `deserialize_enum` the `UnionEnumDeserializer` must be provided with the stored index
4. In `UnionEnumDeserializer::variant_seed` the stored index must be used if available 

This would require the `AvroSchemaComponent` implementation for `Option<T>` to only panic if `T::get_schema` is a union with a `Schema::Null` variant, otherwise modify that schema to include `Schema::Null`.

There is probably more changes needed, but this is what I can think of right now.

PR #519, opened 2026-03-26
feat: Support flatten Option<T> to UnionSchema

PookieBuns:
Drafted up an initial PR to support #517

However, I would like y'alls input before I proceed further. Specifically, whether we should support Union branch lookup via enum variant name (new behavior, breaking change), or keep using variant index. I personally am more in favor of variant name, since it seems more intuitive to me wrt I don't have to worry about the order of my enum variants (and now you can actually use `#[serde(other)]` because that required it to be the last variant). I guess the cons of this is worse performance since there is O(n) lookup rather than direct indexing.

### DIFF
--- a/avro/src/documentation/serde_data_model_to_avro.rs
+++ b/avro/src/documentation/serde_data_model_to_avro.rs
@@ -71,12 +71,12 @@
 //!     - **unit_variant** => [`Schema::Record`] named as the variant and with no fields.
 //!     - **newtype_variant** => [`Schema::Record`] named as the variant and with one field.
 //!       The schema must have the attribute `org.apache.avro.rust.union_of_records` with the value set to `true`.
-//!     - **tuple_variant** => [`Schema::Record`] named as the variant and with as many fields as there are element.
+//!     - **tuple_variant** => [`Schema::Record`] named as the variant and with as many fields as there are elements.
 //!     - **struct_variant** => [`Schema::Record`] named as the variant and with a field for every field of the struct variant.
 //! - As a [`Schema::Union`] without the wrapper [`Schema::Record`], all schemas must be unique:
 //!     - **unit_variant** => [`Schema::Null`].
 //!     - **newtype_variant** => The schema of the inner type.
-//!     - **tuple_variant** => [`Schema::Record`] named as the variant and with as many fields as there are element.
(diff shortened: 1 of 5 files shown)


avro/src/serde/deser_schema/mod.rs
```diff
@@ -738,6 +750,89 @@ impl<'de, 's, 'r, R: Read, S: Borrow<Schema>> Deserializer<'de>
     }
 }
 
+struct ThreePlusVariantUnionDeserializer<'s, 'r, R: Read, S: Borrow<Schema>> {
```
PookieBuns: I also hate this name but I couldn't think of anything better. Am open to suggestions.

avro/src/serde/derive.rs
```diff
@@ -590,11 +593,16 @@ where
...
         enclosing_namespace: NamespaceRef,
     ) -> Schema {
-        let variants = vec![
-            Schema::Null,
-            T::get_schema_in_ctxt(named_schemas, enclosing_namespace),
-        ];
-
+        let variants = match T::get_schema_in_ctxt(named_schemas, enclosing_namespace) {
+            Schema::Union(union) if union.index_of_schema_kind(SchemaKind::Null).is_some() => {
+                union.schemas
```
Kriskras99: This must panic. Otherwise `Option<Enum { A, B(String) }>` is equivalent to `Enum { A, B(String) }`, so you would never be able to deserialize `Some(Enum::A)`.

PookieBuns: updated [here](https://github.com/apache/avro-rs/pull/519/changes/66b6d52ccc46dddf8d1eb2eb5a82f12566888db7)!

On a side note though (and probably out of scope), do we want to panic for library generated impls? Because we have no control over how people define their structs so if they define a struct with `Option<Option<T>>` getting the schema would cause a panic... Maybe returning result would be a potential option (pun not intended)? (Though its a breaking change)

Kriskras99:
Can you edit the target branch to `main`? I can't seem to change it

avro/src/serde/derive.rs
Kriskras99: I think it's okay to have the panic. Any user of the trait will use their type in tests, so the panic should also always trigger in tests.

PR #528, opened 2026-04-03
Support flattening Option<T> into the UnionSchema of T

### LINKED ISSUE #517: Support flattening `Option<T>` into the `UnionSchema` of `T`

PookieBuns:
closes #517. Original PR at #519


Kriskras99:
Hi @PookieBuns,

I've finally had time to review this in depth. I think this is a good thing to add and it's almost at the point where we can merge it. I've pushed a commit which adds documentation and simplifies the code a bit.
Unfortunately there are 3 tests (which you had marked as `should_panic`) which I think should work. The issue common to these 3 tests is an (union) enum with only one variant inside an Option. To the serializer code it looks like an Option so it takes the wrong branch at `serde/ser_schema/mod.rs:409`. As I think this is a common case (there is no serde trickery here with untagged or other attributes) it needs to be fixed before we can merge it (it would be confusing for the user if this case does not work).
I'll therefore put it onto the 0.23 milestone and start thinking about how we can fix this without degrading performance for Options and without requiring more complex code.

PR #530, opened 2026-04-05
feat: seekable reader

splix:
Added a feature to seek to a particular block when reading an Avro file. 

The Reader now provides the current Block position and for a Seek'able input it can seek to the specific position, assuming it's a valid start position of a block.

### DIFF
--- a/avro/src/reader/block.rs
+++ b/avro/src/reader/block.rs
@@ -17,7 +17,7 @@
 
 use std::{
     collections::HashMap,
-    io::{ErrorKind, Read},
+    io::{ErrorKind, Read, Seek, SeekFrom},
     str::FromStr,
 };
 
@@ -35,10 +35,57 @@ use crate::{
     util,
 };
 
+/// Position and size of a single Avro data block within the file stream.
+///
+/// Captured automatically as blocks are read during forward iteration.
+/// Use with [`super::Reader::seek_to_block`] to jump back to a previously-read block.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct BlockPosition {
+    /// Byte offset in the stream where this block starts (before the object-count varint).
+    pub offset: u64,
+    /// Total number of records in this block.
+    pub message_count: usize,
+}
+
+/// Wraps an inner reader and tracks the current byte position.
+///
+/// Avoids requiring `Seek` just to know how many bytes have been consumed.
+/// When the inner reader also implements `Seek`, seeking updates the tracked position.
+#[derive(Debug, Clone)]
+struct PositionTracker<R> {
+    inner: R,
+    pos: u64,
+}
+
+impl<R> PositionTracker<R> {
+    fn new(inner: R) -> Self {
+        Self { inner, pos: 0 }
+    }
+
+    fn position(&self) -> u64 {
+        self.pos
+    }
+}
+
+impl<R: Read> Read for PositionTracker<R> {
+    fn read(&mut self, buf: &mut [u8]) -> std::io::Result<usize> {
+        let n = self.inner.read(buf)?;
+        self.pos += n as u64;
+        Ok(n)
+    }
+}
+
+impl<R: Seek> Seek for PositionTracker<R> {
+    fn seek(&mut self, from: SeekFrom) -> std::io::Result<u64> {
+        self.pos = self.inner.seek(from)?;
+        Ok(self.pos)
+    }
+}
+
 /// Internal Block reader.
 #[derive(Debug, Clone)]
 pub(super) struct Block<'r, R> {
-    reader: R,
+    reader: PositionTracker<R>,
     /// Internal buffering to reduce allocation.
     buf: Vec<u8>,
     buf_idx: usize,
@@ -51,6 +98,10 @@ pub(super) struct Block<'r, R> {
     pub(super) user_metadata: HashMap<String, Vec<u8>>,
     names_refs: Names,
     human_readable: bool,
+    /// Byte offset where data blocks begin (right after header and sync marker).
+    pub(super) data_start: u64,
+    /// Position and record count of the currently loaded block.
+    pub(super) current_block_info: Option<BlockPosition>,
 }
 
 impl<'r, R: Read> Block<'r, R> {
@@ -60,7 +111,7 @@ impl<'r, R: Read> Block<'r, R> {
         human_readable: bool,
     ) -> AvroResult<Block<'r, R>> {
         let mut block = Block {
-            reader,
+            reader: PositionTracker::new(reader),
             codec: Codec::Null,
             writer_schema: Schema::Null,
             schemata,
@@ -71,9 +122,12 @@ impl<'r, R: Read> Block<'r, R> {
             user_metadata: Default::default(),
             names_refs: Default::default(),
             human_readable,
+            data_start: 0,
+            current_block_info: None,
         };
 
         block.read_header()?;
+        block.data_start = block.reader.position();
         Ok(block)
     }
 
@@ -142,6 +196,7 @@ impl<'r, R: Read> Block<'r, R> {
     /// the block. The objects are stored in an internal buffer to the `Reader`.
     fn read_block_next(&mut self) -> AvroResult<()> {
         assert!(self.is_empty(), "Expected self to be empty!");
+        let block_start = self.reader.position();
         match util::read_long(&mut self.reader).map_err(Error::into_details) {
             Ok(block_len) => {
                 self.message_count = block_len as usize;
@@ -155,6 +210,11 @@ impl<'r, R: Read> Block<'r, R> {
                 if marker != self.marker {
                     return Err(Details::GetBlockMarker.into());
                 }
+
+                self.current_block_info = Some(BlockPosition {
+                    offset: block_start,
+                    message_count: block_len as usize,
+                });
 
                 // NOTE (JAB): This doesn't fit this Reader pattern very well.
                 // `self.buf` is a growable buffer that is reused as the reader is iterated.
@@ -292,6 +352,23 @@ impl<'r, R: Read> Block<'r, R> {
                 warn!("User metadata values must be Value::Bytes, found {wrong:?}");
             }
         }
+    }
+}
+
+impl<R: Read + Seek> Block<'_, R> {
+    /// Seek the underlying stream to `offset` and read the block there.
+    /// Validates the sync marker to confirm it's a real block boundary.
+    pub(super) fn seek_to_block(&mut self, offset: u64) -> AvroResult<()> {
+        self.reader
+            .seek(SeekFrom::Start(offset))
+            .map_err(Details::SeekToBlock)?;
+
+        self.buf.clear();
+        self.buf_idx = 0;
+        self.message_count = 0;
+        self.current_block_info = None;
+
+        self.read_block_next()
     }
 }
 
--- a/avro/src/reader/block.rs
+++ b/avro/src/reader/block.rs
@@ -35,7 +35,7 @@ use crate::{
     util,
 };
 
-/// Position and size of a single Avro data block within the file stream.
+/// Byte offset and record count of a single Avro data block.
 ///
 /// Captured automatically as blocks are read during forward iteration.
 /// Use with [`super::Reader::seek_to_block`] to jump back to a previously-read block.
@@ -358,6 +358,8 @@ impl<'r, R: Read> Block<'r, R> {
 impl<R: Read + Seek> Block<'_, R> {
     /// Seek the underlying stream to `offset` and read the block there.
     /// Validates the sync marker to confirm it's a real block boundary.
+    /// Returns an error if no valid block can be read at the offset
+    /// (e.g., the offset is at or past EOF).
     pub(super) fn seek_to_block(&mut self, offset: u64) -> AvroResult<()> {
         self.reader
             .seek(SeekFrom::Start(offset))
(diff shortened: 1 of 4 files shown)


martin-g:
> Added a feature to seek to a particular block when reading an Avro file.

Do you have a use case for this functionality ?

avro/src/reader/block.rs
```diff
@@ -35,10 +35,57 @@ use crate::{
...
+        self.pos
+    }
+}
+
+impl<R: Read> Read for PositionTracker<R> {
+    fn read(&mut self, buf: &mut [u8]) -> std::io::Result<usize> {
+        let n = self.inner.read(buf)?;
+        self.pos += n as u64;
+        Ok(n)
+    }
```
martin-g, line 75: The implementation assumes that `read_buf()`, `read_exact()`, `read_vectored()`, ... use their default impls and delegate to `read()`. But if they are overwritten then the tracking breaks.

avro/src/reader/mod.rs
```diff
@@ -159,6 +164,41 @@ impl<R: Read> Iterator for Reader<'_, R> {
...
+    ///
+    /// The offset must point to the start of a valid data block (before its
+    /// object-count varint). The block is read, decompressed, and its sync
+    /// marker is validated against the file header. After this call, [`Iterator::next`]
+    /// yields the first record in that block.
+    ///
+    /// Typically the caller saves offsets from [`current_block`](Self::current_block)
+    /// during forward iteration and later passes them here to jump back.
+    pub fn seek_to_block(&mut self, offset: u64) -> AvroResult<()> {
+        self.block.seek_to_block(offset)?;
```
martin-g: If this fails then `self.errored` should be set to `true`, no ?

avro/src/reader/block.rs
```diff
@@ -295,6 +355,36 @@ impl<'r, R: Read> Block<'r, R> {
...
+        self.buf.clear();
+        self.buf_idx = 0;
+        self.message_count = 0;
+        self.current_block_info = None;
+
+        // read_block_next treats UnexpectedEof as a clean end-of-stream
+        // (returns Ok with message_count=0). That's correct for forward
+        // iteration but wrong here — the caller asked for a specific block.
+        self.read_block_next()?;
+        if self.is_empty() {
```
martin-g: The Avro spec does not disallow empty blocks.
IIRC it happens when a `types::Value::Null` is flushed.

avro/src/reader/block.rs
```diff
@@ -156,6 +211,11 @@ impl<'r, R: Read> Block<'r, R> {
                     return Err(Details::GetBlockMarker.into());
                 }
 
+                self.current_block_info = Some(BlockPosition {
+                    offset: block_start,
+                    message_count: block_len as usize,
```
martin-g, line 216:
```suggestion
                    message_count: self.message_count,
```

splix:
>>   Added a feature to seek to a particular block when reading an Avro file.

>  Do you have a use case for this functionality ?

I need to read just a few records from a large Avro file, and without this it's incredibly inefficient as I need to read the whole file from the start each time.

avro/src/reader/block.rs
splix: Is there any common examples for this behavior?

avro/src/reader/block.rs
splix: I actually wrote this part specifically this way to ensure it would not lost, if a refactoring or other change are applied. It cannot rely on the meaning of the self.message_count and its current value if those two will be separated into different places of code

avro/src/reader/mod.rs
splix: That makes sense. Please take a look at the update

avro/src/reader/block.rs
splix: That was a confusing part. Please take a look at the update

avro/src/reader/block.rs
Kriskras99: The tracking is done by wrapping the reader in a `PositionTracker` right? So it will always go through `PositionTracker::read`.

avro/src/reader/block.rs
Kriskras99: You can move `self.message_count = block_len as usize` down next to `self.current_block_info` to reduce that risk. As I do think Martin's suggestion is reasonable.

splix: I'm not arguing, any of this would work and I can change, that's no problem. But I'm curios to understand why you think this would be better?

My idea is just using the source of truth variable, so there is no way it gets a wrong number. Using the self.message_count works with the current code, but it doesn't give any guarantee if it changes. So I'm wondering why the second approach is better?

splix:
I'm wondering if you guys expect me to make some changes to the PR or we're waiting for something else?

Kriskras99:
I don't have further comments, @martin-g how about you?

avro/src/reader/block.rs
martin-g: The problem I see is: 
- imagine some custom Read implementation that overrides any of the provided methods (all Read methods but `read()`) for any reason. For example let's say the custom Read impl has a custom version of `read_exact()`
- PositionTracker is not opt-in and thus any usage of Read::read_exact() in the Avro codebase will always use PositionTracker::read_exact(). This will use the default impl (https://doc.rust-lang.org/stable/src/std/io/mod.rs.html#1044-1046) that will delegate to PositionTracker::read(). It will delegate to inner::read() and increment the read bytes (`pos`).
- So, the custom Read impl read_exact() won't be used at all and the user application has nothing to do

PositionTracker either needs to be smarter or opt-in.

avro/src/reader/mod.rs
```diff
@@ -159,6 +164,41 @@ impl<R: Read> Iterator for Reader<'_, R> {
...
+    ///
+    /// The offset must point to the start of a valid data block (before its
+    /// object-count varint). The block is read, decompressed, and its sync
+    /// marker is validated against the file header. After this call, [`Iterator::next`]
+    /// yields the first record in that block.
+    ///
+    /// Typically the caller saves offsets from [`current_block`](Self::current_block)
+    /// during forward iteration and later passes them here to jump back.
+    pub fn seek_to_block(&mut self, offset: u64) -> AvroResult<()> {
+        let seek_status = self.block.seek_to_block(offset);
```
martin-g, line 201: This logic has the side effect of resetting the `errored` flag.
If `errored` was set to `true` earlier, e.g. a call to `Reader::next()` then a call to `Reader::seek_to_block(offset)` that successfully reads the block at that offset will reset the `errored` to `false`. Now the user application can again call Reader::next() (or any other method that checks `errored` before doing any work).

avro/src/reader/block.rs
```diff
@@ -295,6 +355,26 @@ impl<'r, R: Read> Block<'r, R> {
...
 
+impl<R: Read + Seek> Block<'_, R> {
+    /// Seek the underlying stream to `offset` and read the block there.
+    /// Validates the sync marker to confirm it's a real block boundary.
+    /// Returns an error if no valid block can be read at the offset
+    /// (e.g., the offset is at or past EOF).
+    pub(super) fn seek_to_block(&mut self, offset: u64) -> AvroResult<()> {
+        self.reader
+            .seek(SeekFrom::Start(offset))
+            .map_err(Details::SeekToBlock)?;
```
martin-g: It would be safer to reset the state first (lines 368-371).
Currently if SeekToBlock is returned then the state won't be reset and if Block::read_block_next() is used it will fail at https://github.com/apache/avro-rs/blob/a6c7563272484dcd71fc1584b7b574e6d401ccea/avro/src/reader/block.rs#L198

avro/src/reader/block.rs
```diff
@@ -35,10 +35,57 @@ use crate::{
...
 };
 
+/// Byte offset and record count of a single Avro data block.
+///
+/// Captured automatically as blocks are read during forward iteration.
+/// Use with [`super::Reader::seek_to_block`] to jump back to a previously-read block.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct BlockPosition {
+    /// Byte offset in the stream where this block starts (before the object-count varint).
+    pub offset: u64,
```
martin-g: Since BlockPosition is public I'd recommend to make the fields private, add constructor and getters. This way it will be easier to add more fields later if needed without API breaks.

avro/src/reader/mod.rs
```diff
@@ -366,4 +406,111 @@ mod tests {
...
+        r.put("a", 1i64);
+        r.put("b", "x");
+        writer.append_value(r)?;
+        writer.flush()?;
+        let data = writer.into_inner()?;
+
+        let mut reader = Reader::new(Cursor::new(&data))?;
+        let result = reader.seek_to_block(7);
+        assert!(result.is_err());
+
```
martin-g, line 547: Please extend this test with seek to EOF and beyond:

```suggestion
let eof = data.len() as u64;
assert!(reader.seek_to_block(eof).is_err());
assert!(reader.seek_to_block(eof + 1).is_err());
```

avro/src/reader/block.rs
martin-g: `self.current_block_info` will be set even if the decompression fails.
And `Reader::current_block()` will see it.

martin-g:
> > > Added a feature to seek to a particular block when reading an Avro file.
> 
> > Do you have a use case for this functionality ?
> 
> I need to read just a few records from a large Avro file, and without this it's incredibly inefficient as I need to read the whole file from the start each time.

AFAIU you need to read the Avro data once to collect the offsets to be able to seek later, right ?
So, this helps for the following reads ?

avro/src/reader/block.rs
splix: Ah, I see. I misunderstand the original comment.

I made it opt-in, by default id doesn't interfere with the reading now. Ideally, I don't want to wrap anything at all, and just implement it for Read + Seek when available but it requires the _specialization_ which is not in stable yet, so maybe later this part can be rewritten.

avro/src/reader/block.rs
splix: Done

avro/src/reader/block.rs
splix: Done

avro/src/reader/mod.rs
splix: Right. If we seek to a valid record we are not in an _errored_ state anymore.

avro/src/reader/mod.rs
splix: Both seek and read_block_next accept position after EOF. No errors, it's just no messages. 

I've added a test, but checking for `None`, instead of `Err`

splix:
> AFAIU you need to read the Avro data once to collect the offsets to be able to seek later, right ?
So, this helps for the following reads ?

Right. 

First time I needed that was when we had a separate index with records, so we were just opening the Avro file at a needed position. Very large files (audit and traces) and small index for positions. That was in Java, where seeking is supported.

Now I need to show an Avro file on screen, using Rust. User can scroll the file back and forth. One option is keeping the whole file in memory, which is not optimal considering the usual sizes, or just keep the list of positions in file. The file opens from the start, so I always have all the previous positions.

Kriskras99:
I've rebased this branch and fixed the conflicts. @martin-g will we put this in 0.22 or will it be pushed back to 0.23?

martin-g:
Let's leave it for 0.23

avro/src/reader/block.rs
```diff
@@ -162,7 +287,18 @@ impl<'r, R: Read> Block<'r, R> {
                 // and replace `buf` with the new one, instead of reusing the same buffer.
                 // We can address this by using some "limited read" type to decode directly
                 // into the buffer. But this is fine, for now.
-                self.codec.decompress(&mut self.buf)
+                let next = self.codec.decompress(&mut self.buf);
+
+                // Make sure the position points only to a valid block
+                self.current_block_info = match next {
```
martin-g, line 293: Should this be done earlier ?!
If there is a problem with the marker then any call to `Reader::current_block()` will return wrong data.

avro/src/reader/block.rs
```diff
@@ -295,6 +431,26 @@ impl<'r, R: Read> Block<'r, R> {
     }
 }
 
+impl<R: Read + Seek> Block<'_, R> {
+    /// Seek the underlying stream to `offset` and read the block there.
+    /// Validates the sync marker to confirm it's a real block boundary.
+    /// Returns an error if no valid block can be read at the offset
+    /// (e.g., the offset is at or past EOF).
```
martin-g, line 438: https://github.com/apache/avro-rs/pull/530/changes#diff-6d2c5562c5bd4d218417ad6c2eedd5e2a4f1cad8980fba3198a896aed8828adeR303-R306 returns `Ok` on EOF, not an `Err`.

avro/src/reader/block.rs
```diff
@@ -51,14 +159,27 @@ pub(super) struct Block<'r, R> {
...
     pub(super) fn new(
         reader: R,
         schemata: Vec<&'r Schema>,
         human_readable: bool,
+        track_positions: bool,
     ) -> AvroResult<Block<'r, R>> {
+        let reader = if track_positions {
+            PositionTracker::Tracking {
+                inner: reader,
+                pos: 0,
```
martin-g, line 178: If the underlying reader R is already partially read then `pos: 0` won't be correct.
I am not sure how feasible this is though.
We can add one more optional field to ReaderBuilder - `initial_position`, with a default value of `0`.

Issue #531, opened 2026-04-06
"type" in schema JSON objects should only be allowed to be a string.


KitFieldhouse:
Hey all!

This implementation allows us to parse a schema of the form:

```
{
    "type":  { 
                       ....some correct schema here
     }
}
```

for instance, in this implementation, we can parse

```
{
    "type":  { 
                       "type": "enum",
                       "name": "testEnum",
                       "symbols": ["A","B","C"]
                   }
}
```
however, in the Java implementation this fails with:

```
No type: {"type":{"type":"enum","name":"testEnum","symbols":["A","B","C"]}}
```

From the [spec](https://avro.apache.org/docs/++version++/specification/) :

> A Schema is represented in [JSON](https://www.json.org/) by one of:
>
>   - A JSON string, naming a defined type.
>    - A JSON object, of the form:
>
>{"type": "typeName", ...attributes...}
>
>where typeName is either a primitive or derived type name, as defined below. Attributes not defined in this document are permitted as metadata, but must not affect the format of serialized data.
>
>   - A JSON array, representing a union of embedded types.

which, in my reading, makes it clear that when we are inside a JSON object as a schema, "type" can only be a string that equals a type name. 

What seems to be vague in the specification is if this type name should only be allowed to be a built-in type name or if user defined type names are allowed. My belief is that the rational option is to only allow built-in type names. I opened a [Jira issue](https://issues.apache.org/jira/browse/AVRO-4243) about this ambiguity in the spec.

martin-g:
Please see https://github.com/apache/avro-rs/issues/476

martin-g:
Actually you already commented there - https://github.com/apache/avro-rs/issues/476#issuecomment-4009905613

KitFieldhouse:
Yep, thanks for reminding me of that issue!

Someone correct me if I am wrong, but I believe the difference here is the location of "type". In this case we are looking at "type" in the root of a JSON schema instead of a record field JSON object. In addition, whereas "type" in a record field has to be a valid JSON schema object, I am arguing that the "type" field in a JSON schema object has to be a string, and can not be a JSON schema object.

Issue #508, opened 2026-03-07
Allow overriding built-in type names


Kriskras99:
As mentioned by @KitFieldhouse in https://github.com/apache/avro-rs/issues/476#issuecomment-4009905613 this test should pass:

```rust
#[test]
fn avro_rs_508_allow_overriding_type_names() -> TestResult {
    let schemas = Schema::parse_list([
    r#"{
        "type": "record",
        "name": "fixed",
        "fields": [{
            "name": "foo",
            "type": "string"
        }]
    }"#,
    r#"{
        "type": "record",
        "name": "Spam",
        "fields": [{
            "name": "bar",
            "type": "fixed"
        }]
    }"#
    ])?;

    Ok(())
}
```

Issue #353, opened 2025-12-04
schema parsing/resolution improvements?


KitFieldhouse:
Hi all, I have been digging into how schema parsing/resolution is done by this implementation and comparing its features and functionality compared to what I am familiar with Java. I wanted to collect some thoughts and see what the community thinks and, if there is interest, start towards getting it implemented.

## Observed Behavior

First some bugs/unexpected behavior.

### Schema parse ordering
With the current implementation of the parser, when parsing a list of schemata, the order in which the schemata are parsed matters, however, there is no guarantee that the schemata are parsed in input order. 

For instance, we can only reference a named schema type if it has already been parsed by the parser (with an exception being if the named schema happens to be the root schema of one of the provided schemata). However since there is no guarantee that the provided schemata are actually parsed in input order, sometimes parsing will succeed and sometimes it will fail. This can be tested by running this several times in a row. 

```rust
    #[test]
    fn test_ordering() -> TestResult {
        let schema1 = r#"
           {
            "type": "record", 
            "name": "record_1",
            "fields": [
                {
                    "name": "a_field",
                    "type": {
                        "type": "record",
                        "name": "user",
                        "fields": [
                                {
                                    "name": "age",
                                    "type": "float"
                                }
                        ]
                    } 
                 }
              ]
            }
            "#;

        let schema2 =  r#"{
            "type": "record",
            "name": "a_user",
            "fields": [{"name": "the_user", "type": "user"}]
        }"#;


        Schema::parse_list(vec![&schema1, &schema2])?;
        Ok(())
    }

```

Further, even if we parse in input order, there is no forward non-root level resolution of named schemata. For instance, I do not think there is any order in which to supply this pair of schemata to `parse_list` that will succeed.

```rust
        let schema1 = r#"
            {
                "name": "holds_foo_record",
                "type": "record",
                "fields": [{"name": "holds_foo", "type": 
                    {
                        "name": "foo_record",
                        "type": "record",
                        "fields": [{"name": "foo", "type": "baz_record"}]
                    }
                }]
            }
            "#;    

        let schema2 = r#"
            {
                "name": "holds_baz_record",
                "type": "record",
                "fields": [{"name": "holds_baz", "type": 
                    {
                        "name": "baz_record",
                        "type": "record",
                        "fields": [{"name": "baz", "type": "foo_record"}]
                    }}]
            }
            "#;    
```
Maybe it is possible to parse and resolve this type of schemata by using `parse_with_names`? 

[...]

Issue #368, opened 2025-12-16
Missing support for custom attributes on primitive types


yshcz:
Avro allows attaching custom properties to any type, including primitives.

In avro-rs, primitive schemas like `Schema::Long`/`Schema::Int`/etc are modeled as bare enum variants and there is no field to store custom properties. `Schema::custom_attributes()` also returns None for primitives.

As a result, parsing an object-form primitive schema silently drops all extra properties, and serializing the parsed schema emits just "long", so the metadata cannot be inspected or round-tripped.

reproducer: 
```rust
use apache_avro::Schema;
use apache_avro_test_helper::TestResult;
use serde_json::json;

#[test]
fn test1() -> TestResult {
    let input = json!({
        "type": "long",
        "custom-prop": "value"
    });

    let schema = Schema::parse(&input)?;
    assert!(matches!(schema, Schema::Long));

    let serialized = serde_json::to_string(&schema)?;
    assert!(
        serialized.contains("custom-prop"),
        "Expected serialized schema to include custom property key, but it was dropped. Serialized: {serialized}. Parsed schema: {schema:?}"
    );
    Ok(())
}
```

output:
```
  Expected serialized schema to include custom property key, but it was dropped. Serialized: "long". Parsed schema: Long
```

Iceberg encodes types like timestamptz to Avro as a primitive long with additional properties (e.g. `{"type":"long","logicalType":"timestamp-micros","adjust-to-utc":true`} or similar depending on precision).

Since apache-avro cannot represent custom properties on primitive types, the `adjust-to-utc` flag is dropped during parsing and the original schema semantics cannot be preserved.

PR #536, opened 2026-04-20
Schema parsing and resolved schema rework

KitFieldhouse:
Hey you all, this PR reworks schema parsing and `ResolvedSchema`.

I apologize for the size of this PR. I tried making it smaller, but I kept finding the changes made to parsing/`ResovledSchema` would trickle into many other sections of the crate.

This PR addresses issues \#531, \#353, and \#508.

Here is the gist of what this PR does:

# The new stuff
## Introduces `SchemaWithSymbols`.

Instead of representing a schema as a single recursive `Schema` variant, `SchemaWithSymbols` represents a schema as a hash map of all defined named schemata and a hash set of all referenced named schemata from that schema. This, coupled with the root schema enum, is enough to represent the entire schema parse tree.

The upshot of this representation is that when we go to check if a set of schemata form a complete context  we don't need to recurse into each schema and can just compare the "symbol tables" directly.

In addition, `SchemaWithSymbols` is essentially just made up of `Arc`'s and `HashMaps` of `Arc`'s. This should make cloning fairly quick compared to `Schema`.

## Introduces `ResolvedContext`.

`ResovledContext` wraps a list of schema definitions that have been checked for compatibility (no double naming of named schemata, all named references have a definition). This is closest to what the original `ResolvedSchema` looked like.

## Completely reworks `ResolvedSchema`.

As can be seen by the diff, `ResolvedSchema` has been essentially replaced in its entirety. `ResolvedSchema` now wraps a schema and all of the definitions needed to uniquely resolve all named schemata. In addition, we can convert `ResolvedSchema` into a `ResolvedNode`.

## Introduces `ResolvedNode`.

The struct `ResolvedNode` allows other code to walk down a schema tree without having to resolve `Schema::Ref` variants themselves. Further, since a `ResolvedNode` can only be created from a `ResolvedSchema`, we can make the type guarantee that this resolution will always find a unique named schema definition. `ResolvedNode` also makes the promise that any default for a record field has been checked against the field's schema.

## Introduces a `Resolver` trait.

This trait is designed to be used to hook up this crate's schema resolution logic to an external schema registry. Struct's implementing this trait can be provided to `ResolvedContext` either directly or through `ResolvedSchema`. If a local definition for a schema can not be found, the `Resolver` is called asking for it to either provide that schema or to signal an error.

# API changes

This PR introduces several breaking API changes, both in the public API and to the private API. I will try to summarize the changes for both layers.

## Semantic changes -- parsing

- Parser now always generates a `SchemaWithSymbols`. To keep compatibility with existing `Schema` based methods, the original schema parsing methods like `Schema::parse_str` will use the parser to generate a `SchemaWithSymbols` which is then unraveled into a normal `Schema` enum. This adds some overhead, but this is already a "slow" path so I don't think this introduces performance concerns.

- `Schema::parse_list` no longer checks if every named schema has a unique definition somewhere in the list. This is due to the conceptual shift mentioned in issue \#353. In this PR, parser methods do nothing more than parse each schema string and check that each one is internally consistent. Checking that a list of schemata is consistent (no duplicate named definitions, all references resolved) is pushed to `ResolvedSchema`.

## API changes -- public facing

- `ResolvedSchema` completely reworked. Any code that relied on the old `ResolvedSchema` would have to be refactored (though I think the needed changes are rather minimal). The new `ResolvedSchema` wraps a single schema and its resolved context, and is created via `ResolvedSchema::builder()`. See the new builder API: `.additional()` to provide
extra schemata, then `.build_one()`, `.build_array()`, or `.build_list()` as the terminal method depending on how many schemas you need resolved. Variants with `_with_resolver` accept a custom `Resolver` implementation for schema registry lookups.
- Elimination of `Schema::parse_str_with_list`. This method parsed a schema string alongside a list of other schema strings and returned the parsed schema plus the resolved list. The same result can now be achieved by parsing each schema independently with `SchemaWithSymbols::parse_str` and then resolving them together via `ResolvedSchema::builder().additional(others)?.build_one(target)?`.
- Elimination of `Schema::denormalize`. Denormalization (inlining named type definitions into references) is now handled via `ResolvedSchema::unravel()` or `SchemaWithSymbols::unravel()`.
- Elimination of `Schema::independent_canonical_form`. This is now accessed through `CompleteSchema::independent_canonical_form()` which requires a `ResolvedSchema`, making the resolved-context dependency explicit.
- `resolve_names` and `resolve_names_with_schemata` (were `pub(crate)`) are removed. Their role is subsumed by `ResolvedContext`.
- `ResolvedOwnedSchema` is removed. `ResolvedSchema` now owns its data (via `Arc`), making the separate owned variant unnecessary.
- `Schema::name()` now returns `Option<&Arc<Name>>` instead of `Option<&Name>`.
- `Value::validate_schemata(schemata)` replaced by `Value::validate_against_resolved(resolved_schema)`.
- `find_schema_with_known_schemata` is removed. Finding a matching element in a union requires that you have a `ResolvedSchema`. The method that performs union element matching is now `structural_match_on_schema` defined on `ResolvedUnion`.

## API changes -- private facing

- `decode_internal` signature changed from `(schema: &Schema, names: &HashMap<Name, S>, enclosing_namespace: NamespaceRef, reader: &mut R)` to
`(node: ResolvedNode, reader: &mut R)`. The `ResolvedNode` carries the schema, its resolved definitions, and handles namespace tracking internally
- `encode_internal` similarly changed from `(value: &Value, schema: &Schema, names: &HashMap<Name, S>, enclosing_namespace: NamespaceRef, writer: &mut W)`
to `(value: &Value, node: ResolvedNode, writer: &mut W)`.
- New functions `decode_complete` and `encode_complete` are added as the preferred entry points for decoding/encoding with a `ResolvedSchema`.
- `Block` now takes a `ResolvedSchema` instead of `Names`.
- `validate_internal` and `resolve_internal` on `Value` changed from generic `<S: Borrow<Schema> + Debug>` with `(schema, names, namespace)` parameters to taking a `ResolvedNode` directly.
- The generic `S: Borrow<Schema>` pattern used throughout decode, encode, and types for referencing named schemas is entirely replaced by `ResolvedNode`'s transparent reference resolution.
- Elimination of `Schema::parse_with_names` (was `pub(crate)`). Internal callers now use `SchemaWithSymbols` and `ResolvedSchema` instead of passing `Names` maps through parsing.

### DIFF
--- a/avro/src/decode.rs
+++ b/avro/src/decode.rs
@@ -15,19 +15,18 @@
 // specific language governing permissions and limitations
 // under the License.
 
-use crate::schema::{InnerDecimalSchema, NamespaceRef, UuidSchema};
+use crate::schema::{InnerDecimalSchema, UuidSchema};
 use crate::{
     AvroResult, Error,
     bigdecimal::deserialize_big_decimal,
     decimal::Decimal,
     duration::Duration,
     error::Details,
-    schema::{DecimalSchema, EnumSchema, FixedSchema, Name, RecordSchema, ResolvedSchema, Schema},
+    schema::{DecimalSchema, EnumSchema, FixedSchema, ResolvedSchema, Schema, ResolvedNode, ResolvedRecord},
     types::Value,
     util::{safe_len, zag_i32, zag_i64},
 };
 use std::{
-    borrow::Borrow,
     collections::HashMap,
     io::{ErrorKind, Read},
 };
@@ -69,20 +68,26 @@ fn decode_seq_len<R: Read>(reader: &mut R) -> AvroResult<usize> {
 }
 
 /// Decode a `Value` from avro format given its `Schema`.
+/// This function will always convert the supplied schema into a ResolvedSchema which
+/// may cuase performance concerns. Consider using decode_complete.
 pub fn decode<R: Read>(schema: &Schema, reader: &mut R) -> AvroResult<Value> {
-    let rs = ResolvedSchema::try_from(schema)?;
-    decode_internal(schema, rs.get_names(), None, reader)
+    let resolved : ResolvedSchema = schema.clone().try_into()?;
+    decode_complete(&resolved, reader)
 }
 
-pub(crate) fn decode_internal<R: Read, S: Borrow<Schema>>(
-    schema: &Schema,
-    names: &HashMap<Name, S>,
-    enclosing_namespace: NamespaceRef,
-    reader: &mut R,
+
+/// Decode a `Value` from avro format given its `Schema`.
+pub fn decode_complete<R: Read>(resolved_schema: &ResolvedSchema, reader: &mut R) -> AvroResult<Value> {
+    decode_internal(ResolvedNode::new(&resolved_schema), reader)
+}
+
+pub(crate) fn decode_internal<R: Read>(
+    node: ResolvedNode,
+    reader: &mut R
 ) -> AvroResult<Value> {
-    match schema {
-        Schema::Null => Ok(Value::Null),
-        Schema::Boolean => {
+    match node {
+        ResolvedNode::Null => Ok(Value::Null),
+        ResolvedNode::Boolean => {
             let mut buf = [0u8; 1];
             match reader.read_exact(&mut buf[..]) {
                 Ok(_) => match buf[0] {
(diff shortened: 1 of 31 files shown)


martin-g:
There are many conflicts!
You need to rebase!

martin-g:
Also it would be good to address https://github.com/apache/avro-rs/issues/368 if such big changes are going to be made.

KitFieldhouse:
> There are many conflicts! You need to rebase!

Yeah sorry, my bad, I will get that done.

> Also it would be good to address #368 if such big changes are going to be made.

I'll take a look!

KitFieldhouse:
I'm going to close this PR for the time being, this needs more work on my part, especially in hooking the "new" API up with the "old" API and making sure this PR doesn't introduce performance regressions.

PR #542, opened 2026-05-04
Fix for `Value::String` to `Value::Fixed` resolution ignoring the `size` from the schema.

KitFieldhouse:
Hey all,

I noticed that the method `resolve_fixed`  with a `Value::String` will ignore the `size`  of the `Schema::Fixed` and will simply accept whatever sized string is provided, even if its actually too long for the provided fixed.

This fix modifies `resolve_fixed` so that it checks the size of the string and rejects if its too long for the `Schema::Fixed`. In addition, if too short of a string is supplied, it is zero padded to the required length.

### DIFF
--- a/avro/src/types.rs
+++ b/avro/src/types.rs
@@ -1118,7 +1118,7 @@ impl Value {
                     bytes.resize(size, 0);
                     Ok(Value::Fixed(size, bytes))
                 }
-            },
+            }
             Value::Bytes(s) => {
                 if s.len() == size {
                     Ok(Value::Fixed(size, s))
@@ -3654,7 +3654,8 @@ Field with name '"b"' is not a member of the map items"#,
 
     #[test]
     fn avro_rs_542_test_bigger_string_to_smaller_fixed_resolution() -> TestResult {
-        let schema = Schema::parse_str(r#"
+        let schema = Schema::parse_str(
+            r#"
         {
             "name": "test",
             "type": "record",
@@ -3666,11 +3667,14 @@ Field with name '"b"' is not a member of the map items"#,
                     "size": 2
                 }
             }]
-        }"#)?;
+        }"#,
+        )?;
 
         let long_string = "This string is too long and won't fit!!".to_string();
-        let value = Value::Record(vec![("test_field".to_string(),
-                Value::String(long_string.clone()))]);
+        let value = Value::Record(vec![(
+            "test_field".to_string(),
+            Value::String(long_string.clone()),
+        )]);
 
         assert_eq!(
             value.resolve(&schema)
@@ -3685,7 +3689,8 @@ Field with name '"b"' is not a member of the map items"#,
 
     #[test]
     fn avro_rs_542_test_smaller_string_to_bigger_fixed_resolution() -> TestResult {
-        let schema = Schema::parse_str(r#"
+        let schema = Schema::parse_str(
+            r#"
         {
             "name": "test",
             "type": "record",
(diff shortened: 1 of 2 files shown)


avro/src/types.rs
```diff
@@ -3494,4 +3502,92 @@ Field with name '"b"' is not a member of the map items"#,
             "JSON number 18446744073709551615 could not be converted into an Avro value as it's too large"
         );
     }
+
+    #[test]
+    fn avro_rs_test_biggger_string_to_smaller_fixed_resolution() -> TestResult{
```
```suggestion
    fn avro_rs_542_test_bigger_string_to_smaller_fixed_resolution() -> TestResult {
```

avro/src/types.rs
```diff
@@ -3494,4 +3502,92 @@ Field with name '"b"' is not a member of the map items"#,
...
+            .into_details()
+            .to_string(),
+            Details::CompareFixedSizes { size: 2, n: long_string.len() }.to_string()
+        );
+
+        Ok(())
+    }
+
+    #[test]
+    fn avro_rs_test_smaller_string_to_bigger_fixed_resolution() -> TestResult{
```
```suggestion
    fn avro_rs_542_test_smaller_string_to_bigger_fixed_resolution() -> TestResult {
```

avro/src/types.rs
```diff
@@ -3494,4 +3502,92 @@ Field with name '"b"' is not a member of the map items"#,
...
+
+        assert_eq!(value,
+             Value::Record(vec![("test_field".to_string() , Value::Fixed(6, vec![97, 98, 99, 0 ,0 , 0]))])
+            );
+
+        Ok(())
+    }
+
+    #[test]
+    fn avro_rs_test_smaller_string_to_bigger_fixed_resolution_idempotence() -> TestResult{
```
```suggestion
    fn avro_rs_542_test_smaller_string_to_bigger_fixed_resolution_idempotence() -> TestResult {
```

avro/src/types.rs
```diff
@@ -3494,4 +3502,92 @@ Field with name '"b"' is not a member of the map items"#,
...
+                    "size": 6
+                }
+            }]
+        }"#)?;
+
+        let mut value = Value::Record(vec![("test_field".to_string(),
+                Value::String("abc".to_string()))]);
+
+        value = value.resolve(&schema)?;
+        value = value.resolve(&schema)?;
```
martin-g, line 3750:
```suggestion
```

avro/src/types.rs
```diff
@@ -1029,7 +1029,15 @@ impl Value {
                     Err(Details::CompareFixedSizes { size, n }.into())
                 }
             }
-            Value::String(s) => Ok(Value::Fixed(s.len(), s.into_bytes())),
+            Value::String(s) => {
+                if s.len() > size {
+                    Err(Details::CompareFixedSizes { size, n: s.len() }.into())
+                } else {
+                    let mut bytes = s.into_bytes();
+                    bytes.resize(size, 0);
```
martin-g, line 1118: I am not sure the padding applied here is the correct thing to do.
There is no such padding for Bytes->Fixed below. An exact match of the `length` with the `size` is required.
I think the same should be done here too. Otherwise the reader of the fixed data will have do handle the padding somehow.

avro/src/types.rs
```diff
@@ -3494,4 +3502,111 @@ Field with name '"b"' is not a member of the map items"#,
...
+            }]
+        }"#,
+        )?;
+
+        let mut value = Value::Record(vec![(
+            "test_field".to_string(),
+            Value::String("abc".to_string()),
+        )]);
+
+        value = value.resolve(&schema)?;
```
martin-g, line 3750:
```suggestion
        value = value.resolve(&schema)?;
        value = value.resolve(&schema)?;
```
it seems this second resolve was intended.

avro/tests/io.rs
```diff
@@ -127,7 +127,7 @@ fn default_value_examples() -> &'static Vec<(&'static str, &'static str, Value)>
             (
                 r#"{"type": "fixed", "name": "F", "size": 2}"#,
                 r#""a""#,
-                Value::Fixed(1, vec![97]),
+                Value::Fixed(2, vec![97, 0]),
             ), // ASCII 'a' => one byte
```
martin-g, line 131: The comment is outdated here but IMO the padding should be removed. See above.

avro/src/types.rs
KitFieldhouse: Sounds good, I agree with that logic.

martin-g: @KitFieldhouse Shall we remove the padding ?

Kriskras99:
@KitFieldhouse if you remove the padding this can still be merged for 0.22

PR #543, opened 2026-05-12
Deserializer support for BigDecimal when using apache_avro::serde::bigdecimal

jdarais:
It looks like serialization/deserialization was made more strict, and in the process, Deserializer no longer supports deserializing BigDecimal as a string, (which is what BigDecimal serializes/deserializes to/from by default.)  That's ok, since you can just add `#[serde(with="apache_avro::serde::bigdecimal")]`, but then it appears that deserializing BigDecimal as bytes is missing.  This PR adds the missing support for deserializing BigDecimal as bytes.

### DIFF
--- a/avro/src/serde/de.rs
+++ b/avro/src/serde/de.rs
@@ -1802,7 +1802,7 @@ mod tests {
     }
 
     #[test]
-    fn test_deserialize_bigdecimal_as_bytes() -> TestResult {
+    fn avro_543_deserialize_bigdecimal_as_bytes() -> TestResult {
         #[derive(Serialize, Deserialize, PartialEq, Eq, Debug)]
         #[serde(transparent)]
         struct WrappedBigDecimal (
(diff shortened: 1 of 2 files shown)


avro/src/serde/de.rs
```diff
@@ -1799,6 +1801,25 @@ mod tests {
         Ok(())
     }
 
+    #[test]
+    fn test_deserialize_bigdecimal_as_bytes() -> TestResult {
```
martin-g:
```suggestion
    fn avro_543_deserialize_bigdecimal_as_bytes() -> TestResult {
```

avro/src/serde/de.rs
martin-g:
```suggestion
                "Expected a String|Bytes|Fixed|Uuid|BigDecimal|Decimal|Duration, but got {:?}",
```

martin-g:
```suggestion
                "Expected a String|Bytes|Fixed|Uuid|BigDecimal|Decimal|Duration, but got {:?}",
```

martin-g (approved):

martin-g (approved):

martin-g:
Thank you, @jdarais !

PR #548, opened 2026-06-01
fix(schema): return Err instead of panicking on invalid name/alias/type-ref (#547)

### LINKED ISSUE #547: Schema::parse_str panics on invalid alias / type-reference / nested name (DoS) — sibling of #496
### Summary

`Schema::parse_str` / `Schema::parse` panics (instead of returning `Err`) when a schema's **alias**, **type reference**, or **nested type name** is not a valid Avro identifier. Because `parse_str` is a `Result`-returning public API, callers cannot guard against this with `?`/`map_err` — the panic unwinds out of the parse call. Parsing attacker-controlled schema JSON is therefore a denial-of-service.

This is the same `Name::new(..).unwrap()` class that was fixed for `Name::parse` (the top-level `name` field) in #496, but three sibling sites in `schema/parser.rs` were not covered.

### Reproduction

```rust
use apache_avro::Schema;

fn main() {
    // panics: called `Result::unwrap()` on an `Err` value:
    //   Invalid schema name :. It must match the regex '...'
    let _ = Schema::parse_str(r#"{"type":"record","name":"R","aliases":[":"],"fields":[]}"#);
}
```

Observed (apache-avro from `main`, also every release 0.17.0–0.21.0):

```
thread 'main' panicked at avro/src/schema/parser.rs:812:57:
called `Result::unwrap()` on an `Err` value: Error { details: Invalid schema name :. ... }
```

### Affected sites (`avro/src/schema/parser.rs`, current `main`)

- `get_already_seen_schema` — `Name::new_with_enclosing_namespace(typ, ns).unwrap()` on a `type` string reference.
- `fix_aliases_namespace` — `Alias::new(alias).unwrap()` on each `aliases` entry.
- `get_schema_type_name` — `Name::new(name).unwrap()` on a nested `type` object name.

All three call name constructors that already return `AvroResult` (they run `validate_schema_name`), so the validation error exists — it is just `unwrap()`-ed into a panic.

### Proposed fix

Propagate / gracefully handle the validation error instead of unwrapping:
- `get_already_seen_schema`: an unparyable type name cannot match an already-seen schema → return `None` (`.ok()?`).
- `fix_aliases_namespace`: return `AvroResult<Aliases>` and propagate (`?` at the 3 call sites) — an invalid alias is a schema error.
- `get_schema_type_name`: fall back to the enclosing name (`unwrap_or(name)`), matching the existing `_ => name` arm.

PR with fix + regression test to follow.

youichi-uda:
## What

`Schema::parse_str` / `Schema::parse` panicked (instead of returning `Err`) when a schema's **alias**, **type reference**, or **nested type name** was not a valid Avro identifier. Three sites in `schema/parser.rs` called name constructors that already return `AvroResult` and then `unwrap()`-ed them:

- `get_already_seen_schema` — `Name::new_with_enclosing_namespace(typ, ns).unwrap()`
- `fix_aliases_namespace` — `Alias::new(alias).unwrap()`
- `get_schema_type_name` — `Name::new(name).unwrap()`

Because `parse_str` is a `Result`-returning public API, a caller's `?` / `map_err` cannot catch the unwind, so parsing untrusted schema JSON is a denial-of-service. This is the sibling of the `Name::parse` fix in #496 (which covered only the top-level `name` field).

Fixes #547.

## Reproduction (before)

```rust
let _ = apache_avro::Schema::parse_str(r#"{"type":"record","name":"R","aliases":[":"],"fields":[]}"#);
// thread 'main' panicked at avro/src/schema/parser.rs:812:57:
// called `Result::unwrap()` on an `Err` value: Invalid schema name :. ...
```

Reproduces on `main` and on every release 0.17.0–0.21.0.

## How

Propagate / gracefully handle the validation error instead of unwrapping:

- **`get_already_seen_schema`** → `.ok()?`: an unparseable type name cannot match an already-seen schema, so treat it as "not seen"; the real validation error is still surfaced by the normal parse path.
- **`fix_aliases_namespace`** → returns `AvroResult<Aliases>` and propagates (`?` added at the 3 call sites). An invalid alias is a schema error.
- **`get_schema_type_name`** → `unwrap_or(name)`: fall back to the enclosing name, matching the existing `_ => name` arm.

No public API change. `Schema::parse_str` already returns `AvroResult`; these inputs now return `Err` instead of panicking.

## Tests

- New `tests/avro-rs-547.rs`: invalid aliases / nested names / type references all return `Err` (would panic before), plus a positive case ensuring valid namespace-qualified aliases still parse.
- `cargo test -p apache-avro` (lib + integration) passes; `cargo fmt --check` and `cargo clippy` clean.

### DIFF
--- a/avro/src/schema/parser.rs
+++ b/avro/src/schema/parser.rs
@@ -96,7 +96,7 @@ impl Parser {
                 .expect("Key unexpectedly missing");
             let parsed = self.parse(&value, None)?;
             self.parsed_schemas
-                .insert(self.get_schema_type_name(name, &value), parsed);
+                .insert(self.get_schema_type_name(name, &value)?, parsed);
         }
         Ok(())
     }
@@ -197,7 +197,7 @@ impl Parser {
         // parsing a full schema from inside another schema. Other full schema will not inherit namespace
         let parsed = self.parse(&value, None)?;
         self.parsed_schemas.insert(
-            self.get_schema_type_name(fully_qualified_name, &value),
+            self.get_schema_type_name(fully_qualified_name, &value)?,
             parsed.clone(),
         );
 
@@ -805,33 +805,32 @@ impl Parser {
                 aliases
                     .iter()
                     .map(|alias| {
-                        let qualified = if alias.find('.').is_none() {
-                            match namespace {
-                                Some(ns) => format!("{ns}.{alias}"),
-                                None => alias.clone(),
-                            }
-                        } else {
-                            alias.clone()
-                        };
                         // An alias that is not a valid Avro name is a schema
                         // error — propagate it instead of unwrapping (which
                         // would panic on attacker-controlled schema JSON).
-                        Alias::new(qualified.as_str())
+                        if alias.contains('.') {
+                            Alias::new(alias.as_str())
+                        } else {
+                            match namespace {
+                                Some(ns) => Alias::new(&format!("{ns}.{alias}")),
+                                None => Alias::new(alias.as_str()),
+                            }
+                        }
                     })
                     .collect::<AvroResult<Vec<_>>>()
             })
             .transpose()
     }
 
-    fn get_schema_type_name(&self, name: Name, value: &Value) -> Name {
+    fn get_schema_type_name(&self, name: Name, value: &Value) -> AvroResult<Name> {
         match value.get("type") {
             Some(Value::Object(complex_type)) => match complex_type.name() {
-                // Fall back to the enclosing name if the nested `type` name is
+                // Propagate the validation error if the nested `type` name is
                 // not a valid Avro name, rather than panicking on `unwrap()`.
-                Some(type_name) => Name::new(type_name).unwrap_or(name),
-                _ => name,
+                Some(type_name) => Name::new(type_name),
+                _ => Ok(name),
             },
-            _ => name,
+            _ => Ok(name),
         }
     }
 
--- a/avro/src/schema/parser.rs
+++ b/avro/src/schema/parser.rs
@@ -522,8 +522,12 @@ impl Parser {
     ) -> Option<&Schema> {
         match complex.get("type") {
             Some(Value::String(typ)) => {
+                // A `type` that is not a valid Avro name cannot match any
+                // already-seen schema, so treat it as "not seen" rather than
+                // panicking. The real validation error is surfaced later by
+                // the parse path that calls `Name::parse`.
                 let name =
-                    Name::new_with_enclosing_namespace(typ.as_str(), enclosing_namespace).unwrap();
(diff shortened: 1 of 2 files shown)


avro/src/schema/parser.rs
```diff
@@ -795,29 +799,36 @@ impl Parser {
...
+                                Some(ns) => format!("{ns}.{alias}"),
+                                None => alias.clone(),
+                            }
+                        } else {
+                            alias.clone()
+                        };
+                        // An alias that is not a valid Avro name is a schema
+                        // error — propagate it instead of unwrapping (which
+                        // would panic on attacker-controlled schema JSON).
+                        Alias::new(qualified.as_str())
```
martin-g: Let's improve this a bit further:
1) use `contains()`
2) pass &str to avoid some clones
 
```suggestion
                        // An alias that is not a valid Avro name is a schema
                        // error — propagate it instead of unwrapping (which
                        // would panic on attacker-controlled schema JSON).
                        if alias.contains('.') {
                            Alias::new(alias.as_str())
                        } else {
                            match namespace {
                                Some(ns) => Alias::new(&format!("{ns}.{alias}")),
                                None => Alias::new(alias.as_str()),
                            }
                        }
```

avro/src/schema/parser.rs
```diff
@@ -795,29 +799,36 @@ impl Parser {
...
+            .transpose()
     }
 
     fn get_schema_type_name(&self, name: Name, value: &Value) -> Name {
         match value.get("type") {
             Some(Value::Object(complex_type)) => match complex_type.name() {
-                Some(name) => Name::new(name).unwrap(),
+                // Fall back to the enclosing name if the nested `type` name is
+                // not a valid Avro name, rather than panicking on `unwrap()`.
+                Some(type_name) => Name::new(type_name).unwrap_or(name),
```
martin-g: I am not sure about this one.
Maybe we should change the signature to `Result<Name>` ?!
Silently ignoring the nested type name feels wrong.

avro/src/schema/parser.rs
youichi-uda: Done in ac0282c — switched to `contains('.')` and pass `&str` / `&format!("{ns}.{alias}")` straight into `Alias::new`, dropping the intermediate `qualified` String and the per-alias clones. Thanks!

avro/src/schema/parser.rs
youichi-uda: Agreed — silently dropping the nested type name is wrong. Changed `get_schema_type_name` to return `AvroResult<Name>` and propagate the validation error; the two callers now use `?`. Done in ac0282c.

martin-g (approved):

martin-g:
Thank you, @youichi-uda !

PR #557, opened 2026-06-17
fix: Optimize `Alias::new()` and `Parser::fix_aliases_namespace`

Kriskras99:
I noticed a missed optimization opportunity.
Using `impl Into<String> + AsRef<str>` allows us to reuse the allocation if the input is a `String`.
And by using `.into_iter()` in `fix_aliases_namespace` we get an allocation we can reuse.


martin-g (approved):

martin-g:
Maybe the PR title should be `refactor:....` ?

Kriskras99:
I've renamed it to `perf: ...`, which seems the best fit!

PR #558, opened 2026-06-17
feat: Namespace for named types defined by this library

Kriskras99:
While working on the enum generation I realised it would be useful to put the Avro types we define into their own namespace. This allows users to use their own `Duration`/`u64`/... types and we are free to implement `AvroSchema` for more Rust types if we're inclined to do so without breaking user code (unless they use the `org.apache.avro.rust` namespace but then it's their own problem).


avro/src/serde/ser_schema/union.rs
martin-g:
```suggestion
                r#"Expected Schema::Fixed(name: "org.apache.avro.rust.i128", size: 16) in variants"#,
```
It seems there is no test case for this and it was not caught by CI.

martin-g:
```suggestion
                r#"Expected Schema::Fixed(name: "org.apache.avro.rust.u64", size: 8) in variants"#,
```

martin-g:
```suggestion
                r#"Expected Schema::Fixed(name: "org.apache.avro.rust.u128", size: 16) in variants"#,
```

martin-g (approved):

PR #561, opened 2026-06-20
refactor: Centralize generating the actual `impl` in the `Implementation` struct

Kriskras99:
This makes it clearer how everything fits together, and allows extending shared logic further (for example, adding generated tests). And because it is not a `TokenStream`, accidentally returning only a schema expression instead of the entire implementation results in a compilation error.

This also fixes an issue where you couldn't override the `field_default` implementation for a transparent struct via a field attribute.

This should be the last refactor PR before the actual enum PRs

### DIFF
--- a/avro_derive/src/attributes/mod.rs
+++ b/avro_derive/src/attributes/mod.rs
@@ -17,8 +17,8 @@
 
 use crate::case::RenameRule;
 use darling::{FromAttributes, FromMeta};
-use proc_macro2::{Span, TokenStream};
-use quote::quote;
+use proc_macro2::Span;
+use serde_json::Value;
 use syn::{AttrStyle, Attribute, Expr, Ident, Path, spanned::Spanned};
 
 mod avro;
@@ -31,7 +31,7 @@ pub struct NamedTypeOptions {
     pub aliases: Vec<String>,
     pub rename_all: RenameRule,
     pub transparent: bool,
-    pub default: Option<TokenStream>,
+    pub default: Option<Value>,
 }
 
 impl NamedTypeOptions {
@@ -88,6 +88,7 @@ impl NamedTypeOptions {
                 || avro.name.is_some()
                 || avro.namespace.is_some()
                 || avro.doc.is_some()
+                || avro.default.is_some()
                 || !avro.alias.is_empty()
                 || avro.rename_all != RenameRule::None
                 || serde.rename_all.serialize != RenameRule::None
@@ -113,16 +114,12 @@ impl NamedTypeOptions {
         let doc = avro.doc.or_else(|| extract_rustdoc(attributes));
 
         let default = if let Some(default_value) = avro.default {
-            let _: serde_json::Value =
-                serde_json::from_str(default_value.as_str()).map_err(|e| {
-                    vec![syn::Error::new(
-                        ident.span(),
-                        format!("Invalid Avro `default` JSON: \n{e}"),
-                    )]
-                })?;
-            Some(quote! {
-                ::std::option::Option::Some(::serde_json::from_str(#default_value).expect(format!("Invalid JSON: {:?}", #default_value).as_str()))
-            })
+            Some(serde_json::from_str(default_value.as_str()).map_err(|e| {
+                vec![syn::Error::new(
+                    ident.span(),
+                    format!("Invalid Avro `default` JSON: \n{e}"),
+                )]
+            })?)
         } else {
             None
         };
@@ -140,6 +137,7 @@ impl NamedTypeOptions {
 
 pub struct VariantOptions {
     pub rename: Option<String>,
+    pub skip: bool,
 }
 
 impl VariantOptions {
@@ -176,6 +174,9 @@ impl VariantOptions {
 
         Ok(Self {
             rename: serde.rename,
+            // For variants we don't care about defaults for skipping, as Serde will error if a skipped
+            // variant is serialized or deserialized.
+            skip: serde.skip || (serde.skip_serializing && serde.skip_deserializing),
         })
     }
 }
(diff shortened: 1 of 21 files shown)


martin-g:
@Kriskras99 This needs resolving of conflicts.

Kriskras99:
@martin-g Thanks for the headsup, is fixed now!

avro_derive/src/attributes/mod.rs
```diff
@@ -112,20 +112,19 @@ impl NamedTypeOptions {
...
+        let default = if let Some(default_value) = avro.default {
+            let _: serde_json::Value =
+                serde_json::from_str(default_value.as_str()).map_err(|e| {
+                    vec![syn::Error::new(
+                        ident.span(),
+                        format!("Invalid Avro `default` JSON: \n{e}"),
+                    )]
+                })?;
+            Some(quote! {
+                ::std::option::Option::Some(::serde_json::from_str(#default_value).expect(format!("Invalid JSON: {:?}", #default_value).as_str()))
```
```suggestion
                ::std::option::Option::Some(::serde_json::from_str(#default_value).expect(concat!("Invalid JSON: ", #default_value)))
```
Same, but at compile time.

avro_derive/src/enums/plain.rs
```diff
@@ -17,25 +17,27 @@
...
+    container_attrs: NamedTypeOptions,
+    data: DataEnum,
+) -> Result<Implementation, Vec<syn::Error>> {
     let doc = preserve_optional(container_attrs.doc.as_ref());
     let enum_aliases = aliases(&container_attrs.aliases);
-    if data_enum.variants.iter().all(|v| Fields::Unit == v.fields) {
-        let default_value = default_enum_variant(data_enum, ident_span)?;
+    if data.variants.iter().all(|v| Fields::Unit == v.fields) {
+        let default_value = default_enum_variant(&data, input_span)?;
         let default = preserve_optional(default_value);
```
martin-g: I wonder should `container_attrs.rename_all` be used also for the `default` value ?!

avro_derive/src/structs.rs
```diff
@@ -162,9 +187,18 @@ pub fn get_transparent_struct_schema_def(
             }
 
             if let Some((field, attrs)) = found {
-                Ok((
-                    field_to_schema_expr(&field, attrs.with.clone())?,
-                    field_to_record_fields_expr(&field, attrs.with)?,
+                let field_default_expr = if container_attrs.default.is_none() {
```
martin-g: Hm. If there is no container default then use the field's default. Shouldn't the field's default overwrite the container's one ?

avro_derive/src/structs.rs
```diff
@@ -1,3 +1,5 @@
+use crate::attributes::{FieldOptions, NamedTypeOptions};
```
martin-g: These should be below the ASLv2 header comment.

avro_derive/src/structs.rs
Kriskras99: I think you're right, and even stricter: I don't think `default` should be allowed as a container attribute on a transparent struct, so I've added it to the deny list.

avro_derive/src/enums/plain.rs
Kriskras99: Yes it should! This did cause a kind of snowball effect as to fix this:
- I needed the `serde_json::Value` instead of a `TokenStream`
- Needed to restructure the plain enum code
- Decided to also implement a `Value` to `TokenStream` function so we don't need to parse at runtime
- I also made the serializer less struct (but keeping a fast path!), allowing `skip` to work on enum variants

avro_derive/src/enums/plain.rs
```diff
@@ -17,49 +17,128 @@
...
+            ));
+            continue;
+        }
+
+        // When a variant has the `#[default]` attribute and we don't have a default yet, we assign
+        // it as the default. If an enum has multiple variants with the default attribute, the compiler
+        // throws an error anyway so we can ignore that situation. If the default attribute is on a
+        // skipped variant, we ignore it as Serde will throw an error if we try to deserialize that
+        // variant.
+        if default.is_none() && has_default_attr(&variant.attrs) {
```
martin-g, line 81:
```suggestion
        if has_default_attr(&variant.attrs) {
```
shouldn't the variant's `#[default]` be used even when there is a container default ?

avro_derive/src/enums/mod.rs
```diff
@@ -18,50 +18,30 @@
...
+    container_attrs: NamedTypeOptions,
+    data: DataEnum,
+) -> Result<Implementation, Vec<syn::Error>> {
+    if container_attrs.transparent {
+        return Err(vec![syn::Error::new(
+            input_span,
+            "AvroSchema: `#[serde(transparent)]` is only supported on structs",
+        )]);
+    }
+    if data.variants.iter().all(|v| Fields::Unit == v.fields) {
```
martin-g: This check contradicts with the check for `.skip` at enums/plain.rs line 61: https://github.com/apache/avro-rs/pull/561/changes#diff-96375cb0af5acdd83b14cd2b9a9100754ce6dc2c50a7684a04ae82a6db4e15a5R62-R64

avro_derive/tests/derive.rs
```diff
@@ -2570,3 +2570,18 @@ fn avro_rs_476_skip_serializing_fielddefault_trait_none() {
         },
     }
 }
+
+#[test]
+fn avro_rs_561_transparent_struct_with_default_override() {
+    #[derive(AvroSchema)]
+    #[serde(transparent)]
+    struct T {
+        #[avro(default = "0")]
```
martin-g: let's use non-zero value for the default, to make it more obvious that it does not initialize to the primitive's default value as some other languages do

avro_derive/src/utils.rs
```diff
@@ -65,3 +66,65 @@ pub fn field_aliases(op: &[impl AsRef<str>]) -> TokenStream {
...
+            let mut values = Vec::with_capacity(len);
+            for (key, value) in object {
+                keys.push(key);
+                values.push(json_value_expr(value));
+            }
+
+            quote! {
+                ::serde_json::Value::Object({
+                    let mut map = ::serde_json::Map::with_capacity(#len);
+                    #(map.insert(::std::string::String::from(#keys), #values));*;
```
martin-g:
```suggestion
                    #(map.insert(::std::string::String::from(#keys), #values);)*
```
to avoid empty statements (stray semicolons) when the object is empty

avro_derive/src/enums/plain.rs
Kriskras99: The `#[default]` attribute is from `#[derive(Default)]` so I think the user should be able to overwrite that with the `#[avro(default = "...")]` attribute.

martin-g (approved):

PR #564, opened 2026-06-22
fix: Enable `#[serde(remote = "..")]` support

Kriskras99:
Remote just works, doesn't need any changes.

I have added a test to show how it works.


Kriskras99:
One thing that can improve this functionality is getting the name from the `remote` attribute (in the test the name of the schema will now be `DurationDef` instead of `Duration`)

avro_derive/src/attributes/serde.rs
```diff
@@ -117,8 +117,10 @@ pub struct ContainerAttributes {
     pub _default: Option<SerdeDefault>,
     /// This type is the serde implementation for a "remote" type.
     ///
-    /// This makes the (de)serialisation use/return a different type.
-    pub remote: Option<String>,
+    /// This makes the (de)serialisation use/return a different type. However, the this type should
+    /// match the other type so the schema we generate is correct.
```
martin-g: It is not clear what `other type` is.

I am saying it as someone who never never needed to use Serde `remote` before and I have not consulted with the Serde docs before making this comment. So, maybe it is clear enough for people trying to use this attribute.

avro_derive/tests/serde.rs
```diff
@@ -370,6 +370,74 @@ mod container_attributes {
             a: vec!["spam".to_string()],
         });
     }
+
+    #[test]
+    fn avro_rs_562_remote() {
```
martin-g:
```suggestion
    fn avro_rs_564_remote() {
```

martin-g (approved):

avro_derive/src/attributes/serde.rs
Kriskras99: I've tried to improve the documentation, but the best reference is the Serde docs itself. The most important part is that it doesn't affect the schema we generate.

PR #569, opened 2026-06-27
feat: Full enum support

Kriskras99:


Kriskras99:
Tests have now been added, so should be ready for review.

There will almost certainly be edgecases that neither of us think of, but we can fix that in a later release.

avro_derive/src/attributes/mod.rs
```diff
@@ -24,14 +24,117 @@ use syn::{AttrStyle, Attribute, Expr, Ident, Path, spanned::Spanned};
...
+    fn from_avro_and_serde(
+        avro: Option<avro::Repr>,
+        tag: Option<String>,
+        content: Option<String>,
+        untagged: bool,
+        span: Span,
+    ) -> Result<Option<Self>, syn::Error> {
+        match avro {
+            Some(avro::Repr::Enum) => {
+                if tag.is_some() || content.is_some() || untagged {
```
martin-g, line 62: nit: IMO it would be friendlier for the user if each condition has its own error. This way (s)he will know exactly what the problem is instead of checking which one from the listed incompatibilities is the issue.
If you agree then please also consider returning a `Vec<syn::Error>`

avro_derive/src/enums/bare_union.rs
```diff
@@ -0,0 +1,107 @@
...
+    data: DataEnum,
+) -> Result<Implementation, Vec<syn::Error>> {
+    let Some(Repr::BareUnion { untagged }) = container_attrs.repr else {
+        unreachable!()
+    };
+
+    let mut errors = Vec::new();
+    let mut variant_exprs = Vec::new();
+
+    // Used for checking that the resulting schema is usefull
```
martin-g:
```suggestion
    // Used for checking that the resulting schema is useful
```

avro_derive/src/enums/record_internally_tagged.rs
```diff
@@ -0,0 +1,215 @@
...
+            Ok(attrs) => attrs,
+            Err(errs) => {
+                errors.extend(errs);
+                continue;
+            }
+        };
+
+        if variant_attrs.skip {
+            continue;
+        } else if !variant_attrs.only_skip_rename_and_alias_can_be_set() {
```
martin-g: The error message at line 55 says `On unit variants...` but there is no check here that the variant type is a `unit` here.

```suggestion
        } else if matches!(&variant.fields, Fields::Unit)
            && !variant_attrs.only_skip_rename_and_alias_can_be_set() {
```

avro_derive/tests/derive.rs
```diff
@@ -1791,7 +1791,7 @@ fn avro_rs_247_serde_flatten_support_duplicate_field_name() {
         a: i32,
     }
 
-    Foo::get_schema();
+    panic!("{:?}", Foo::get_schema());
```
martin-g: debug leftover ?

avro_derive/src/enums/bare_union.rs
```diff
@@ -0,0 +1,107 @@
...
+    let mut have_null = false;
+    let mut names = HashSet::new();
+    let mut tuple_sizes = HashSet::new();
```
martin-g:
```suggestion
    let mut tuple_sizes = BTreeSet::new();
```
To make the order deterministic. Otherwise the UI tests may fail due to different order.

avro_derive/src/enums/bare_union.rs
```diff
@@ -0,0 +1,107 @@
...
+            container_attrs.rename_all,
+            container_attrs.rename_all_fields,
+            true,
+            true,
+            |info| {
+                match (info.schema_type, info.variant) {
+                    (_, FieldVariants::Named(0)) if untagged => {
+                        Err("AvroSchema: Empty struct variants are not allowed for `#[serde(untagged)]`".to_string())
+                    }
+                    (_, FieldVariants::Unnamed(len)) if untagged && len >= 2 && !tuple_sizes.insert(len) => {
```
martin-g, line 72: Why `len >= 2` ?
If you have an enum with two variants: `A(u32)` and `B(u32)` then both of them will serialize to Avro Int schemas.

avro_derive/src/enums/record_tag_content.rs
```diff
@@ -0,0 +1,161 @@
...
+) -> Result<Implementation, Vec<syn::Error>> {
+    let Some(Repr::RecordTagContent { tag, content }) = container_attrs.repr else {
+        unreachable!()
+    };
+    let mut errors = Vec::new();
+    let mut symbols = Vec::new();
+    let mut variant_exprs = Vec::new();
+
+    for variant in data.variants {
+        let variant_attrs = VariantOptions::new(&variant.attrs, variant.span())?;
```
martin-g: The `?` breaks the error accumulation. Is this intentional ?

avro_derive/src/enums/record_tag_content.rs
```diff
@@ -0,0 +1,161 @@
...
+            container_attrs.rename_all_fields,
+            true,
+            true,
+            |_| Ok(()),
+        ) {
+            Ok(expr) => variant_exprs.push(expr),
+            Err(errs) => errors.extend(errs),
+        }
+
+        symbols.push(name);
```
martin-g: Should this be moved to the `Ok(expr)` arm above (line 66) ?

avro_derive/src/enums/mod.rs
```diff
@@ -37,23 +45,223 @@ pub fn to_implementation(
...
+    if !options.alias.is_empty() {
+        errors.push(syn::Error::new(
+            span,
+            r#"AvroSchema: `#[avro(alias = "..")]` only works on newtype variants when the enum uses `#[avro(repr = "union_of_records")`"#
+        ));
     }
+    if options.rename.is_some() {
+        errors.push(syn::Error::new(
+            span,
+            r#"AvroSchema: `#[avro(alias = "..")]` only works on newtype variants when the enum uses `#[avro(repr = "union_of_records")`"#
```
martin-g:
```suggestion
            r#"AvroSchema: `#[avro(rename = "..")]` only works on newtype variants when the enum uses `#[avro(repr = "union_of_records")`"#
```

avro_derive/src/lib.rs
```diff
@@ -84,3 +84,37 @@ fn derive_avro_schema(input: DeriveInput) -> Result<Implementation, Vec<syn::Err
         )]),
     }
 }
+
+// #[cfg(test)]
```
martin-g: remove ?!

avro_derive/src/enums/bare_union.rs
```diff
@@ -0,0 +1,107 @@
...
+                        Err(format!("AvroSchema: Duplicate tuple sizes detected which is incompatible with `#[serde(untagged)]`: new: {len}, already seen: {tuple_sizes:?}"))
+                    }
+                    (SchemaType::Null, _) if have_null => Err(r#"AvroSchema: Two variants resolve to Schema::Null, this is not supported for `#[avro(repr = "bare_union")]"#.to_string()),
```
martin-g:
```suggestion
                    (SchemaType::Null, _) if have_null => Err(r#"AvroSchema: Two variants resolve to Schema::Null, this is not supported for `#[avro(repr = "bare_union")]`"#.to_string()),
```

avro_derive/tests/enum.rs
```diff
@@ -0,0 +1,721 @@
...
+        A(),
+    }
+
+    serde_assert(C::A());
+}
+
+#[test]
+fn avro_rs_561_bare_union_empty_tuple_variant() {
+    #[derive(Debug, Eq, PartialEq, AvroSchema, Serialize, Deserialize)]
+    #[avro(repr = "bare_union")]
```
martin-g, line 87: The documentation for BareUnion says that `#[serde(untagged)]` is required. It seems there is no check for this requirement. Or the documentation is out of date.

avro_derive/tests/enum.rs
```diff
@@ -0,0 +1,721 @@
...
+                    }
+                ]
+            }
+        ]
+    }"#,
+    )
+    .unwrap();
+
+    assert_eq!(Foo::get_schema(), schema);
+    serde_assert(Foo::A);
```
martin-g, line 246:
```suggestion
    serde_assert(Foo::A);
    serde_assert(Foo::Alt());
```

avro_derive/src/enums/mod.rs
```diff
@@ -37,23 +45,223 @@ pub fn to_implementation(
...
+                }
+                Fields::Unnamed(mut fields) if transparent_newtype && fields.unnamed.len() == 1 => {
+                    check_fn(FieldInfo {
+                        variant: FieldVariants::Unnamed(1),
+                        schema_type: SchemaType::Transparent,
+                    })
+                    .map_err(|m| vec![syn::Error::new(variant_span, m)])?;
+                    let pair = fields.unnamed.pop().expect("There is one field");
+                    let field = pair.into_value();
+                    let field_attributes = FieldOptions::new(&field.attrs, field.span())?;
```
martin-g:
```suggestion
                    let field_attributes = FieldOptions::new(&field.attrs, field.span())
+                        .and_then(|o| newtype_extra_attribute_checks(o, field.span()))?;
```

avro_derive/src/enums/plain.rs
martin-g:
```suggestion
            let renamed = rename_ident(
                &variant.ident,
                variant_attrs.rename.clone(),
                container_attrs.rename_all,
                RenameRule::apply_to_variant,
            );
```

avro_derive/tests/expanded/avro_rs_569_tuple_struct.expanded.rs
```diff
@@ -0,0 +1,101 @@
...
+                                        schema: <String as ::apache_avro::AvroSchemaComponent>::get_schema_in_ctxt(
+                                            named_schemas,
+                                            enclosing_namespace,
+                                        ),
+                                        custom_attributes: ::std::collections::BTreeMap::new(),
+                                    },
+                                ],
+                            ),
+                        ),
+                    )
```
martin-g, line 55:
```suggestion
                    )
                    .attributes(
                        [(
                            "org.apache.avro.rust.tuple".to_string(),
                            ::serde_json::value::Value::Bool(true),
                        )]
                        .into(),
                    )
```

avro_derive/tests/expanded/mod.rs
```diff
@@ -27,3 +27,9 @@ mod avro_rs_501_basic;
 mod avro_rs_501_namespace;
 mod avro_rs_501_reference;
 mod avro_rs_501_struct_with_optional;
+mod avro_rs_569_non_basic_enum;
+mod avro_rs_569_rename_all_fields;
+mod avro_rs_569_tag_content_enum;
+mod avro_rs_569_tag_enum;
+mod avro_rs_569_tag_struct;
```
martin-g, line 34:
```suggestion
mod avro_rs_569_tag_struct;
mod avro_rs_569_tuple_struct;
```

avro_derive/tests/ui/avro_rs_561_transparent_enum.rs
```diff
@@ -0,0 +1,26 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+use apache_avro::AvroSchema;
+
+#[derive(AvroSchema)]
+#[serde(transparent)]
+enum D {
+    A(A)
```
martin-g: What is the inner `A` here ?

```suggestion
    A(i32)
```
?

avro_derive/tests/ui/avro_rs_569_untagged_enum_multiple_null.stderr
```diff
@@ -0,0 +1,5 @@
+error: AvroSchema: Two variants resolve to Schema::Null, this is not supported for `#[avro(repr = "bare_union")]
```
martin-g:
```suggestion
error: AvroSchema: Two variants resolve to Schema::Null, this is not supported for `#[avro(repr = "bare_union")]`
```

avro_derive/src/enums/bare_union.rs
Kriskras99: An untagged bare union will do the following to tuple variants of different sizes:
- `()`: `Schema::Null`
- `(T1)`: `T1::get_schema_in_ctxt`
- `(T1,..,Tn`): `Schema::Record`

So only for tuple variants of size 2 or higher we need to check if their sizes are unique because the serializer will look for the first record schema (with the tuple attribute) that has the same length.
For tuple variants of size 1, it would be nice to check that the schemas are unique at compile time but we don't have that information, and it is caught at schema construction by the Union builder.
For size 0 we have the `(SchemaType::Null, _) if have_null` check.

I've added the following comment to the check:
```
// Tuples of size 2 or larger are represented as a Schema::Record with an "org.apache.avro.rust.tuple" attribute.
// As the only information Serde provides for untagged enums is the size of the tuple, the amount of fields in
// records with the tuple tag must be unique for it to find the correct variant.
// This check is not watertight as it will fail for `enum Abc { A((i32, i32)), B(i64, i64) }`
// as we don't know the schema for the newtype inside Abc::A.
```

avro_derive/src/enums/record_internally_tagged.rs
Kriskras99: I moved it into the match statement below

avro_derive/src/enums/record_tag_content.rs
Kriskras99: I don't think it matters, any errors thrown inside `variant_to_schema_expr` are related to creating the schema itself. If the symbol is a duplicate then we still want to inform the user even if one of the schema expressions failed to create.
I've moved the statement up directly after we create the `name` so they are together

avro_derive/tests/enum.rs
Kriskras99: Documentation is outdated, I realized `bare_union` doesn't require `untagged` but `untagged` does require `bare_union`

avro_derive/tests/ui/avro_rs_561_transparent_enum.rs
Kriskras99: Good question, probably leftover from some test I was doing. I've replaced it with a `String`.

avro_derive/tests/enum.rs
Kriskras99: This actually surfaced something we can't support. I've now disallowed empty tuple variants on enums with `tag` and `content` set. `Foo::Alt{}` does work, so I've added that as a test.

avro_derive/tests/expanded/avro_rs_569_tuple_struct.expanded.rs
Kriskras99: Tuple structs don't actually require the attribute, only real tuples and tuple variants

avro_derive/src/attributes/mod.rs
Kriskras99: If the list were longer I'd agree, but with only 3 attributes (of which untagged conflicts with tag/content) I don't think it's an issue here.

avro/src/schema/record/schema.rs
```diff
@@ -82,11 +82,13 @@ impl<S: record_schema_builder::State> RecordSchemaBuilder<S> {
...
 /// Calculate the lookup table for the given fields.
 fn calculate_lookup_table(fields: &[RecordField]) -> BTreeMap<String, usize> {
-    fields
+    let map: BTreeMap<_, _> = fields
         .iter()
         .enumerate()
         .map(|(i, field)| (field.name.clone(), i))
-        .collect()
+        .collect();
+    assert_eq!(map.len(), fields.len(), "Duplicate field names found");
```
martin-g, line 90:
```suggestion
    debug_assert_eq!(map.len(), fields.len(), "Duplicate field names found");
```

avro_derive/src/enums/mod.rs
```diff
@@ -37,23 +45,224 @@ pub fn to_implementation(
...
+
+fn newtype_extra_attribute_checks(
+    options: FieldOptions,
+    span: Span,
+) -> Result<FieldOptions, Vec<syn::Error>> {
+    let mut errors = Vec::new();
+    if options.doc.is_some() {
+        errors.push(syn::Error::new(
+            span,
+            r#"AvroSchema: `#[avro(doc = "..")]` only works on newtype variants when the enum uses `#[avro(repr = "union_of_records")`"#
```
martin-g:
```suggestion
            r#"AvroSchema: `#[avro(doc = "..")]` only works on newtype variants when the enum uses `#[avro(repr = "union_of_records")]`"#
```

avro_derive/src/enums/mod.rs
```diff
@@ -37,23 +45,224 @@ pub fn to_implementation(
...
+        ));
+    }
+    if !matches!(options.default, FieldDefault::Trait) {
+        errors.push(syn::Error::new(
+            span,
+            r#"AvroSchema: `#[avro(default = ..)]` only works on newtype variants when the enum uses `#[avro(repr = "union_of_records")`"#
```
martin-g:
```suggestion
            r#"AvroSchema: `#[avro(default = ..)]` only works on newtype variants when the enum uses `#[avro(repr = "union_of_records")]`"#
```

avro_derive/src/enums/mod.rs
```diff
@@ -37,23 +45,224 @@ pub fn to_implementation(
...
+        ));
+    }
+    if !options.alias.is_empty() {
+        errors.push(syn::Error::new(
+            span,
+            r#"AvroSchema: `#[avro(alias = "..")]` only works on newtype variants when the enum uses `#[avro(repr = "union_of_records")`"#
```
martin-g:
```suggestion
            r#"AvroSchema: `#[avro(alias = "..")]` only works on newtype variants when the enum uses `#[avro(repr = "union_of_records")]`"#
```

avro_derive/src/enums/mod.rs
```diff
@@ -37,23 +45,224 @@ pub fn to_implementation(
...
+        ));
     }
+    if options.rename.is_some() {
+        errors.push(syn::Error::new(
+            span,
+            r#"AvroSchema: `#[avro(rename = "..")]` only works on newtype variants when the enum uses `#[avro(repr = "union_of_records")`"#
```
martin-g:
```suggestion
            r#"AvroSchema: `#[avro(rename = "..")]` only works on newtype variants when the enum uses `#[avro(repr = "union_of_records")]`"#
```

avro_derive/src/enums/mod.rs
```diff
@@ -37,23 +45,224 @@ pub fn to_implementation(
...
+        ));
+    }
+    if options.flatten {
+        errors.push(syn::Error::new(
+            span,
+            r#"AvroSchema: `#[avro(flatten)]` only works on newtype variants when the enum uses `#[avro(repr = "union_of_records")`"#,
```
martin-g:
```suggestion
            r#"AvroSchema: `#[avro(flatten)]` only works on newtype variants when the enum uses `#[avro(repr = "union_of_records")]`"#,
```

avro_derive/src/attributes/mod.rs
```diff
@@ -24,14 +24,117 @@ use syn::{AttrStyle, Attribute, Expr, Ident, Path, spanned::Spanned};
...
+                    ))
+                }
+            }
+            None if tag.is_some() && content.is_some() => Ok(Some(Self::RecordTagContent {
+                tag: tag.unwrap(),
+                content: content.unwrap(),
+            })),
+            None if tag.is_some() => Ok(Some(Self::RecordInternallyTagged { tag: tag.unwrap() })),
+            None if untagged => Ok(Some(Self::BareUnion { untagged })),
+            None => Ok(None),
```
```suggestion
            None => match (tag, content, untagged) {
                (Some(tag), Some(content), false) => {
                    Ok(Some(Self::RecordTagContent { tag, content }))
                }
                (Some(tag), None, false) => {
                    Ok(Some(Self::RecordInternallyTagged { tag }))
                }
                (None, None, true) => Ok(Some(Self::BareUnion { untagged: true })),
                (None, None, false) => Ok(None),
                _ => Err(syn::Error::new(
                    span,
                    "AvroSchema: incompatible Serde tagging attributes",
                )),
            }
```
To prevent usage of incompatible Serde attributes.

avro_derive/tests/enum.rs
```diff
@@ -0,0 +1,764 @@
...
+            },
+            {
+                "name": "value",
+                "type": [
+                    "null",
+                    {
+                        "type": "record",
+                        "name": "Alt",
+                        "fields": [],
+                        "org.apache.avro.rust.tuple": true
```
martin-g: Why is `Alt {}` marked as a tuple here ?

avro/src/schema/record/schema.rs
Kriskras99: I'm not sure we should silence these in release mode, this will fail in the serializer otherwise (with `Details::FieldNameDuplicate`) but it will be far removed from the reason (an incorrect schema).

It would be nice if we could return an error, but `#[bon(skip)` doesn't allow that.

avro_derive/tests/enum.rs
Kriskras99: It's actually not in the generated schema, this is leftover from when it was `Alt()`. The test still passes because `Schema == Schema` does not check attributes. This is not the first time this has bitten us, we should think about changing the behaviour of `PartialEq` to check for absolute equality and have `Schema::specification_eq` and `Schema::fast_specification_eq` functions for the user to do specification related comparisons.

Kriskras99: By setting `schema_equality.rs:256` `false` to `true` I've caught 9 other tests where this happens. I've fixed them all, but cases like these will likely come back in the future when we add more tests.

Kriskras99: Maybe we could put a `#[cfg(test)]` on `schema_equality.rs:256` to set the boolean to `true` for tests.

martin-g (approved):

PR #571, opened 2026-06-30
feat: Derive tests for derived schemas

Kriskras99:
This adds a test to every generated schema to check if the schema can actually be created. Tests are automatically disabled for types with generics, consts, or lifetimes. They can also be disabled by setting `#[avro(test = false)]` on the type, for if a user uses a different validator at runtime or the type is inside a function which would trigger the `unnameable_test_items` lint.

In the future we could extend the attribute syntax to provide a different validator, so that use case would also be supported. I also think we should be able to test types with lifetimes, but that needs some further experimentation.

### DIFF
--- a/avro/src/serde/derive.rs
+++ b/avro/src/serde/derive.rs
@@ -105,7 +105,7 @@ use crate::{
 ///  - `#[avro(test = false)]`
 ///
 ///    To disable generating tests for derived schemas. This is useful if you use a different [validator](crate::validator)
-///    or the `unnameable_test_items` lint is triggered because the type your deriving is inside a function.
+///    or the `unnameable_test_items` lint is triggered because the type you are deriving is inside a function.
 ///
 ///
 /// #### Variant attributes
--- a/avro/src/serde/derive.rs
+++ b/avro/src/serde/derive.rs
@@ -102,6 +102,11 @@ use crate::{
 ///
 ///    Use the schema of the inner field directly. Is only allowed on structs with only one unskipped field.
 ///
+///  - `#[avro(test = false)]`
+///
+///    To disable generating tests for derived schemas. This is useful if you use a different [validator](crate::validator)
+///    or the `unnameable_test_items` lint is triggered because the type your deriving is inside a function.
+///
 ///
 /// #### Variant attributes
 ///
--- a/avro/src/serde/derive.rs
+++ b/avro/src/serde/derive.rs
@@ -102,7 +102,7 @@ use crate::{
 ///
 ///    Use the schema of the inner field directly. Is only allowed on structs with only one unskipped field.
 ///
-///  - `#[avro(test = false)]`
+///  - `#[avro(tests = false)]`
 ///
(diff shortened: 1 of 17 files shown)


avro/src/serde/derive.rs
```diff
@@ -102,6 +102,11 @@ use crate::{
 ///
 ///    Use the schema of the inner field directly. Is only allowed on structs with only one unskipped field.
 ///
+///  - `#[avro(test = false)]`
+///
+///    To disable generating tests for derived schemas. This is useful if you use a different [validator](crate::validator)
+///    or the `unnameable_test_items` lint is triggered because the type your deriving is inside a function.
```
```suggestion
///    or the `unnameable_test_items` lint is triggered because the type you're deriving is inside a function.
```

avro_derive/src/attributes/avro.rs
```diff
@@ -85,6 +85,9 @@ pub struct ContainerAttributes {
     /// Force the generator to use a certain schema representation
     #[darling(default)]
     pub repr: Option<Repr>,
+    /// Force the generator to use a certain schema representation
```
martin-g: The docstring is a copy/paste from line 85

avro_derive/src/implementation.rs
```diff
@@ -77,15 +81,46 @@ impl Implementation {
...
+    fn add_schema_test(&mut self) {
+        // Only add the schema test if there are no generics to deal with. I think it should be possible
+        // to be able to generate the test if there are only lifetimes, but that needs more experimentation.
+        if self.generics.lifetimes().count() == 0
+            && self.generics.type_params().count() == 0
+            && self.generics.const_params().count() == 0
+        {
+            let ident = &self.ident;
+            self.add_test(Test {
+                name: "schema".to_string(),
```
martin-g, line 117: I don't get the idea of this PR :-/
It is invasive by generating tests no one asked for.
The end user will have to opt-out instead of opt-in for those tests.

It is not clear what is being tested too.
1. The `name` is always `"schema"`. Is it really needed ?!
2. The body always prints `"Invalid schema"` to stderr
3. `should_panic` is always `None`
4. The body just calls `get_schema()`, so it tests runtime construction of the derived Schema and validates its `Name`. But one could assume that the users make use of the derived schema too in their application and tests, so this is kinda already covered.

avro/src/serde/derive.rs
```diff
@@ -102,6 +102,11 @@ use crate::{
 ///
 ///    Use the schema of the inner field directly. Is only allowed on structs with only one unskipped field.
 ///
+///  - `#[avro(test = false)]`
```
```suggestion
///  - `#[avro(tests = false)]`
```

avro_derive/src/implementation.rs
Kriskras99: I think especially point 4 is a good point, so let's close this for now.

(1 and 3 are to allow generating other kinds of tests, 2 is so that the user gets some feedback when the test fails).

Kriskras99:
Closing for now, needs more work

PR #577, opened 2026-07-06
fix: Pin `dtolnay/rust-toolchain` to the new allowed version

Kriskras99:


Kriskras99:
@martin-g I'm merging this without review as it fixes the broken main, and the cause of it. It also brings the Github Actions to the requirements of https://github.com/apache/infrastructure-actions, which should prevent this from happening again.

avro/Cargo.toml
```diff
@@ -57,7 +57,7 @@ crc32fast = { default-features = false, version = "1.5.0", optional = true }
 digest = { default-features = false, version = "0.11.3", features = ["block-api"] }
 miniz_oxide = { default-features = false, version = "0.9.1", features = ["with-alloc"] }
 log = { workspace = true }
-num-bigint = { default-features = false, version = "0.5.0", features = ["std", "serde"] }
+num-bigint = { default-features = false, version = "0.4.8", features = ["std", "serde"] }
```
martin-g, line 60: Why ?

avro_test_helper/Cargo.toml
```diff
@@ -31,7 +31,7 @@ readme = "README.md"
 
 [dependencies]
 better-panic = { default-features = false, version = "0.3.0" }
-ctor = { default-features = false, version = "0.11.1", features = ["dtor", "proc_macro"]  }
+linktime = { default-features = false, version = "0.18.0", features = ["ctor", "dtor", "proc_macro"]  }
```
martin-g, line 34: This change should have been a separate PR. It is not related to the ASF rules.

Maybe we should disable the `"link-section"` since we do not use it ?!

avro/Cargo.toml
Kriskras99: Upgrading `num-bigint` (in #572) is what broke main, the `bigdecimal` dependency is still on `0.4.8` so the types were incompatible. We should probably use the re-export provided by `bigdecimal` (if we don't use the type somewhere else) so we don't have to manually keep upgrading it together with `bigdecimal`.

avro_test_helper/Cargo.toml
Kriskras99: It should have been, sorry about that. I just noticed that it was very outdated and tried to upgrade it, the noticed that the `dtor` feature was gone and the crate had a new collection crate called `linktime`.

The `link-section` feature should be indeed disabled, I'll open up a PR to fix that.

Kriskras99: Actually the `link-section` feature is not enabled (`default-features = false`), it is in the `Cargo.lock` but that is because Cargo does put optional dependencies in the lockfile (see https://github.com/rust-lang/cargo/issues/10801).

avro/Cargo.toml
Kriskras99: We use `num-bigint` also for `Decimal`, so we shouldn't use the re-export from `bigdecimal`.

PR #579, opened 2026-07-06
feat: Allow to supply own names when using `Value::resolve` and `Value::Validate`

### LINKED ISSUE #574: Proposal: Reuse a pre-built `ResolvedSchema` in `Value` resolution
# Problem

`apache_avro::types::Value` exposes two public methods for schema resolution:

1. `Value::resolve(self, schema: &Schema)`
2. `Value::resolve_schemata(self, schema: &Schema, schemata: Vec<&Schema>)`

Both rebuild a `ResolvedSchema` from the target schema on **every call**. `resolve`
delegates to `resolve_schemata`, which in turn constructs the `ResolvedSchema`
before doing any work:

```rust
pub fn resolve(self, schema: &Schema) -> AvroResult<Self> {
    self.resolve_schemata(schema, Vec::with_capacity(0))
}

pub fn resolve_schemata(self, schema: &Schema, schemata: Vec<&Schema>) -> AvroResult<Self> {
    let enclosing_namespace = schema.namespace();
    let rs = if schemata.is_empty() {
        ResolvedSchema::try_from(schema)?      // rebuilt every call
    } else {
        ResolvedSchema::try_from(schemata)?    // rebuilt every call
    };
    self.resolve_internal(schema, rs.get_names(), enclosing_namespace, &None)
}
```

Building the `ResolvedSchema` walks the entire schema tree to collect every named
type (records, enums, fixed, and other named schemas) and populates a fresh
`NamesRef` (`HashMap<Name, &Schema>`). This cost is paid once per value.

When the reader schema is fixed and many values are resolved against it — for
example decoding a stream of messages that all share one schema — the schema tree
is re-walked and the name map is re-allocated for each value, even though the
result is identical every time. The per-value resolution itself is unavoidable
work; the repeated `ResolvedSchema` construction is not.

# Proposed API

Add a method that accepts a caller-provided name map, so the map can be built once
and reused. Taking the `NamesRef` directly (rather than a concrete
`ResolvedSchema`) keeps the method usable with any source of names:

```rust
impl Value {
    /// Perform schema resolution using a pre-built name map.
    ///
    /// Equivalent to [`Value::resolve`] / [`Value::resolve_schemata`] but skips
    /// rebuilding the name map on each call. The `names` must have been built from
    /// `schema` (or a set of schemata that includes it), e.g. via
    /// `ResolvedSchema::get_names()`.
    pub fn resolve_with_names(
        self,
        schema: &Schema,
        names: &NamesRef,
    ) -> AvroResult<Self> {
        let enclosing_namespace = schema.namespace();
        self.resolve_internal(schema, names, enclosing_namespace, &None)
    }
}
```

`NamesRef`, `ResolvedSchema`, and `get_names()` are already public, and
`resolve_internal` accepts `&HashMap<Name, S: Borrow<Schema>>`, so a `&NamesRef`
(`&HashMap<Name, &Schema>`) lines up without further changes. Callers obtain the
map from `ResolvedSchema::get_names()`:

```rust
let resolved = ResolvedSchema::try_from(&reader_schema)?;
value.resolve_with_names(&reader_schema, resolved.get_names())?;
```

# Caveat: lifetimes when caching

`ResolvedSchema<'s>` borrows from the schema it was built from, so the pattern
above only works within a single scope: the `ResolvedSchema` cannot be stored next
to an owned `Schema` in the same struct — that would be a self-referential borrow.

The crate already has `ResolvedOwnedSchema` (an ouroboros-based owning variant) for
exactly this case, but it is currently `pub(crate)`. To make a pre-built name map
usable across the lifetime of a long-lived consumer, `ResolvedOwnedSchema` should
be exported alongside this new method. It also exposes `get_names() -> &NamesRef`,
so it feeds `resolve_with_names` directly:

```rust
let resolved = ResolvedOwnedSchema::try_from(reader_schema)?; // owns the schema
value.resolve_with_names(resolved.get_root_schema(), resolved.get_names())?;
```

# Effect

- Removes the repeated schema-tree walk and `NamesRef` allocation on each
  resolution when the schema is reused.
- Does **not** make resolution allocation-free: `resolve` still builds a new
  `Value` (records, arrays, maps, strings), so allocation remains proportional to
  the value size. The saving is limited to the name-map construction.
- Behaviour is otherwise identical to `resolve` / `resolve_schemata`.

Kriskras99:
I ended up using the suggestion from @flxo instead of my own suggestion as it is not clear from a `ResolvedSchema` which schema needs to be resolved against.

This also optimizes `Value::resolve_internal` to take an `Option<&JsonValue>` instead of an `&Option<JsonValue>` which saves a memory load on every function call.

Closes #574

### DIFF
--- a/avro/src/types.rs
+++ b/avro/src/types.rs
@@ -426,9 +426,7 @@ impl Value {
     ) -> bool {
         match self.validate_internal(schema, names, None) {
             Some(reason) => {
-                let log_message =
-                    format!("Invalid value: {self:?} for schema: {schema:?}. Reason: {reason}");
-                error!("{log_message}");
+                error!("Invalid value: {self:?} for schema: {schema:?}. Reason: {reason}");
                 false
             }
             None => true,
--- a/avro/src/types.rs
+++ b/avro/src/types.rs
@@ -790,8 +790,8 @@ impl Value {
             }) => self.resolve_enum(symbols, default, field_default),
             Schema::Array(inner) => self.resolve_array(&inner.items, names, enclosing_namespace),
             Schema::Map(inner) => self.resolve_map(&inner.types, names, enclosing_namespace),
-            Schema::Record(RecordSchema { fields, .. }) => {
-                self.resolve_record(fields, names, enclosing_namespace)
+            Schema::Record(RecordSchema { fields, name, .. }) => {
+                self.resolve_record(fields, names, name.namespace().or(enclosing_namespace))
             }
             Schema::Decimal(DecimalSchema {
                 scale,
--- a/avro/src/types.rs
+++ b/avro/src/types.rs
@@ -370,6 +370,10 @@ impl TryFrom<Value> for JsonValue {
 impl Value {
     /// Validate the value against the given [`Schema`].
     ///
+    /// Note: This will first build a reference map from the schema (see [`ResolvedSchema`]). If this
+    /// function is called more than once for the same schema, it is recommended to use
+    /// [`validate_with_names`](Self::validate_with_names) instead.
+    ///
     /// See the [Avro specification](https://avro.apache.org/docs/++version++/specification)
     /// for the full set of rules of schema validation.
     ///
@@ -380,6 +384,13 @@ impl Value {
     }
 
     /// Validate the value against the given schemata.
+    ///
+    /// Note: This will first build a reference map from the schema (see [`ResolvedSchema`]). If this
+    /// function is called more than once for the same schema, it is recommended to use
+    /// [`validate_with_names`](Self::validate_with_names) instead.
+    ///
+    /// See the [Avro specification](https://avro.apache.org/docs/++version++/specification)
+    /// for the full set of rules of schema validation.
     ///
     /// # Panics
     /// Will panic if the schemata contain unresolved references or duplicate schemas.
@@ -402,6 +413,26 @@ impl Value {
                 None => true,
             },
         )
+    }
+
+    /// Validate the value against the given schema using `names` to resolve any references.
+    ///
+    /// See the [Avro specification](https://avro.apache.org/docs/++version++/specification)
+    /// for the full set of rules of schema validation.
+    pub fn validate_with_names<S: Borrow<Schema> + Debug>(
+        &self,
+        schema: &Schema,
+        names: &HashMap<Name, S>,
+    ) -> bool {
+        match self.validate_internal(schema, names, None) {
+            Some(reason) => {
+                let log_message =
+                    format!("Invalid value: {self:?} for schema: {schema:?}. Reason: {reason}");
+                error!("{log_message}");
+                false
+            }
+            None => true,
+        }
     }
 
     fn accumulate(accumulator: Option<String>, other: Option<String>) -> Option<String> {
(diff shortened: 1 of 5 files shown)


flxo (approved):
LGTM. 

Thanks for the addition - this is really valuable for higher throughputs.

avro/src/types.rs
```diff
@@ -404,6 +415,26 @@ impl Value {
...
+    /// for the full set of rules of schema validation.
+    pub fn validate_with_names<S: Borrow<Schema> + Debug>(
+        &self,
+        schema: &Schema,
+        names: &HashMap<Name, S>,
+    ) -> bool {
+        match self.validate_internal(schema, names, None) {
+            Some(reason) => {
+                let log_message =
+                    format!("Invalid value: {self:?} for schema: {schema:?}. Reason: {reason}");
```
martin-g: Is `log_message` needed ? `error!(...)` will do the same

avro/src/types.rs
```diff
@@ -656,38 +687,70 @@ impl Value {
...
+    /// Note: This will first build a reference map from the schema (see [`ResolvedSchema`]). If this
+    /// function is called more than once for the same schema, it is recommended to use
+    /// [`resolve_with_names`](Self::resolve_with_names) instead.
     ///
     /// See [Schema Resolution](https://avro.apache.org/docs/++version++/specification/#schema-resolution)
-    /// in the Avro specification for the full set of rules of schema
-    /// resolution.
+    /// in the Avro specification for the full set of rules of schema resolution.
     pub fn resolve_schemata(self, schema: &Schema, schemata: Vec<&Schema>) -> AvroResult<Self> {
-        let enclosing_namespace = schema.namespace();
```
martin-g, line 676: Why ?

avro/src/types.rs
Kriskras99: Leftover from copying parts of `validate_schemata`

avro/src/types.rs
Kriskras99: There is no enclosing namespace for `schema`. Either schema has it's own namespace (in which `resolve_internal` will use that) or schema has no namespace in which case this is `None`. So it is useless to set it here.

martin-g: I still do not understand.
The idea of `enclosing_namespace` is to provide the namespace to any (grand)children schemas which do not have their own `"namespace"` attribute.
If we pass `None` to https://github.com/apache/avro-rs/blob/7fda1076ab2c2d351f0e65631165d1b5ec8b5e24/avro/src/types.rs#L752 then `resolve_record()` at https://github.com/apache/avro-rs/blob/7fda1076ab2c2d351f0e65631165d1b5ec8b5e24/avro/src/types.rs#L796 will not know the namespace of the closest parent with a namespace.

Kriskras99: The only schemas for which `schema.namespace()` will return a namespace are:
- `Schema::Ref`
- `Schema::Record`
- `Schema::Enum`
- `Schema::Fixed`
- `Schema::Decimal(Fixed)`
- `Schema::Uuid(Fixed)`
- `Schema::Duration`

`resolve_internal` will only use the `enclosing_namespace` parameter in the following situations:
- `schema == Schema::Ref` to get the fqn of the reference. But if the reference has a namespace that will be used for the fqn
- `schema == Schema::Union`, but directly from `resolve_schemata` the enclosing namespace will not exist
- `schema == Schema::Array`, same as union
- `schema == Schema::Map`, same as union
- `schema == Schema::Record`, it should use the namespace from the record first but I see now that it doesn't. 

I've updated the `Schema::Record` branch to use the namespace from the record first then the enclosing namespace if it does not have one.

avro/src/types.rs
```diff
@@ -729,8 +790,8 @@ impl Value {
             }) => self.resolve_enum(symbols, default, field_default),
             Schema::Array(inner) => self.resolve_array(&inner.items, names, enclosing_namespace),
             Schema::Map(inner) => self.resolve_map(&inner.types, names, enclosing_namespace),
-            Schema::Record(RecordSchema { fields, .. }) => {
-                self.resolve_record(fields, names, enclosing_namespace)
+            Schema::Record(RecordSchema { fields, name, .. }) => {
+                self.resolve_record(fields, names, name.namespace().or(enclosing_namespace))
```
martin-g, line 794: Continuation of https://github.com/apache/avro-rs/pull/579/changes#r3578241892:
because of the change at https://github.com/apache/avro-rs/pull/579/changes#diff-9ce454ff2db8b0477e609ee586622825012be35efe856246c9ec0de5936b3879L682-R717 the `enclosing_namespace` will be `None` here and won't help much.

Kriskras99: It's not useful when used directly from `resolve_schemata` but `resolve_internal` is a recursive function. So at some point in the recursion there can be a Record which does not have its own namespace but has an enclosing namespace.

martin-g (approved):

PR #586, opened 2026-07-13
Validate negative OCF block size before conversion

MitNarodia:
### Summary

The OCF reader currently converts the block size from `i64` to `usize` using an unchecked cast.

This change validates the block size using `usize::try_from()` before conversion and returns the existing `ConvertI64ToUsize` error if a malformed negative block size is encountered.

### Motivation

The Java Avro implementation explicitly validates negative OCF block sizes before using them.

In the current Rust implementation, a negative block size is eventually rejected by `safe_len()`, but only after the unchecked cast wraps the value to `usize`, resulting in a `MemoryAllocation` error.

While the malformed input is still rejected, the reported error reflects a failed memory allocation rather than the actual issue: invalid OCF block metadata. This change validates the block size immediately after it is read and returns a more precise error that directly reflects the malformed metadata, while preserving the behavior for valid inputs.

### Testing

- Added a regression test covering a negative OCF block size.
- Verified that the existing test suite passes with:

```bash
cargo test -p apache-avro --lib --tests --features derive
```


avro/tests/negative_block_size.rs
```diff
@@ -0,0 +1,72 @@
...
+// under the License.
+
+//! Regression: the OCF block reader must reject a negative block size with a
+//! conversion error instead of wrapping it via `as usize`.
+
+use apache_avro::{Error, Reader, Schema, Writer, error::Details, types::Record};
+use apache_avro_test_helper::TestResult;
+
+#[test]
+fn negative_block_size_returns_conversion_error() -> TestResult {
```
martin-g:
```suggestion
fn avro_rs_586_negative_block_size_returns_conversion_error() -> TestResult {
```

martin-g (approved):

Kriskras99:
I've expanded the check to the object count. As this would have duplicate error logic I've replaced `util::read_long` with a new `util::read_ulong`.
I've also moved the test into `block.rs`.
@martin-g can you do one last check on my modifications?

Kriskras99 (approved):

avro/src/util.rs
```diff
@@ -73,8 +73,10 @@ impl MapHelper for Map<String, Value> {
     }
 }
 
-pub(crate) fn read_long<R: Read>(reader: &mut R) -> AvroResult<i64> {
-    zag_i64(reader)
+/// Decode a long from the reader and convert it to a usize.
+pub(crate) fn read_ulong<R: Read>(reader: &mut R) -> AvroResult<usize> {
```
martin-g:
```suggestion
pub(crate) fn read_usize<R: Read>(reader: &mut R) -> AvroResult<usize> {
```
`ulong` sounds weird to me.
I guess you want to make it obvious that it reads "an unsigned long" but then I'd expect `u64`. So, I suggest to just rename it to `read_usize` and use the docstring to explain more.

Kriskras99:
Thank you for your contribution @MitNarodia!

MitNarodia:
Thank you! I really appreciate the review and feedback.

PR #587, opened 2026-07-14
fix: Avro map deterministic encoding

radkawar:
Update `encode_internal` function to sort map entries by key before encoding, ensuring deterministic output.


Kriskras99:
Hi @radkawar, why do you need this? This will severely degrade encoding performance, and the specification gives no guarantees on the key order.

martin-g:
I agree with @Kriskras99 that this is not a good idea.
It will even break interop with other SDKs.

PR #590, opened 2026-07-14
feat: Add optional custom allocator support (CustomAllocator + CountingAllocator)

### LINKED ISSUE #589: Add optional custom allocator support (CustomAllocator + CountingAllocator)
### Problem/motivation

`apache-avro` currently offers no first-class hook for customizing or observing heap allocation. The existing `util::max_allocation_bytes` guard only rejects a single oversized length prefix at decode time; it gives no visibility into the actual live/peak memory a workload uses, and no reusable building block for users who want to plug in custom allocation behavior.

### Proposed solution

Add an **opt-in** cargo feature `custom_allocator` (off by default) exposing a new `apache_avro::alloc` module with two custom `GlobalAlloc` implementations:

1. **`CustomAllocator`** — a minimal allocator that delegates to the `System` allocator. A ready-made base users can install as `#[global_allocator]`   or extend to add their own behavior (logging, pooling, etc.) without reimplementing platform allocation logic.
2. **`CountingAllocator<A = System>`** — wraps any inner allocator and tracks live bytes (`allocated()`), peak (`peak_allocated()`), cumulative
(`total_allocated()`), and allocation/deallocation counts, with `reset_peak()` for scoped measurement. `const fn` constructors let it back a `#[global_allocator]` static.

Both are opt-in and inert unless the user installs one as the global allocator, so there is zero impact when the feature is disabled. `CountingAllocator`
complements `max_allocation_bytes` by measuring real process memory, including pressure from many individually-small allocations.

### Alternatives considered

- A plain user-side `#[global_allocator]`: works, but every user must re-implement the counting/peak logic; std provides no counting allocator.
- Re-exporting a third-party allocator (jemalloc/mimalloc): low novelty and users can already do that themselves.

siddharthmittal13:
## What

Closes #589.

Adds an **opt-in** cargo feature `custom_allocator` (off by default) that exposes a new `apache_avro::alloc` module with two custom `GlobalAlloc` implementations:

1. **`CustomAllocator`** — a minimal allocator that delegates every operation to the `System` allocator. A ready-made base users can install directly as `#[global_allocator]` or extend to add their own behavior (logging, pooling, ...) without reimplementing platform allocation logic.
2. **`CountingAllocator<A = System>`** — wraps any inner allocator (defaults to `System`) and tracks:
   - live bytes — `allocated()`
   - high-water mark — `peak_allocated()`
   - cumulative bytes — `total_allocated()`
   - allocation / deallocation counts
   - `reset_peak()` for scoped measurement

   `const fn` constructors (`CountingAllocator::system()`, `CountingAllocator::new(inner)`) let it back a `#[global_allocator]` static.

## Why

`apache-avro` currently offers no first-class hook for customizing or observing heap allocation. The existing `util::max_allocation_bytes` guard only rejects a *single* oversized length prefix at decode time — it gives no visibility into the actual live/peak memory a workload uses, and no reusable building block for custom allocation behavior.

`CountingAllocator` complements `max_allocation_bytes` by measuring real process memory, including pressure that builds up across many individually-small allocations.

## Notes

- **Fully opt-in / zero cost when disabled.** The module is gated behind the `custom_allocator` feature and is inert unless the user installs one of the types as their `#[global_allocator]`. The default build is unchanged.
- Counters use relaxed atomics (observability, not hard real-time accounting).

## Testing

- New unit tests for `CustomAllocator` (delegation) and `CountingAllocator` (alloc/dealloc, peak high-water mark, `reset_peak`, realloc growth/shrink) — all passing.
- Both module doc examples compile (`--doc`).
- `cargo fmt` and `cargo clippy --features custom_allocator --lib` are clean.
- Default (feature-off) build verified.

### DIFF
--- a/avro/Cargo.toml
+++ b/avro/Cargo.toml
@@ -30,6 +30,7 @@ documentation.workspace = true
 
 [features]
 bzip = ["dep:bzip2"]
+custom_allocator = []
 derive = ["dep:apache-avro-derive"]
 snappy = ["dep:crc32fast", "dep:snap"]
 xz = ["dep:liblzma"]
--- a/avro/src/alloc.rs
+++ b/avro/src/alloc.rs
@@ -1,0 +1,319 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+//! Optional allocation tracking support.
+//!
+//! This module is only available when the `custom_allocator` feature is enabled.
+//!
+//! It provides [`CountingAllocator`], a [`GlobalAlloc`] adapter that wraps any
+//! other allocator (the [`System`] allocator by default) and keeps live, peak
+//! and cumulative allocation statistics. Installing it as the process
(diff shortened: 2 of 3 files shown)


martin-g:
It is not apache-avro's business to provide such functionalities.
You could use any crate specializing in these matters in your application.

PR #625, opened 2026-08-10
chore(docs): parameterize RC version in RELEASE.md and add archive prefix

iemejia:
## What

Improve the release instructions in `RELEASE.md`:

- Introduce an `AVRO_RC_VERSION` variable (e.g. `RC0`) and use it for the git tag, tarball name, and SVN upload steps.
- Add `--prefix=apache-avro-rs-$AVRO_RS_VERSION-$AVRO_RC_VERSION/` to `git archive` so the tarball extracts into a top-level directory instead of dumping files into the current directory.
- Quote variable expansions and guard `cd` calls, per shellcheck (SC2086, SC2164).

## Why

The previous `git archive` command produced a tarball without a prefix directory, and the RC identifier was hardcoded in several places.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -5,11 +5,12 @@
 # in avro-rs Git clone
 git switch main && git pull --rebase
 export AVRO_RS_VERSION=0.22.0
+export AVRO_RC_VERSION=RC0
 # update the version in Cargo.toml files to $AVRO_RS_VERSION
 git commit -am "Update version to $AVRO_RS_VERSION"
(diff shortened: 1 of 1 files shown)


martin-g (approved):

Kriskras99 (approved):

Kriskras99:
Can I also suggest (not for this PR) to publish the release candidates on crates.io?
This will make it significantly easier for users to test the code.

martin-g:
> Can I also suggest (not for this PR) to publish the release candidates on crates.io?

Do other projects publish RCs ? Do you have some examples ?

Kriskras99:
Yes, and Cargo has full support for release candidates (it won't auto update to them).
[`rand`](https://crates.io/crates/rand/versions) is one example.

PR #633, opened 2026-08-22
chore: migrate HawkEye to v7

tisonkun:
## Summary

- migrate the HawkEye configuration and integration to v7
- install HawkEye from its latest release without pinning the tool version
- validate the migrated configuration with HawkEye v7 (176 files, 0 changes, 0 conflicts, 0 unsupported)
- exclude one header-only Rust placeholder where HawkEye's separator conflicts with `rustfmt`
- `cargo fmt --all --check` passes

### DIFF
--- a/.github/workflows/test-lang-rust-ci.yml
+++ b/.github/workflows/test-lang-rust-ci.yml
@@ -40,7 +40,7 @@ jobs:
     steps:
       - uses: taiki-e/install-action@7f4eb899022d8fe70b20c4f3de697aa85c309026 # v2.85.11
         with:
-          tool: hawkeye@6.3
+          tool: hawkeye
       - name: Checkout
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
--- a/licenserc.toml
+++ b/licenserc.toml
@@ -15,13 +15,22 @@
 # specific language governing permissions and limitations
 # under the License.
 
-headerPath = "Apache-2.0-ASF.txt"
+[header]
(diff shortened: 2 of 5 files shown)


.github/workflows/test-lang-rust-ci.yml
```diff
@@ -40,7 +40,7 @@ jobs:
     steps:
       - uses: taiki-e/install-action@7f4eb899022d8fe70b20c4f3de697aa85c309026 # v2.85.11
         with:
-          tool: hawkeye@6.3
+          tool: hawkeye
```
tisonkun, line 43:
```suggestion
          tool: hawkeye@7.0.0
```

... if you'd prefer to pin a version.

licenserc.toml
```diff
@@ -15,13 +15,22 @@
...
+[header]
+builtin = "Apache-2.0-ASF"
+
+[files]
 excludes = [
   "*.stderr",
   "*.expanded.rs",
   "*.avro",
+  "**/*.lock",
+  "**/*.md",
```
martin-g, line 27: It would be useful to check the doc files too.
I will do it in a follow-up!

martin-g (approved):

.github/workflows/test-lang-rust-ci.yml
martin-g: Let's use `latest`.
We can pin it if causes any problems!

martin-g:
Thank you, @tisonkun !

tisonkun:
Thanks for your review and merging @martin-g!

PR #639, opened 2026-08-25
fix: Nested collections multiply the 512MB allocation budget arbitrarily

martin-g:
decode.rs::decode_internal (for Schema::Array & Map): safe_collection_len bounds each individual Vec/HashMap against max_allocation_bytes, but no state tracks the running total across the decoded value tree. Elements that cost zero wire bytes (null, empty records, zero-size fixed) let each ~7-byte inner array<null> block materialize ~512MB of live Value slots, so array<array<null>> from an attacker OCF header holds K x 512MB simultaneously from ~130 datum bytes.

Reported-by: Security scans

PR #640, opened 2026-08-25
fix: Fixed-schema decode allocates attacker-chosen size with no budget check

martin-g:
Cap allocations to max_allocation_bytes when decoding OCF / deserializing schemas

Reported-by: Security scans

### DIFF
--- a/avro/src/decode.rs
+++ b/avro/src/decode.rs
@@ -612,6 +612,29 @@ mod tests {
     }
 
     #[test]
+    fn avro_rs_640_test_decode_fixed_size_above_budget_is_rejected() -> TestResult {
+        use crate::schema::Name;
+
+        // The schema (and with it the fixed size) can be attacker-supplied
+        // via an OCF header; the decoder must refuse to allocate more than
+        // the budget, before reading a single payload byte.
+        let schema = Schema::Fixed(
+            FixedSchema::builder()
+                .name(Name::new("huge")?)
+                .size(usize::MAX / 2)
+                .build(),
+        );
+        let mut input: &[u8] = &[0u8; 4];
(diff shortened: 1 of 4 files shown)


avro/src/decode.rs
```diff
@@ -226,7 +226,7 @@ pub(crate) fn decode_internal<R: Read, S: Borrow<Schema>>(
             }
         }
         Schema::Fixed(FixedSchema { size, .. }) => {
-            let mut buf = vec![0u8; *size];
+            let mut buf = vec![0u8; safe_len(*size)?];
```
martin-g: This should use `DecodeContext::debit()` from https://github.com/apache/avro-rs/pull/639

avro/src/schema/parser.rs
```diff
@@ -772,6 +772,11 @@ impl Parser {
                 .ok_or_else(|| Details::GetFixedSizeFieldPositive(size.clone())),
             None => Err(Details::GetFixedSizeField),
         }?;
+        // A fixed schema's size directs decoders to allocate that many bytes,
+        // and the schema itself may be attacker-supplied (e.g. an OCF header).
+        // Always bound it before allocating.
+        let size = usize::try_from(size).map_err(|e| Details::ConvertU64ToUsize(e, size))?;
+        let size = safe_len(size)?;
```
Kriskras99: I don't think the schema parser is the right place to do this. This should only be done in de reader/decoder.

Kriskras99 (changes requested):

avro/src/schema/parser.rs
martin-g: Agreed!
Reverted the check!

Kriskras99 (approved):

PR #642, opened 2026-08-25
fix: No recursion depth limit in decode_internal / resolve_internal / serde deserializer

martin-g:
A recursive schema can lead to stack overflow (i.e. an abort) in OCF decoding, Value resolving and Schema deserialization. Add `util::max_decode_recursion_depth(usize)` to be able to control the recursipn depth

Reported-by: Security scans

### DIFF
--- a/avro/src/serde/deser_schema/mod.rs
+++ b/avro/src/serde/deser_schema/mod.rs
@@ -20,7 +20,7 @@ use std::{borrow::Borrow, collections::HashMap, io::Read};
 use serde::de::{Deserializer, Visitor};
 
 use crate::{
-    Error, Schema,
+    AvroResult, Error, Schema,
     decode::decode_len,
     error::Details,
     schema::{DecimalSchema, InnerDecimalSchema, Name, UnionSchema, UuidSchema},
@@ -60,11 +60,26 @@ pub struct Config<'s, S: Borrow<Schema>> {
 
 impl<'s, S: Borrow<Schema>> Config<'s, S> {
     /// Get the schema for this name.
-    fn get_schema(&self, name: &Name) -> Result<&'s Schema, Error> {
+    fn get_schema(&self, name: &Name) -> AvroResult<&'s Schema> {
         self.names
             .get(name)
             .map(Borrow::borrow)
             .ok_or_else(|| Details::SchemaResolutionError(name.clone()).into())
+    }
+
+    /// Increments the recursion depth counter and checks whether it exceeds the defined recursion limit.
+    ///
+    /// # Errors
+    /// Returns an error of type `AvroError` with `DecodeRecursionLimit` details if the recursion depth
+    /// exceeds the configured maximum.
+    fn recurse(&mut self) -> AvroResult<()> {
+        self.recursion_depth += 1;
+        let maximum = decode_recursion_limit();
+        if self.recursion_depth > maximum {
+            Err(Details::DecodeRecursionLimit { maximum }.into())
+        } else {
+            Ok(())
+        }
     }
 }
 
@@ -95,13 +110,8 @@ impl<'s, 'r, R: Read, S: Borrow<Schema>> SchemaAwareDeserializer<'s, 'r, R, S> {
         reader: &'r mut R,
         schema: &'s Schema,
         mut config: Config<'s, S>,
-    ) -> Result<Self, Error> {
-        // Bound the recursion depth to guard against stack exhaustion
-        config.recursion_depth += 1;
-        let maximum = decode_recursion_limit();
-        if config.recursion_depth > maximum {
-            return Err(Details::DecodeRecursionLimit { maximum }.into());
-        }
+    ) -> AvroResult<Self> {
+        config.recurse()?;
         if let Schema::Ref { name } = schema {
             let schema = config.get_schema(name)?;
             Ok(Self {
@@ -132,7 +142,8 @@ impl<'s, 'r, R: Read, S: Borrow<Schema>> SchemaAwareDeserializer<'s, 'r, R, S> {
     /// Create a new deserializer with the existing reader and config.
     ///
     /// This will resolve the schema if it is a reference.
-    fn with_different_schema(mut self, schema: &'s Schema) -> Result<Self, Error> {
+    fn with_different_schema(mut self, schema: &'s Schema) -> AvroResult<Self> {
+        self.config.recurse()?;
         self.schema = if let Schema::Ref { name } = schema {
             self.config.get_schema(name)?
         } else {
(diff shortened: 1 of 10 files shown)


avro/src/util.rs
```diff
@@ -32,6 +33,15 @@ use std::{
 pub const DEFAULT_MAX_ALLOCATION_BYTES: usize = 512 * 1024 * 1024;
 static MAX_ALLOCATION_BYTES: OnceLock<usize> = OnceLock::new();
 
+/// Maximum recursion depth when decoding or resolving Avro-encoded values.
+///
+/// This protects against stack exhaustion (an abort, not a catchable error) from deeply nested
+/// data: a recursive schema lets an attacker drive one recursion level with roughly one wire byte.
+///
+/// See [`max_decode_recursion_depth`] to change this limit.
+pub const DEFAULT_MAX_DECODE_RECURSION_DEPTH: usize = 32;
```
Kriskras99, line 41: At 62 the `decode::tests::avro_rs_642_resolve_recursion_depth_is_bounded` test overflows.
At 558(!) the `recursion_depth_deser::avro_rs_642_recursion_depth_is_bounded` test overflows.
So I think there might be some optimisation room left for the decode implementation (not for this PR obviously).

avro/src/util.rs
```diff
@@ -32,6 +33,15 @@ use std::{
 pub const DEFAULT_MAX_ALLOCATION_BYTES: usize = 512 * 1024 * 1024;
 static MAX_ALLOCATION_BYTES: OnceLock<usize> = OnceLock::new();
 
+/// Maximum recursion depth when decoding or resolving Avro-encoded values.
+///
+/// This protects against stack exhaustion (an abort, not a catchable error) from deeply nested
+/// data: a recursive schema lets an attacker drive one recursion level with roughly one wire byte.
```
Kriskras99:
```suggestion
/// This protects against stack exhaustion aborts from deeply nested data.
```

The attack information is a nice tidbit, but practically useless for the user.

avro/src/serde/deser_schema/mod.rs
Kriskras99: This should also increment the recursion depth as it does not use `new`

avro/src/serde/deser_schema/mod.rs
```diff
@@ -50,6 +50,12 @@ pub struct Config<'s, S: Borrow<Schema>> {
     pub names: &'s HashMap<Name, S>,
     /// Was the data serialized with `human_readable`.
     pub human_readable: bool,
+    /// Current recursion depth of the deserializer.
+    ///
+    /// Every nesting level of the deserialized value creates a new
+    /// [`SchemaAwareDeserializer`], which increments this; the depth is
+    /// bounded by [`crate::util::max_decode_recursion_depth`].
+    pub(crate) recursion_depth: usize,
```
Kriskras99:
```suggestion
    pub recursion_depth: usize,
```
`Config` is not a public type, so `pub(crate)` does nothing

avro/src/serde/deser_schema/mod.rs
```diff
@@ -88,8 +94,16 @@ impl<'s, 'r, R: Read, S: Borrow<Schema>> SchemaAwareDeserializer<'s, 'r, R, S> {
...
+        mut config: Config<'s, S>,
     ) -> Result<Self, Error> {
+        // Bound the recursion depth so deeply nested (possibly hostile) data
+        // yields an error instead of exhausting the stack. A recursive schema
+        // lets roughly one wire byte drive one nesting level.
+        config.recursion_depth += 1;
+        let maximum = decode_recursion_limit();
+        if config.recursion_depth > maximum {
+            return Err(Details::DecodeRecursionLimit { maximum }.into());
+        }
```
Kriskras99: I think it would be better to split this off into an implementation on `Config`, otherwise this would need to be duplicated in `with_different_schema`

avro/src/serde/deser_schema/mod.rs
```diff
@@ -88,8 +94,16 @@ impl<'s, 'r, R: Read, S: Borrow<Schema>> SchemaAwareDeserializer<'s, 'r, R, S> {
     pub fn new(
         reader: &'r mut R,
         schema: &'s Schema,
-        config: Config<'s, S>,
+        mut config: Config<'s, S>,
     ) -> Result<Self, Error> {
+        // Bound the recursion depth so deeply nested (possibly hostile) data
+        // yields an error instead of exhausting the stack. A recursive schema
+        // lets roughly one wire byte drive one nesting level.
```
Kriskras99:
```suggestion
        // Bound the recursion depth to guard against stack exhaustion
```

avro/tests/recursion_depth_deser.rs
```diff
@@ -0,0 +1,51 @@
...
+// specific language governing permissions and limitations
+// under the License.
+
+use apache_avro::Schema;
+use apache_avro::reader::datum::GenericDatumReader;
+use apache_avro::util::max_decode_recursion_depth;
+use apache_avro_test_helper::TestResult;
+use serde::{Deserialize, Serialize};
+
+// This is an IT test because it sets the default recursion depth limit (OnceLock).
```
Kriskras99:
```suggestion
// This is an integration test because it sets the default recursion depth limit (OnceLock).
```

avro/tests/recursion_depth_value_resolve.rs
```diff
@@ -0,0 +1,53 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+use apache_avro::Schema;
+use apache_avro::types::Value;
+use apache_avro::util::max_decode_recursion_depth;
+use apache_avro_test_helper::TestResult;
+
+// This is an IT test because it sets the default recursion depth limit (OnceLock).
```
Kriskras99:
```suggestion
// This is an integration test because it sets the default recursion depth limit (OnceLock).
```
This file can be merged with the other integration test right?

Kriskras99 (changes requested):

avro/tests/recursion_depth_value_resolve.rs
martin-g: At the moment they use the same depth, so - yes!

avro/src/serde/deser_schema/mod.rs
martin-g: Done

avro/src/serde/deser_schema/mod.rs
martin-g: Done!
Thanks for catching this!

Kriskras99 (approved):

PR #656, opened 2026-09-04
feat!: Make schema parser reuse the JSON allocations

### LINKED ISSUE #654: regression bug, `logicalType` is dropped from custom attributes
Hey, in version 0.22.0 the attribute `logicalType` was dropped entirely. I believe this is a regression, and is incorrect behaviour. I am pretty new to Avro, so I'm looking for some help on asserting if this is right. I have replicated the behaviour in both the Java and Rust Avro projects, and the results are different. 

# Steps to reproduce

 * You create a valid Json schema which contains a `logicalType` attribute.
 * You parse the Json into an Avro type.
 * You then serialise the schema back to Json.

## What do I expect?

 * The `logicalType` should still be present.
 * The behaviour should match the Java implementation.

## What happened?

 * The `logicalType` is missing.
 * The behaviour does not match the Java implementation.
 * The `logicalType` is entirely dropped, making it impossible to correctly identify `map` types for Iceberg.

# Context

I am trying to upgrade the Iceberg repo from Avro 0.21.0 to 0.22.0. One test there is failing, which is for map types. There is no Avro equivalent, so the Iceberg spec asks for an array of key/value pairs + adding `logicalType: map` to the schema (see https://iceberg.apache.org/spec/#avro).

Avro 0.22.0 drops `logicalType`, which makes it impossible to truly identify the map type.

The Java version preserved the `logicalType`. The Java version has tests which perform a round trip and asserts what they started and ended with are the same. This implies the `logicalType` is preserved. See: https://github.com/apache/avro/blob/main/lang/java/avro/src/test/java/org/apache/avro/TestLogicalType.java#L190-L195

Finally the original PR was about fixing **duplicate** outputs of `logicalType`, not removing `logicalType` entirely. I suspect removing it entirely has crept in as a mistake.

# Tests to reproduce

## For Java

Here is a *passing* test for the Java Apache Avro repo at (https://github.com/apache/avro). It's a bit noisy. The key part is `logicalType` is at the end of the Json schema, and turning it from the schema to string yields the original schema.,
```java
@Test
void logicalTypeTestReplication() {
  String rawSchema = "{\n" + "  \"type\" : \"array\",\n" + "  \"items\" : {\n" + "    \"type\" : \"record\",\n"
      + "    \"name\" : \"k12_v13\",\n" + "    \"fields\" : [ {\n" + "      \"name\" : \"key\",\n"
      + "      \"type\" : \"int\",\n" + "      \"field-id\" : 12\n" + "    }, {\n" + "      \"name\" : \"value\",\n"
      + "      \"type\" : \"string\",\n" + "      \"field-id\" : 13\n" + "    } ]\n" + "  },\n"
      + "  \"logicalType\" : \"map\"\n" + "}";

  Schema schema = SchemaParser.parseSingle(rawSchema);
  String output = schema.toString(true);

  assertEquals(rawSchema, output, "logicalType and custom properties should survive parsing");
}
```

## For Rust

Here is the same test for your repo:
```rust
#[test]
fn it_should_preserve_map_logical_type_on_outer_item() -> TestResult {
    let raw_schema = r#"{
        "type": "array",
        "logicalType": "map",
        "items": {
            "type": "record",
            "name": "k12_v13",
            "fields": [
                {
                    "name": "key",
                    "type": "int",
                    "field-id": 12
                },
                {
                    "name": "value",
                    "type": "string",
                    "field-id": 13
                }
            ]
        }
    }"#;

    let output = serde_json::to_string_pretty(&schema).unwrap();
    pretty_assertions::assert_eq!(r#"{
  "type": "array",
  "logicalType": "map", // <-- this is missing!
  "items": {
    "type": "record",
    "name": "k12_v13",
    "fields": [
      {
        "name": "key",
        "type": "int",
        "field-id": 12
      },
      {
        "name": "value",
        "type": "string",
        "field-id": 13
      }
    ]
  }
}"#, output);

    let logical_type = schema
        .custom_attributes()
        .unwrap()
        .get("logicalType");

    // !!! This fails !!!
    assert!(logical_type.is_some());

    Ok(())
}
```

Blocks: github.com/apache/iceberg-rust/issues/3063

(If someone can double check and confirm my findings, then I would be happy to make a PR that fixes this.)

Kriskras99:
This changes the schema parser from iterating over the JSON into consuming it. This allows reusing the existing String allocations (and also collection allocations if the sizes match and the compiler can make it work).
Because we remove everything from the JSON while consuming it, custom attributes is just everything leftover. Therefore this also fixes https://github.com/apache/avro-rs/issues/654

I've tried my best to improve the error messages where possible.

This is a breaking change because the input for `Schema::parse` changes         
from a reference to an owned `Value`. Users can fix their usage by        
calling `.clone()` where needed.

@JosephLenton I've added your schema to one of the tests, can you verify everything works in your application?


avro/src/schema/mod.rs
```diff
@@ -546,7 +546,16 @@ impl Schema {
             let json = json.as_ref();
             let schema: JsonValue = serde_json::from_str(json).map_err(Details::ParseSchemaJson)?;
             if let JsonValue::Object(inner) = &schema {
-                let name = Name::parse(inner, None)?;
+                // Only clone the keys needed for the name parsing, can be a significant time/memory
```
```suggestion
                // Only clone the values needed for the name parsing, can be a significant time/memory
```

avro/src/schema/mod.rs
```diff
@@ -588,7 +597,16 @@ impl Schema {
...
```
```suggestion
                // Only clone the values needed for the name parsing, can be a significant time/memory
```

avro/src/schema/parser.rs
```diff
@@ -450,15 +438,18 @@ impl Parser {
...
+                _ => {
+                    // re-insert unknown logical type
+                    complex.insert(key, Value::String(t));
+                }
             },
             // The spec says to ignore invalid logical types and just pass through the
             // underlying type. It is unclear whether that applies to this case or not, where the
             // `logicalType` is not a string.
-            Some(value) => return Err(Details::GetLogicalTypeFieldType(value.clone()).into()),
+            Some((_, value)) => return Err(Details::GetLogicalTypeFieldType(value.clone()).into()),
```
martin-g:
```suggestion
            Some((_, value)) => return Err(Details::GetLogicalTypeFieldType(value).into()),
```
does it need the clone ?!

martin-g (approved):

JosephLenton:
Hey @Kriskras99 , I have pulled down your PR and can confirm my Avro test passes and so does my Iceberg test! Thank you very much for getting this done.

If it helps the tests I have are these. They failed before, and pass now. Feel free to add them to your PR (or a followup) if you find them helpful.

I had them in: `avro/tests/schema.rs`
```rust
#[test]
fn it_should_preserve_map_logical_type_on_outer_item() -> TestResult {
    let raw_schema = r#"{
        "type": "array",
        "logicalType": "map",
        "items": {
            "type": "record",
            "name": "k12_v13",
            "fields": [
                {
                    "name": "key",
                    "type": "int",
                    "field-id": 12
                },
                {
                    "name": "value",
                    "type": "string",
                    "field-id": 13
                }
            ]
        }
    }"#;

    let output = serde_json::to_string_pretty(&schema).unwrap();
    pretty_assertions::assert_eq!(
        r#"{
  "type": "array",
  "items": {
    "type": "record",
    "name": "k12_v13",
    "fields": [
      {
        "name": "key",
        "type": "int",
        "field-id": 12
      },
      {
        "name": "value",
        "type": "string",
        "field-id": 13
      }
    ]
  },
  "logicalType": "map"
}"#,
        output
    );

    let logical_type = schema.custom_attributes().unwrap().get("logicalType");
    assert_eq!(
        logical_type,
        Some(&serde_json::Value::String("map".to_string()))
    );

    Ok(())
}

#[test]
fn it_should_preserve_map_logical_type_on_inner_item() -> TestResult {
    let raw_schema = r#"{
        "type": "record",
        "name": "test_record",
        "fields": [
            {
            "name": "example_map",
            "type": {
                "type": "array",
                "logicalType": "map",
                "items": {
                    "type": "record",
                    "name": "k12_v13",
                    "fields": [
                        {
                            "name": "key",
                            "type": "int",
                            "field-id": 12
                        },
                        {
                            "name": "value",
                            "type": "string",
                            "field-id": 13
                        }
                    ]
                }
            }
            }
        ]
    }"#;

    let Schema::Record(record) = &schema else {
        panic!("Expected a record schema");
    };
    let example_map_schema = &record.fields[0].schema;
    let logical_type = example_map_schema
        .custom_attributes()
        .unwrap()
        .get("logicalType");
    assert_eq!(
        logical_type,
        Some(&serde_json::Value::String("map".to_string()))
    );

    Ok(())
}
```

Kriskras99:
Perfect!

PR #657, opened 2026-09-04
chore: Remove code that is no longer needed

Kriskras99:


martin-g (approved):
How does this work now ?
I guess some version of Rust fixed it ?!

Kriskras99:
Or maybe because there are other imports now? I don't know the original issue, but am happy it works without.

PR #660, opened 2026-09-05
Improve Snappy CRC32 with crc-fast

### LINKED ISSUE #659: Replace crc32fast with crc-fast for faster Snappy CRC32
## Motivation

The Rust Snappy codec currently uses `crc32fast` only for the Avro block checksum:

- on compression, it computes CRC32 over the uncompressed block and appends the big-endian 4-byte checksum;
- on decompression, it computes CRC32 over the decoded block and compares it with that trailer.

That makes this a narrow, low-risk replacement candidate: there are only two call sites, both operate on a contiguous byte slice, and neither needs streaming/incremental hashing state.

## Why `crc-fast`

`crc-fast` 1.10.0 provides a specialized `crc32_iso_hdlc(&[u8]) -> u32` helper. CRC-32/ISO-HDLC is the standard CRC32 variant used by `crc32fast` and by Avro Snappy block checksums. The standard check vector `b"123456789" -> 0xcbf43926` matches the expected algorithm.

The upstream crate is also built around SIMD acceleration:

- acceleration paths for `aarch64`, `x86_64`, and `x86`;
- safe table-based fallback for other targets;
- one-shot helpers for common algorithms, avoiding the extra hasher object/update/finalize flow used in the current code.

For this repository's actual call pattern, a local microbenchmark on `Darwin arm64` (`Mac17,6`, `rustc 1.98.1`) compared the current `crc32fast::Hasher::new(); update(); finalize()` flow with `crc_fast::crc32_iso_hdlc(data)`. The checksum was asserted equal for every case.

| Block size | crc32fast | crc-fast | Speedup |
| ---: | ---: | ---: | ---: |
| 128 B | 12.12 GiB/s | 21.02 GiB/s | 1.73x |
| 1 KiB | 10.93 GiB/s | 70.03 GiB/s | 6.41x |
| 16 KiB | 22.56 GiB/s | 81.98 GiB/s | 3.63x |
| 1 MiB | 33.27 GiB/s | 101.22 GiB/s | 3.04x |
| 16 MiB | 33.42 GiB/s | 97.76 GiB/s | 2.93x |

This matters for Snappy because CRC32 is paid on every compressed block in both write and read paths. The larger the uncompressed block, the more directly the checksum cost shows up next to Snappy's own encode/decode work.

## Trade-offs

The main trade-off is MSRV: `crc-fast` 1.10.0 declares `rust-version = "1.89"`, while this workspace currently declares and tests `1.88.0`. A correct replacement therefore needs to bump the workspace MSRV and the CI MSRV entry to `1.89.0`.

Dependency-wise, using `crc-fast` with `default-features = false, features = ["std"]` avoids its default `ffi`/`panic-handler` features while enabling the std-backed runtime SIMD detection needed for the performance win. This adds its `std` dependency chain (`digest 0.10`, `crypto-common 0.1`, `generic-array`, and `spin`) and removes `crc32fast`.

## Suggested change

- Replace the optional Snappy dependency `crc32fast` with `crc-fast = "1.10.0"`.
- Use `crc_fast::crc32_iso_hdlc(...)` at the two Snappy checksum call sites.
- Add a regression test that verifies the Snappy trailer for `b"123456789"` is the big-endian CRC-32/ISO-HDLC check value `0xcbf43926`.
- Bump workspace and CI MSRV from `1.88.0` to `1.89.0`.

## Local validation

- `cargo check -p apache-avro --features snappy`
- `cargo test -p apache-avro --features snappy,derive snappy`
- `cargo fmt --all --check`
- `cargo build --all-features --all-targets`
- `cargo check -p apache-avro --target aarch64-unknown-linux-gnu --features snappy`

Not completed locally:

- `wasm32-unknown-unknown` CI target validation. Installing the target locally stalled while downloading `rust-std`, so this should still be watched in CI.

houseme:
## Summary

- replace the optional Snappy `crc32fast` dependency with `crc-fast` 1.10.0
- use `crc_fast::crc32_iso_hdlc(...)` for Avro Snappy block CRC32 calculation
- add a regression test for the standard CRC-32/ISO-HDLC trailer bytes
- add `avro/benches/crc32_snappy.rs` to compare `crc32fast` and `crc-fast` across raw CRC, Snappy compression + CRC, and Snappy decompression + CRC
- document aarch64 and x86_64 benchmark results in `avro/benches/crc32_snappy_results.md`
- bump workspace and CI MSRV from 1.88.0 to 1.89.0 because `crc-fast` 1.10.0 requires Rust 1.89

Closes #659.

## Rationale

The Snappy codec only needs CRC32 over contiguous decoded/uncompressed blocks, so `crc-fast`'s specialized one-shot `crc32_iso_hdlc` helper fits the existing call sites directly. The algorithm remains CRC-32/ISO-HDLC, and the new test locks the public Avro trailer format to the standard `b"123456789" -> 0xcbf43926` check value encoded as big-endian bytes.

The benchmark now covers both isolated CRC cost and Snappy-shaped encode/decode cost. On local aarch64, Snappy compression + CRC improved up to 1.58x and decompression + CRC up to 2.29x. On the x86_64 EPYC test host, Snappy compression + CRC improved up to 4.32x and decompression + CRC up to 3.72x.

The SIMD reason is architecture-specific. On aarch64, `crc-fast` can use PMULL/AES folding, with SHA3/EOR3-assisted variants when available. On x86_64, SSE4.2 CRC32 is CRC32C/ISCSI rather than CRC-32/ISO-HDLC, so the relevant acceleration for this Avro checksum is PCLMULQDQ / VPCLMULQDQ carry-less multiply.

`crc-fast` is configured with `default-features = false, features = ["std"]` to avoid its default `ffi`/`panic-handler` features while retaining std-backed runtime SIMD detection.

## Testing

- `cargo check -p apache-avro --features snappy`
- `cargo test -p apache-avro --features snappy,derive snappy`
- `cargo fmt --all --check`
- `cargo build --all-features --all-targets`
- `cargo check -p apache-avro --target aarch64-unknown-linux-gnu --features snappy`
- `cargo bench -p apache-avro --bench crc32_snappy --features snappy,derive --no-run`
- `cargo bench -p apache-avro --bench crc32_snappy --features snappy,derive` on local aarch64
- `cargo bench -p apache-avro --bench crc32_snappy --features snappy,derive` on `azure-4074205200` x86_64

Not completed locally:

- `wasm32-unknown-unknown` CI target validation. Installing the target locally stalled while downloading `rust-std`, so this should still be watched in CI.


Kriskras99:
Do you actually measure a difference when decoding or encoding Avro files?

houseme:
Yes. I added a dedicated benchmark for this in `avro/benches/crc32_snappy.rs`.

It now measures three paths on both aarch64 and x86_64:

- raw CRC-32/ISO-HDLC;
- Snappy compression plus the Avro CRC32 trailer;
- Snappy decompression plus CRC32 verification.

The results are documented in `avro/benches/crc32_snappy_results.md`. The short version is that the isolated CRC gain does carry into the Snappy-shaped encode/decode paths, especially from 1 KiB upward. On the tested aarch64 machine, Snappy compress + CRC improved up to 1.58x and decompress + CRC up to 2.29x. On the tested x86_64 EPYC host, compress + CRC improved up to 4.32x and decompress + CRC up to 3.72x.

The SIMD detail is architecture-specific: on aarch64 the improvement comes from `crc-fast` using PMULL/AES folding, with SHA3/EOR3-assisted variants when available. On x86_64, SSE4.2 CRC32 is CRC32C/ISCSI rather than CRC-32/ISO-HDLC, so the relevant acceleration is the PCLMULQDQ / VPCLMULQDQ carry-less multiply family.

Kriskras99 (approved):
Thank you for providing a benchmark! I still think you won't see much of a performance difference because the Avro encoder/decoder will be the limiting factor (unless your sending Avro objects with large `bytes`/`string`).

I have removed the benchmark in a new commit, so it will still be available in the Git history. The benchmark is good, but something we are going to use only once.

If @martin-g agrees with the MSRV bump from `1.88` to `1.89` this can be merged

martin-g:
> If @martin-g agrees with the MSRV bump from `1.88` to `1.89` this can be merged

Sure!
My approach so far was to support `stable` - 5, i.e. current stable is 1.98, so Avro MSRV could be even 1.93.
There is no hard rule here. Just no need to require very recent version if apache_avro or its dependencies do not need something from recent Rust.

martin-g:
@houseme There is a merge conflict in Cargo.lock

martin-g (approved):

martin-g:
Thank you, @houseme !

PR #663, opened 2026-09-06
feat: Optimize UUID decoding in `decode.rs`

Kriskras99:
This way we don't have to do any allocations or check that the bytes are valid UTF-8.

### DIFF
--- a/avro/src/decode.rs
+++ b/avro/src/decode.rs
@@ -232,9 +232,7 @@ fn decode_internal_body<R: Read, S: Borrow<Schema>>(
             let len = decode_len(reader)?;
             if len == 16 {
                 let mut buf = [0u8; 16];
-                reader
-                    .read_exact(&mut buf)
-                    .map_err(|e| Details::ReadFixed(e, 16))?;
+                reader.read_exact(&mut buf).map_err(Details::ReadBytes)?;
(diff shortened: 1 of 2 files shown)


avro/src/decode.rs
```diff
@@ -222,46 +222,42 @@ fn decode_internal_body<R: Read, S: Borrow<Schema>>(
...
-                )
-            };
-            let uuid = Uuid::from_slice(&bytes).map_err(Details::ConvertSliceToUuid)?;
-            Ok(Value::Uuid(uuid))
+            let len = decode_len(reader)?;
+            if len == 16 {
+                let mut buf = [0u8; 16];
+                reader
+                    .read_exact(&mut buf)
+                    .map_err(|e| Details::ReadFixed(e, 16))?;
```
martin-g: Is `ReadFixed` the correct error here ? This is the `Bytes` arm

martin-g (approved):

avro/src/decode.rs
Kriskras99: No it is not, I'll fix it tonight!

Kriskras99:
Ignoring the CodeQL issues, they are fixed in a separate PR

PR #664, opened 2026-09-06
fix: Don't return `Value::Null` on EOF in `decode.rs`

Kriskras99:
This corrupt a union that is being decoded and makes the decoded value not match the schema.

Reported-by: Security scans

### DIFF
--- a/avro/src/decode.rs
+++ b/avro/src/decode.rs
@@ -180,19 +180,13 @@ fn decode_internal_body<R: Read, S: Borrow<Schema>>(
         Schema::Null => Ok(Value::Null),
         Schema::Boolean => {
             let mut buf = [0u8; 1];
-            match reader.read_exact(&mut buf[..]) {
-                Ok(_) => match buf[0] {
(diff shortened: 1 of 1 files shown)


martin-g (approved):

kevinjqliu:
I was just looking into this issue from the iceberg-rust side, thank you for fixing it! 😄

Kriskras99:
@kevinjqliu I would recommend you to use the Serde deserializer where possible. It is significantly faster if you need to have Rust datatypes after decoding/before encoding. But if you work with generic data that is of course not always possible.

kevinjqliu:
Yessss I believe that’s part of the new 0.22 release. We need to upgrade the library, alraedy tracking that work in https://github.com/apache/iceberg-rust/issues/3063. I ran some numbers it’s around 5x faster than our current implementation and on par with a Cython avro reader from iceberg-python. Super excited to include it in iceberg-rust.

Thanks again for the awesome work!!

PR #666, opened 2026-09-06
fix!: Don't log potentially sensitive value in `Value::validate`

Kriskras99:
This is done by removing the logging completely and returing an error instead of a boolean. The code will also no longer panic when it fails to resolve the schema.

This is a breaking change, users can fix their code by adding `.is_ok()` to get the previous behaviour.

Reported-by: CodeQL

### DIFF
--- a/avro/src/error.rs
+++ b/avro/src/error.rs
@@ -729,6 +729,12 @@ pub enum Details {
         position: usize,
         total_elements: usize,
     },
+
+    #[error("The value is invalid for the given schema: {0}")]
+    InvalidValueForSchema(String),
+
+    #[error("The value is invalid for all the given schemas: {0:?}")]
+    InvalidValueForAllSchemas(Vec<String>),
 }
 
 #[derive(thiserror::Error, PartialEq)]
--- a/avro/src/schema/mod.rs
+++ b/avro/src/schema/mod.rs
@@ -2975,7 +2975,7 @@ mod tests {
         };
 
         let avro_value = crate::to_value(foo)?;
-        assert!(avro_value.validate(&schema));
+        avro_value.validate(&schema)?;
 
         let mut writer = crate::Writer::new(&schema, Vec::new())?;
 
@@ -3065,10 +3065,7 @@ mod tests {
             bar_use: Bar::Bar1,
         };
         let avro_value = crate::to_value(foo)?;
-        assert!(
-            avro_value.validate(&writer_schema),
-            "value is valid for schema",
-        );
+        avro_value.validate(&writer_schema)?;
         let datum = GenericDatumWriter::builder(&writer_schema)
             .build()?
             .write_value_to_vec(avro_value)?;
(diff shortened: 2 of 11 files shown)


martin-g:
IIRC this has been requested by someone to be able to debug validation issues. They said that it is hard to debug such issues when using big/huge schemas. The actual request was to print the path to the failing field but this was rather complex to implement.

Maybe log it as a `trace` ?! Or log only when `debug_assertions` is enabled ?!
Or make it configurable ?!
Or use custom Debug/Display/our-own impl that skips (sensitive) values ?!

Kriskras99:
I was already thinking about adding the path to the error. So I'll try that!

Kriskras99:
I still need to do schema, but here is a sample of improved error messages (top is new, bottom is old):
```
The value is invalid for the given schema: Unsupported value-schema combination! Value: Record.b.Record.z.Null, schema: Int
The value is invalid for the given schema: Unsupported value-schema combination! Value: Null, schema: Int
```
```
The value is invalid for the given schema: Unsupported value-schema combination! Value: Record.b.String, schema: Record(RecordSchema { name: Name { name: "Inner", .. }, fields: [RecordField { name: "z", schema: Int, .. }], .. })
The value is invalid for the given schema: Unsupported value-schema combination! Value: String("testing"), schema: Record(RecordSchema { name: Name { name: "Inner", .. }, fields: [RecordField { name: "z", schema: Int, .. }], .. })
```
```
The value is invalid for the given schema: Unsupported value-schema combination! Value: Record.a.Boolean, schema: Long
The value is invalid for the given schema: Unsupported value-schema combination! Value: Boolean(false), schema: Long
```
```
The value is invalid for the given schema: Enum("spades") is not at position 1 in the schema
The value is invalid for the given schema: Symbol 'spades' is not at position '1'
```
```
The value is invalid for the given schema: Size of Fixed (5) is different than the schema's size (4)
The value is invalid for the given schema: The value's size (5) is different than the schema's size (4)
```
```
Invalid value: Int(42) for schema: Boolean. Reason: Unsupported value-schema combination! Value: Int(42), schema: Boolean
Invalid value: Int(42) for schema: Boolean. Reason: Unsupported value-schema combination! Value: Int, schema: Boolean
```
```
Value Int(1) does not match schema String: Reason: Unsupported value-schema combination! Value: Int, schema: String
Value Int(1) does not match schema String: Reason: Unsupported value-schema combination! Value: Int(1), schema: String
```
You'll notice that when the actual value (`1`, `True`, `"Some string"`) is not relevant to the error, it is not in included.

It costs a bit of extra stack space and two extra references being passed to `validate_internal`, one for tracking the value path and one for tracking the schema path. There are no heap allocations and it writes the path directly into the `format!` string.

martin-g:
Would be possible to print the `name` for named schemas ? E.g. `Record[ns.myname].a.Boolean`

Kriskras99:
Absolutely, I was thinking of `Record{name}`. But that is only possible for the schema oath as `Value` doesn't store a name.

Kriskras99:
Schema path is now also added. I think this would also be useful in the errors for the serializer and deserializer. Let me know what you think.

PR #681, opened 2026-09-16
fix!: `Decimal` should not be able to decode from a zero-length bytes

### LINKED ISSUE #680: Decimal with zero length bytes can be constructed and decoded but fails to encode
```rs
    #[test]
    fn test_round_trip_decimal_length_zero() -> TestResult {
        use crate::decode::decode;
        use crate::encode::encode;
        use crate::encode::tests::success;

        let schema = Schema::parse_str(
            r#"{"type": "bytes", "logicalType": "decimal", "precision": 4, "scale": 2}"#,
        )?;
        let expected = Value::Decimal(Decimal::from(Vec::<u8>::new()));
        dbg!(&expected);
        // 0x00 is the varint encoding of length 0 (zero-length bytes)
        let mut buf: &[u8] = &[0x00];
        let decoded = decode(&schema, &mut buf)?;
        assert_eq!(decoded, expected);

        // Re-encoding the decoded value should round-trip, but it fails
        let mut buffer = Vec::new();
        encode(&decoded, &schema, &mut buffer).expect(&success(&decoded, &schema));

        Ok(())
    }
```

Output:
```
---- types::tests::test_round_trip_decimal_length_zero stdout ----
[avro/src/types.rs:1921:9] &expected = Decimal(
    Decimal {
        value: 0,
        len: 0,
    },
)

thread 'types::tests::test_round_trip_decimal_length_zero' (126727) panicked at avro/src/types.rs:1929:48:
Value: Decimal(Decimal { value: 0, len: 0 })
 should encode with schema:
Decimal(DecimalSchema { precision: 4, scale: 2, inner: Bytes }): Error { details: Number of bytes requested for decimal sign extension 0 is less than the number of bytes needed to decode 1 }


failures:
    types::tests::test_round_trip_decimal_length_zero
```

For avro cpp, it treats this as bytes so it succeeds in both encode/decode.
Java failed to decode.
```
Exception in thread "main" java.lang.NumberFormatException: Zero length BigInteger
        at java.base/java.math.BigInteger.<init>(BigInteger.java:312)
        at java.base/java.math.BigInteger.<init>(BigInteger.java:346)
        at org.apache.avro.Conversions$DecimalConversion.fromBytes(Conversions.java:111)
        at DecimalTest.main(DecimalTest.java:19)
```

I'm happy to make PR to fix this but it first needs to decide which behavior whether to:
- reject on decode (like java)
- accept and fix encode so it can round-trip (like cpp)

Kriskras99:
This is a breaking change because it is not possible to do `impl<T: AsRef<[u8]>> TryFrom<T> for Decimal` as we have a `TryFrom<Decimal>` implementation on `Vec<u8>` which causes auto-trait conflicts.

Fixes #680


Kriskras99:
I left some stuff uncommitted, I'll fix it tonight

avro/src/decode.rs
```diff
@@ -1182,4 +1182,16 @@ mod tests {
 
         Ok(())
     }
+
+    #[test]
+    fn avro_rs_680_decimal_bytes_zero_length() -> TestResult {
```
martin-g, line 1187: Let's add a test for Fixed too to cover the path at https://github.com/apache/avro-rs/blob/f7b8a7715b5e67fe4dc301143c4bd375f6f78c32/avro/src/decode.rs#L197

```
{"type": "fixed", "name": "f", "size": 0, "logicalType": "decimal", "precision": 4}
```

martin-g (approved):

avro/src/decimal.rs
```diff
@@ -111,13 +121,14 @@ impl From<Decimal> for BigInt {
...
-/// use apache_avro::Decimal;
-/// use std::convert::TryFrom;
-///
-/// let decimal = Decimal::from(vec![1, 24]);
-/// let maybe_bytes = <Vec<u8>>::try_from(&decimal);
+/// # use std::convert::TryFrom;
+/// # use apache_avro::{Decimal, Error};
+/// #
+/// let decimal = Decimal::new([1, 24])?;
+/// let maybe_bytes = <Vec<u8>>::try_from(decimal);
```
martin-g:
```suggestion
/// let maybe_bytes = <Vec<u8>>::try_from(&decimal);
```

